PR #1, opened 2023-06-02
WIP Config

davidwaroquiers:


utf:
Hey guys, I just wanted to discuss some thoughts I had about the config. I'm trying to think about the ideal config for my use case and I realise this might differ from what you have in mind.

My main concern is making it as simple as possible for the most common workflows. I imagine this would be:
- One database for the queue (e.g., one launchpad)
- One remote cluster for execution.

My main issues with the current config are:
1. There are a lot of levels you have to configure, host, cluster, resources. You need to write labels in each case and then reference between them.
2. Projects are another level which seems to be a very advanced feature and again add more complications. Even if you don't have to specify the project, you still need to understand what they are and what they impact.

I think a lot of the current configuration makes sense in terms of minimising duplication. E.g., if you have multiple resource types for the same cluster then you only need to write the cluster information once. A similar thing is true for projects if I understand them right.

My proposal would be to flatten everything. This will result in some duplication if you have many workers on the same system but you'd gain a lot in terms of simplicity and ease of understanding.

### My ideal configuration

This is looking towards the future (for when jobflow_remote is part of jobflow itself) but below would be my ideal configuration for the simple use case.

```yaml
job_store:
  docs_store:
    type: MongoStore
    database: database
    host: myhost.com
    port: 27017
    username: myuser
    password: mypass
    collection_name: outputs

queue:
  type: MongoStore
  database: mydb
  host: myhost.com
  port: 27017
  username: myuser
  password: mypass
  collection_name: queue
  
workers:
  archer2:
    type: remote
    host: login.archer2.ac.uk
    ssh_key: ~/.id_rsa
    username: myuser
    password: mypass
    
    scheduler: slurm
    nodes: 8
    walltime: 24:00:00
    project: myproject
    pre_run: |
      export ATOMATE2_VASP_CMD="mpirun ~/bin/vasp_std"
      module load anaconda
      conda activate jobflow
```

If you want to add another worker (e.g., with different resources) then you'd have to duplicate some information. E.g.,


```yaml
workers:
  archer2-small:
    type: remote
    host: login.archer2.ac.uk
    ssh_key: ~/.id_rsa
    username: myuser
    password: mypass
    
    scheduler: slurm
    nodes: 2
    walltime: 24:00:00
    project: myproject
    pre_run: |
      export ATOMATE2_VASP_CMD="mpirun ~/bin/vasp_std"
      module load anaconda
      conda activate jobflow

  archer2-big:
    type: remote
    host: login.archer2.ac.uk
    ssh_key: ~/.id_rsa
    username: myuser
    password: mypass
    
    scheduler: slurm
    nodes: 4
    walltime: 24:00:00
    project: myproject
    pre_run: |
      export ATOMATE2_VASP_CMD="mpirun ~/bin/vasp_std"
      module load anaconda
      conda activate jobflow
```

But ultimately it is only copy and paste rather than anything too complex.

### Projects

My proposal would be to remove the concept of projects entirely. Projects can be managed manually by just using different configuration files. If you really want some complex project management, this could be provided as a meta layer on top of jobflow to make managing multiple projects easier.

### Quickstart configuration

If we can replace the use of fireworks with the launchpad clone I have on my jobflow branch then I think we should try and make it so that you can use jobflow remote without any configuration. I should put this in a separate issue but want to put some motivation here too.

I'd like to make the default job store and default queue stores (e.g., the launchpad store) be MontyDb stores in the `~/.jobflow` folder. Just as a recap, the MontyDb store uses an sqlite database with a mongodb interface. So it is persistent and supports multiple simultaneous connects etc. That way you can run:
```
pip install jobflow
jobflow start
```
and then submit calculations without having to do any configuration. Obviously this would only allow you to run flows locally, but all your flows and outputs will automatically be stored rather than getting lost (e.g., what happens with the default memory store).

If you want to add a remote resource, all you'd have to do is set the `workers` part of the config. Similarly, if you wanted to use an external mongoDB database for the queue/job store. You'd just have to set those parts of the config.

### Conclusion

Obviously these are only suggestions. But it touches on things we've discussed in our meetings and I wanted to document some of the discussion here for @janosh and @arosen93.

Interested to hear your thoughts.

Andrew-S-Rosen:
I 100% agree with this! We need to minimize the getting started time for new users.

As an aside, I didn't know you could have a MongoDB interface to an SQLite database. Going to have to look into that!

janosh:
Forgot to reply here. One of the reasons I love YAML is it's little-known but very handy concept of anchors.

@utf You can alias single values or whole blocks and then refer to them, even override individual keys in a block. [Simple example](https://stackoverflow.com/a/71223888):

```yml
default_car: &default_car
  brand: xxx
  model: xxx
  color: red

cars_sold:
  - <<: *default_car
    color: blue

cars_available:
  - <<: *default_car
    color: yellow
```

Applied here:

```yml
default_worker: &default_worker
  type: remote
  host: login.archer2.ac.uk
  ssh_key: ~/.id_rsa
  username: myuser
  password: mypass
  
  scheduler: slurm
  walltime: 24:00:00
  project: myproject
  pre_run: |
    export ATOMATE2_VASP_CMD="mpirun ~/bin/vasp_std"
    module load anaconda
    conda activate jobflow

workers:
  archer2-small:
    <<: *default_worker
    nodes: 2

  archer2-big:
    <<: *default_worker
    nodes: 4
```

gpetretto:
Hi @utf,

thanks for the detailed comments. I definitely agree with simlifying the first approach with the tool. There are several points that you are making in this direction. Let me try to summarize them:

* Projects should be removed
* Merge Machine and Host in Worker
* Move Resources to the base of worker/machine
* A dictionary of workers instead of a list of workers
* Use a maggma Store to define run_db (thus allowing aMontyStore)
* Remove fireworks and switch to the jobflow queue manager

Hoping that I did not misunderstand or forgotten anything here are my thoughts on these:

### Projects should be removed

I don't agree on the point that even if you don't have to specify the project, you still need to understand what they are and what they impact. I am not really sure what you have in mind, but I would say that, as it is now, it can be seen from the user as a collection of configuration files in a folder. And if you have just one project it seems to me that a user can even never know that the concept of project exists.
One thing that is implied by the project is that the default folders where the transfered files, logs and files required for the daemon to run will be determined by the project name. But I rather see this as a benefit, because in your configuration file this part is missing and should be defined, without a natural choice for a default.
On the other hand this will ease things quite a bit as soon as one wants to move to multiple projects. Personally one of the issues with fireworks, when dealing with multiple projects, was indeed the need to keep working with different folders and environmental variables without a clear way to define them. 
In summary, to me the projects do not require particular knowledge from the beginner and could really be beneficial for the power users.

### Merge machine and Host in Worker

As you said, this was mainly done to reduce duplication. Another reason was that it makes it easier to handle the ssh connections, so that all the Machines go through the same connection. 
Indeed merging the objects will simplify the set up and I would go forward with that. In the internal implementation I would still try to share the connections among those that have the same configuration. Maybe by keeping the Host as an separate object, that Worker has an instance of. But this can be dealt with even at a later stage

### Move Resources to the base of worker/machine

I believe that the example you proposed has some issues concerning the resources definition. I think that the easiest way for most of the users would be to pass SLURM/PBS keywords. However, in this flattened document, one would need to distringuish between keywords related to the connection (can be local host or a remote host) and keywords related to the resources. If they are SLURM/PBS keywords they are hard to be properly defined/identified. We have defined a `QReources` object in QToolKit that could help standardize, but relying only on that would be a very strong limitation.

In general I think that for the user, having all the same level could even be more confusing.
Something like this seems cleaner compared to the completely flattened version and it does not seem to really add a barrier for the user:

```yaml
workers:
  archer2:
    type: remote
    host: login.archer2.ac.uk
    ssh_key: ~/.id_rsa
    username: myuser
    password: mypass
    
    scheduler: slurm
    pre_run: |
      export ATOMATE2_VASP_CMD="mpirun ~/bin/vasp_std"
      module load anaconda
      conda activate jobflow
    default_resources:
      nodes: 8
      walltime: 24:00:00
      project: myproject
```
 
### A dictionary of workers instead of a list of workers

I was relying mostly on the list with a specified name because internally the information is desirailized in an object and it is convenient that the name of the Machine/Worker was also an attribute of the object itself. I think it is indeed more readable in this way. Associating the name to the to the Worker object could be done internally in different ways, if needed.

### Use a maggma Store to define run_db + Remove fireworks and switch to the jobflow queue manager

I would discuss these points together because they might be related.
In general I was also thinking of moving away from the definition of the queue database from the one of the LaunchPad, so that it would make it easier to transition to some other manager. But is a maggma Store a good solution for that? The mongolike Stores are all based on a single collection. This is a bad definition for the fireworks database as it will force the user to define one collection, but the fireworks ones are forced to be used instead.
However, even for a new implementation, I am not convinced that putting all the documents in a single collection would be the best choice from a design point of view. Would you go for that option if not for the constraint of the maggma Store? 

MontyDB would definitely be a great option for the beginners. In the current implementation I am using aggregations in a few queries, that are not curently supported by MontyDB. But maybe some of these queries could also be written through multiple queries if MontyDB is being used. In general I believe that it would be a waste to give up advanced Mongodb features, limiting to what is available in MontyDB instead.

PR #12, opened 2023-08-03
Fix handling of null worker in `submit_flow`

### LINKED ISSUE #11: Worker can be set to `None` leaving dead jobs
In #7 it looks like I introduced one bug and uncovered another. At the moment, if you do not specify a worker for `submit_flow`, the config manager will pull out the first configured worker, but `submit_flow` does not make use of that worker and instead passes `None` to the firework creator. All machinery from then on seems to accept a null worker so a job enters the queue with no worker. I can fix this now.

ml-evs:
Closes #11


gpetretto:
Thanks!

Issue #13, opened 2023-08-07
Cryptic failures when worker's `workdir` is missing


ml-evs:
I was trying to set my work dir on a remote using an env var `$HOME`, but this lead to IOError without a particularly useful stack trace (just deep inside paramiko/fabric). Would be nice to add better logging for this and/or add it to the project check script (can make a PR).

gpetretto:
closed by #12

PR #15, opened 2023-08-14
Check `work_dir` with project check CLI and enforce absolute paths

### LINKED ISSUE #14: Runner fails when default `resources` are provider per worker
I have a config that specifies some default resources for a given worker, e.g.,

```yml
workers: 
  my_worker: 
    type: remote
    host: example.org
    resources:
       queue_name: long_queue
       account: budget123
       time_limit: 3600
```

This config will lead to an error like `ValueError: The following keys are not present in the template: time_limit, processes, queue_name`, as currently the `QResources` field names in the config are not being mapped to the queuing system-specific field names, as the `resources` field in the worker config is just a dict rather than a QResources. I will try to make this change and see what happens.

ml-evs:
~Continuation of my last PR, this closes #14~ and also adds a task for checking the configured `work_dir` (c.f. #13). I fell into various pits trying to set mine up (using e.g., `~`, `$HOME` or relative dirs did not work, so I changed the config description to force absolute paths).

### DIFF
--- a/src/jobflow_remote/cli/formatting.py
+++ b/src/jobflow_remote/cli/formatting.py
@@ -40,7 +40,6 @@ def get_job_info_table(jobs_info: list[JobInfo], verbosity: int):
     for ji in jobs_info:
         state = ji.state.name
         if ji.remote_state is not None and ji.state not in excluded_states:
-
             if ji.retry_time_limit is not None:
                 state += f" [[bold red]{ji.remote_state.name}[/]]"
             else:
@@ -136,6 +135,9 @@ def format_job_info(job_info: JobInfo, show_none: bool = False):
     error_remote = d.get("error_remote")
     if error_remote:
         d["error_remote"] = ReprStr(error_remote)
+    error_job = d.get("error_job")
+    if error_job:
+        d["error_job"] = ReprStr(error_job)
     return render_scope(d)
 
 
--- a/src/jobflow_remote/config/base.py
+++ b/src/jobflow_remote/config/base.py
@@ -146,6 +146,12 @@ class WorkerBase(BaseModel):
             raise ValueError(f"Unknown scheduler type {scheduler_type}")
         return scheduler_type
 
+    @validator("work_dir", always=True)
+    def check_work_dir(cls, v) -> Path:
+        if not v.is_absolute():
+            raise ValueError("`work_dir` must be an absolute path")
+        return v
+
     def get_scheduler_io(self) -> BaseSchedulerIO:
         """
         Get the BaseSchedulerIO from QToolKit depending on scheduler_type.
--- a/src/jobflow_remote/config/base.py
+++ b/src/jobflow_remote/config/base.py
@@ -111,8 +111,8 @@ class WorkerBase(BaseModel):
     scheduler_type: str = Field(
         description="Type of the scheduler. Depending on the values supported by QToolKit"
     )
-    work_dir: str = Field(
-        description="Path to the directory of the worker where subfolders for "
+    work_dir: Path = Field(
+        description="Absolute path of the directory of the worker where subfolders for "
         "executing the calculation will be created"
     )
     resources: dict | None = Field(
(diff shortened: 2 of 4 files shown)


src/jobflow_remote/jobs/submit.py
```diff
@@ -53,12 +53,15 @@ def submit_flow(
...
+    worker_obj = config_manager.get_worker(worker_name=worker, project_name=project)
     if isinstance(exec_config, str):
         config_manager.get_exec_config(
             exec_config_name=exec_config, project_name=project
         )
 
+    if not resources:
+        resources = worker_obj.resources
+
     wf = flow_to_workflow(
```
ml-evs: This was the simplest place to add/"fix" this behavior, as the configured QResources fields need to be mapped before the submission script is created (see #14)

gpetretto:
Thanks for the implementation.

I have some questions/concerns about this:
* Does switching from `str` to `Path` as a type for `work_dir` help in determining the fact that path should be absolute? Does this allow to use elements like `~`? 
* To enforce the fact that the path should be absolute the easier and safer way would probably be to add a validator in the model. 
* In the `_check_workdir` maybe the file could be deleted after the test?
* I don't really want to enforce the use of the QResources in the config file at this stage for multiple reasons. Mainly because it is basically untested and I don't think it will be good that the only way the users have to pass the configuration is to learn our own naming convention, without even the guarantee that will continue to exist in the future. QResources also does not cover all the possible cases. I also think that passing slurm/pbs keywords will be a request from the customers at this stage. As far as I am concerned `resources` should either be simply a `dict` and eventually be converted to Qresources at runtime, depending on its content, or being a `dict | QResources`. In both cases it will probably be a bit annoying to handle and the reason I did not do it in the first place.

ml-evs:
> Thanks for the implementation.
> 
> I have some questions/concerns about this:
> 
>     * Does switching from `str` to `Path` as a type for `work_dir` help in determining the fact that path should be absolute? Does this allow to use elements like `~`?

No, but it seemed like the more appropriate type hint. It will still be stored as e.g., `Path(~/blah)` with `str(Path("~/blah")) == "~/blah"` so current behavior should not change. It does not seem like the Fabric connection does any shell/user expansion at the moment.

>     * To enforce the fact that the path should be absolute the easier and safer way would probably be to add a validator in the model.

Yep, was going to ask if you wanted this.

>     * In the `_check_workdir` maybe the file could be deleted after the test?

👍 

>     * I don't really want to enforce the use of the QResources in the config file at this stage for multiple reasons. Mainly because it is basically untested and I don't think it will be good that the only way the users have to pass the configuration is to learn our own naming convention, without even the guarantee that will continue to exist in the future. QResources also does not cover all the possible cases. I also think that passing slurm/pbs keywords will be a request from the customers at this stage. As far as I am concerned `resources` should either be simply a `dict` and eventually be converted to Qresources at runtime, depending on its content, or being a `dict | QResources`. In both cases it will probably be a bit annoying to handle and the reason I did not do it in the first place.

Fair enough -- if the expected behaviour is to pass queue-specific stuff in this config, then I can change that back to a raw dict. I think the way it is currently laid out, it will still need the other change that passes the config resources into the flow converter function, but I can verify this first.

src/jobflow_remote/jobs/submit.py
gpetretto: In principle it picks the resources from the worker at runtime:
https://github.com/Matgenix/jobflow-remote/blob/808ca4a6ff9dae2c8d7d0840458ac4d2dae6f8c2/src/jobflow_remote/jobs/runner.py#L382
Did this not work? 
I don't have a strong opinion whether it is better to fix it at creation time or run time, but I would keep only one point where it is set, otherwise it might be confusing.

ml-evs: My bad, after checking again with the new slurm fields, it does indeed work at runtime without this change. I've undone these changes

ml-evs:
I've just rebased this PR to remove the initial concern in #14 and instead just contains a few tweaks. Feel free to do what you want with it!

gpetretto (approved):
Thanks a lot for all the updates.

PR #17, opened 2023-08-20
Restore queries based on job id

gpetretto:
Restoring the option to query based on job_id.

* for commands where a single id can be passed:
  *  the id can either be an integer or a string, it will then be recognized as a db_id or job_id depending on the type (thus the `--db-id` does not exist anymore)
  * an optional additional argument to specify the index has been added. if omitted the job corresponding to the last index will be used. 
  * query based on descending sort on the index
* for commands where multiple ids can be passed:
  * preserved the difference between job_ids and db_ids. Not sure if it would be easy to handle in case one passes a mix of the two if a single option is used. Could be enforced if it would be better to have a single option like in the previous type of commands
  * when querying based on job_ids the index is mandatory and should be provided in the form `UUID:INDEX` (e.g. `e1d66c4f-81db-4fff-bda2-2bf1d79d5961:2`)
  * the query is *not* an aggregation, using uuid+index to identify each job.
* when querying to get flows 
  * if job_ids are used the index is not need. Since all the jobs with the same uuid but different index will belong to the same flow it would be redundant.
  * passing the index will not work, but there is no check whether the index was actually passed in the format `UUID:INDEX`.It will just result it will simply not find the job, since there is no job with id with the form `e1d66c4f-81db-4fff-bda2-2bf1d79d5961:2`. Is it fine like this or should there be such a check? 
  * the query is the same as before

### DIFF
--- a/src/jobflow_remote/cli/types.py
+++ b/src/jobflow_remote/cli/types.py
@@ -10,12 +10,25 @@ from jobflow_remote.cli.utils import SerializeFileFormat, SortOption
 from jobflow_remote.config.base import LogLevel
 from jobflow_remote.jobs.state import JobState, RemoteState
 
+job_ids_indexes_opt = Annotated[
+    Optional[List[str]],
+    typer.Option(
+        "--job-id",
+        "-jid",
+        help="One or more pair of job ids (i.e. uuids) and job index formatted "
+        "as UUID:INDEX (e.g. e1d66c4f-81db-4fff-bda2-2bf1d79d5961:2). "
+        "The index is mandatory",
+    ),
+]
+
+
 job_ids_opt = Annotated[
     Optional[List[str]],
     typer.Option(
         "--job-id",
         "-jid",
-        help="One or more job ids (i.e. uuids)",
+        help="One or more job ids (i.e. uuids). Only the id is needed since "
+        "jobs with the same uuid belong to the same flow",
     ),
 ]
 
@@ -150,21 +163,18 @@ reverse_sort_flag_opt = Annotated[
 ]
 
 
-job_id_arg = Annotated[str, typer.Argument(help="The ID of the job (i.e. the uuid)")]
-
-db_id_arg = Annotated[
-    int, typer.Argument(help="The DB id of the job (i.e. an integer)")
+job_db_id_arg = Annotated[
+    str,
+    typer.Argument(
+        help="The ID of the job can the db id (i.e. an integer) or a string (i.e. the uuid)",
+        metavar="ID",
+    ),
 ]
-
-
-db_id_flag_opt = Annotated[
-    bool,
-    typer.Option(
-        "--db-id",
-        "-db",
-        help=(
-            "If set the id passed would be considered to be the DB id (i.e. an integer)"
-        ),
+job_index_arg = Annotated[
+    Optional[int],
+    typer.Argument(
+        help="The index of the job. If not defined the job with the largest index is selected",
+        metavar="INDEX",
     ),
 ]
 
--- a/src/jobflow_remote/cli/utils.py
+++ b/src/jobflow_remote/cli/utils.py
@@ -48,19 +48,19 @@ class ReprStr(str):
         return self
 
 
-def exit_with_error_msg(message, code=1, **kwargs):
+def exit_with_error_msg(message: str, code: int = 1, **kwargs):
     kwargs.setdefault("style", "red")
     err_console.print(message, **kwargs)
     raise typer.Exit(code)
 
 
-def exit_with_warning_msg(message, code=0, **kwargs):
+def exit_with_warning_msg(message: str, code: int = 0, **kwargs):
     kwargs.setdefault("style", "gold1")
     err_console.print(message, **kwargs)
     raise typer.Exit(code)
 
 
-def print_success_msg(message="operation completed", **kwargs):
+def print_success_msg(message: str = "operation completed", **kwargs):
     kwargs.setdefault("style", "green")
     out_console.print(message, **kwargs)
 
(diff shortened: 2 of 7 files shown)


src/jobflow_remote/cli/types.py
```diff
@@ -150,21 +163,18 @@
 ]
 
 
-job_id_arg = Annotated[str, typer.Argument(help="The ID of the job (i.e. the uuid)")]
-
-db_id_arg = Annotated[
-    int, typer.Argument(help="The DB id of the job (i.e. an integer)")
+job_db_id_arg = Annotated[
+    str,
```
davidwaroquiers, line 167: Shouldn't this be a Union[int, str] ? Would it have a strong impact anywhere ?

src/jobflow_remote/cli/utils.py
```diff
@@ -114,19 +114,38 @@ def loading_spinner(processing: bool = True):
...
+    return db_id, job_id
+
+
+def get_job_ids_indexes(job_ids: list[str] | None) -> list[tuple[str, int]] | None:
+    if not job_ids:
+        return None
+    job_ids_indexes = []
+    for j in job_ids:
+        split = j.split(":")
+        if len(split) != 2 or not split[1].isnumeric():
```
davidwaroquiers, line 142: could also check whether the uuid is correct, e.g. using something like (another implementation could use a regex):

```
import uuid

def is_valid_uuid(uuid_str):
    try:
        uuid_obj = uuid.UUID(uuid_str)
        return str(uuid_obj) == uuid_str
    except ValueError:
        return False
```

Is it worth ? What do you think? We could inform the user in case it's not.

davidwaroquiers: Note this could also be checked in the above function if we decide to do so here.

src/jobflow_remote/fireworks/launchpad.py
```diff
@@ -333,24 +338,43 @@ def get_fw_dict(self, fw_id: int | None = None, job_id: str | None = None):
...
+        sort: list | None = None
         if fw_id:
             query["fw_id"] = fw_id
         if job_id:
             query[FW_UUID_PATH] = job_id
+            if job_index is None:
+                sort = [[FW_INDEX_PATH, DESCENDING]]
+            else:
+                query[FW_INDEX_PATH] = job_index
         if not query:
```
davidwaroquiers, line 362: What happens if both are passed ? Should we prevent that ?

src/jobflow_remote/fireworks/launchpad.py
```diff
@@ -333,24 +338,43 @@ def get_fw_dict(self, fw_id: int | None = None, job_id: str | None = None):
...
+        self,
+        fw_id: int | None = None,
+        job_id: str | None = None,
+        job_index: int | None = None,
+    ):
         if job_id is None and fw_id is None:
             raise ValueError("At least one among fw_id and job_id should be defined")
         if job_id:
-            fw_id = self.get_fw_id_from_job_id(job_id)
+            fw_id = self.get_fw_id_from_job_id(job_id, job_index)
```
davidwaroquiers, line 377: Maybe we should prevent to pass both a job_id and a fw_id ? Or is that a use case ? (I would think no but maybe I'm wrong)

davidwaroquiers (approved):
Great!
I have one or two points I am wondering (but maybe it is not really relevant/needed):
- what happens in the python interface if you were to pass both a db_id and a uuid ? Maybe this could be checked somewhere so that the user cannot pass both ?
- we could check whether the uuid part of the UUID:INDEX is indeed a uuid (you are already checking that index is an integer). I proposed a small is_valid_uuid function.
Anyway these are tiny things. I let you decide if it is worth or not.
Thanks a lot!

src/jobflow_remote/cli/types.py
gpetretto: Typer currently does not support that

src/jobflow_remote/cli/utils.py
gpetretto: I considered that, but I am not sure if it is worth. We also don't check that the user doesn't pass a negative db_id for example. In general I expect that when using the uuid it will be a copy/paste. The addition is small though, if you feel it will be better I can add it.

davidwaroquiers: It's just that the user may copy paste only part of the uuid by mistake (e.g. miss one character at the beginning or end), so maybe informing him that the format is wrong might give him a clue as to copy it again. Anyway, I let you decide whether to add this.

src/jobflow_remote/cli/types.py
davidwaroquiers: Ok thx, didn't know that. Resolving this.

src/jobflow_remote/fireworks/launchpad.py
gpetretto: There was a check before in JobController. I removed it but forgot to move it here. I will add it.

src/jobflow_remote/fireworks/launchpad.py
gpetretto: fine to check it here as well

davidwaroquiers: Maybe there is only one place where it is needed to be checked, just not completely sure of the waterfall of methods called.

gpetretto: yes, I will try to reduce the amount of places where the check is implemented

gpetretto:
I added what discussed above. I also slipped in the fix for the problem about the `--flow-id` option that @marcodigennaro  pointed out.

davidwaroquiers (approved):
Thanks!

PR #18, opened 2023-08-21
Fixed jf flow list.

### LINKED ISSUE #16: List of flows by state does not work
It seems that the jf flow list -s SOMESTATE does not work. I will try to investigate.

davidwaroquiers:
All flows would always be returned without this. Basically the flow state was transformed to a JobState and never entered one of the if statements in jobcontroller. I also added the check in the jobcontroller so that if it is not one of the flow states it raises a runtimerror (well its the final else basically).


gpetretto (approved):
Great. Thanks!

PR #27, opened 2023-09-21
Add docker-based integration test framework

ml-evs:
This WIP PR adds the start of integration tests for jobflow-remote.

The Python Docker SDK is used within pytest to build an image that combines slurm, ssh and MongoDB, which are then launched as ephemeral containers that can receive jobs in future test cases.

There's still a couple of bits to iron out before this all works.


davidwaroquiers:
Very nice! This will be very useful and definitely needed to have good integration tests, thanks a lot for looking into that!

We can probably replicate that (more or less, e.g. no need for mongodb there, but might be good to see if we can have an openpbs docker as well ?) for QToolKit.

I have a few questions/comments (which I'm not sure are relevant as I am not an expert) at this stage (without entering into details, I went fast through the code/files).
Do you know if there is some caching mechanism in the github actions ? I would guess that if the docker has to be reconstructed from scratch everytime, it would be rather long ? Or maybe not ?
I don't think the setup of fireworks is needed (also, it will disappear soon in principle, @gpetretto is working on the new queue)

ml-evs:
> Very nice! This will be very useful and definitely needed to have good integration tests, thanks a lot for looking into that!

> We can probably replicate that (more or less, e.g. no need for mongodb there, but might be good to see if we can have an openpbs docker as well ?) for QToolKit.

Should be easy enough! Someone had already made a slurm image preconfigured for this kind of thing so it was fairly quick, but I can investigate PBS too. It might be worth splitting MongoDB out to a separate container in that case, but it seemed like overkill for now.

> Do you know if there is some caching mechanism in the github actions ? 

Yes, though its a bit fiddly, you can cache layers for a given repo. As mentioned above, the base image with slurm is already on Dockerhub, so the build just needs to install ssh + mongo. If we split off mongo to a new container, then we could just use the mongo pre-built images (I think I will do this anyway). 

> I would guess that if the docker has to be reconstructed from scratch everytime, it would be rather long ?

It's on the order of a few minutes locally (and then once built the overhead is about 10-15 seconds per test run to launch the containers).

> I don't think the setup of fireworks is needed (also, it will disappear soon in principle, @gpetretto is working on the new queue)

Well, my intention was to add it here so we can properly test the new queue (I think that was my initial motivation for doing this now!) There's not too much to do now, and if I separate out the MongoDB then it will be easy to switch Fireworks out for the new queue.

davidwaroquiers:
Thanks for the details ! :)

In any case, this is very good!

ml-evs:
As we don't have the CI running from forks enabled atm, this PR is superseded by #32

PR #29, opened 2023-09-29
[WIP] pydantic2 updates

gpetretto:
Updates to make jobflow_remote compatible with pydantic version >=2.
In principle I tested and it seems to work correctly with the latest versions of jobflow and maggma. However, I prefer not to merge right away, as this will be incompatible with pydantic<2 and it would require updating the whole set of MP related packages. I started to open this PR in case anybody needs it or wants to review/test.


ml-evs:
Is there a blocker for this @gpetretto? Now atomate2 has updated (and is now adding many new workflows each release) I guess we need to catch up... I'll test this out and see if I run into any issues.

davidwaroquiers:
> Is there a blocker for this @gpetretto? Now atomate2 has updated (and is now adding many new workflows each release) I guess we need to catch up... I'll test this out and see if I run into any issues.

These changes are included in the #37 I think. Do you think it would be good to merge it before anyway ?

gpetretto:
Indeed these changes are also included in the other PR. 
I merged there the commits from this branch, so in principle it should not be a problem to merge this one as well.

PR #31, opened 2023-10-06
Use the qtoolkit and jobflow PyPI packages, add direct pydantic dep

### LINKED ISSUE #30: Jobflow dependency
I think we could now use the official jobflow repo. Or should we wait (e.g. because of pydantic v2 ?).

https://github.com/Matgenix/jobflow-remote/blob/92f9a570fa18da9470248993fd855efcd778d296/pyproject.toml#L29

ml-evs:
Closes #30 

Related to #29

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,11 +26,11 @@ classifiers = [
 ]
 requires-python = ">=3.8"
 dependencies =[
-    "jobflow[strict] @ git+ssh://git@github.com/materialsproject/jobflow",
+    "jobflow[strict]<=0.1.13",
     "fireworks",
     "fabric",
     "tomlkit",
-    "qtoolkit @ git+ssh://git@github.com/matgenix/qtoolkit",  # TODO: Should be updated here when released
+    "qtoolkit",
     "typer",
     "rich",
     "psutil",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,8 @@ classifiers = [
 ]
 requires-python = ">=3.8"
 dependencies =[
-    "jobflow[strict]<=0.1.13",
+    "jobflow[strict]",
+    "pydantic<2",
     "fireworks",
     "fabric",
     "tomlkit",


ml-evs:
@gpetretto / @davidwaroquiers this would be helpful to merge before #29 is finished, then we can undo the changes after. (I also realise I do not have the power to request reviewers here... perhaps I could be added as a maintainer?)

gpetretto:
The current jobflow[strict] version pins pydantic to version 2.4.2. The latest jobflow version supporting pydantic <2 was 0.1.13. Maybe it would be better to also pin the jobflow requirement to that version? I am not sure if having `pydantic<2` here could cause some conflicts.

pyproject.toml
```diff
@@ -26,11 +26,12 @@ classifiers = [
 ]
 requires-python = ">=3.8"
 dependencies =[
-    "jobflow[strict] @ git+ssh://git@github.com/materialsproject/jobflow",
+    "jobflow[strict]",
+    "pydantic<2",
```
davidwaroquiers, line 30: Will these work ? jobflow (v0.1.14) is now based on pydantic 2 (maybe it should be jobflow[strict]<0.1.14 at this stage then ?). In any case, it will work when everything has migrated of course.

davidwaroquiers (approved):
Just the comment on jobflow version wrt pydantic version. Not sure about it.

pyproject.toml
ml-evs: It works now too, as pip does the resolution and figures out it has to use 0.1.13 of jobflow to fulfill the next requirement.

davidwaroquiers: Ok nice, then all good to me!

ml-evs:
> The current jobflow[strict] version pins pydantic to version 2.4.2. The latest jobflow version supporting pydantic <2 was 0.1.13. Maybe it would be better to also pin the jobflow requirement to that version? I am not sure if having `pydantic<2` here could cause some conflicts.

It could be pinned, but this seems like the lightest weight way of doing it -- all that #29 needs to do afterwards is lower pin pydantic~=2.0 and then jobflow>=0.1.14 will be used. Up to you if you'd rather pin jobflow too though!

ml-evs:
I guess @gpetretto you also mean whether it is worth even adding pydantic as a dependency, instead just going along with jobflow? I think as we are using specific pydantic code here, then it should definitely appear as a dependency, even if the `strict` mode of jobflow is installing a specific version of it already. There were quite a few breaking pydantic changes even within the v1.x series...

gpetretto:
If pip resolves the correct version of jobflow to use it is fine as it is.

gpetretto (approved):

PR #32, opened 2023-10-10
Add containerized integration tests for remote and local workers using slurm

ml-evs:
Supersedes #27 as the CI currently won't run from forks.

Currently blocked by #34

This PR adds the start of integration tests for jobflow-remote.

The Python Docker SDK is used within pytest to build containers for slurm+ssh and MongoDB, which are then launched as ephemeral containers that can receive jobs in future test cases.


.github/workflows/testing.yml
```diff
@@ -54,5 +54,9 @@ jobs:
           python -m pip install --upgrade pip
             pip install .[strict,tests,docs]
 
-      - name: Test
-        run: pytest --cov=jobflow_remote --cov-report=xml
+      - name: Unit tests
+        run: pytest --cov=jobflow_remote --cov-report=xml --ignore tests/integration
+      
+      - name: Integration tests
+        # Unfortunate workaround required until we have better test/settings isolation
```
davidwaroquiers: Could we use the mark feature of pytest ? Or is it bad practice ? Integration tests could be marked as e.g. "integration" (with a custom `@pytest.mark.integration` decorator) and unit tests as "unit" (similar) then we run integration tests with `pytest --cov=jobflow_remote --cov-report=xml -m integration` and unit tests with `pytest --cov=jobflow_remote --cov-report=xml -m unit`

One question though here is the coverage. There will be first a coverage from the unit tests, then overwritten by integration tests. I guess (but I may be wrong) that e.g. codecov only have one coverage "report" per repository or can we define multiple ones ?

ml-evs: Yes, at the moment this will overwrite the coverage. Instead you can pass `--append` (or something) to get it to work, but as there are no other tests at the moment I think it's fine... Long-term I'm not sure how to fix this without more of a restructure, the problem is that the first jobflow_remote import loads the settings, which I need to overwrite to get the integration tests to work. So probably fine just to keep them separate forever and mark them not to run locally.

I'm still having some outstanding issues where the runner picks up the job but doesn't actually submit it in the queue (and does not make the queue database), hopefully will be able to iron these out soon...

ml-evs: I also can't figure out why `mypy` is complaining so much in this PR, and not in #34...

davidwaroquiers:
Hi @ml-evs,

I hope you are fine!

Just a quick update on the new queue by @gpetretto. I wanted to know if you think it would be fine to merge this new queue in develop and then later on (when you have time) merge in your integration test infrastructure ?

The idea would be to
1. make a fireworks branch from develop (just to keep a "historical" branch for some time)
2. merge @gpetretto's queue into develop
3. make the repo public (merging queue into develop is for that reason, otherwise, people would start testing the old fireworks-based implementation). Also related to my mail on the jobflow paper where we could barge in jobflow-remote a little bit more.
4. merge your add_integration_tests branch into develop (maybe a few conflicts to fix but I would guess not much, if at all ?)
5. ...

If you think it's a problem, we can wait.

Best!

ml-evs:
Hi @davidwaroquiers & @gpetretto, I am actually back and able to sit down at my desk for the first time today, and was planning on taking a pass a this. If you can give me until the end of the day, then I can stop holding you up either way if I can't get this to work! I just want to rewrite things to be a bit more general, e.g., use the jobflow status to verify the job rather than fireworks, then ideally we can migrate the tests for the new queue shortly afterwards. Excited for this to be public.

davidwaroquiers:
Hi @ml-evs , sure ! Thanks for taking a look at this! We will wait for your go of course.

ml-evs:
Hi @gpetretto and @davidwaroquiers, I think this PR is now ready to be reviewed properly. For some reason the linter is failing in files that I have not touched, but probably best not to try to fix that here. This PR boosts the coverage to around 40% but can now be extended to include more test cases and CLI interactions if desired.

tests/integration/conftest.py
```diff
@@ -0,0 +1,259 @@
...
+import tempfile
+import time
+from pathlib import Path
+
+import docker
+import pytest
+from docker.models.containers import Container
+
+
+def _get_free_port(upper_bound=50_000):
```
gpetretto, line 25: For some reason on my laptop (with MacOs) it never found a good port. 

Running this function outside of the tests I could see that at each loop it seemed to always return a port larger than the previous ones (not necessarily consecutives). Even between different executions of the loops and different shells. The first one was already aboe 50000 and so were all the subsequents. I suppose the choice of the port is performed by the OS, so I am not sure if it a problem of my laptop or MacOS in general.

I managed to run the tests by increasing the upper_bound. I don't know if there is any way to force a real randomization. An option could be to instead try random numbers in python, but it may be less likely to find an available port.

tests/integration/conftest.py
```diff
@@ -0,0 +1,259 @@
...
+    if attempts == max_attempts:
+        raise RuntimeError(
+            f"Could not find a free port to use with the provided {upper_bound=}."
+        )
+
+    return port
+
+
+def _get_random_name(length=6):
+    return "".join(random.choice("abcdef") for _ in range(length))
```
gpetretto, line 50: Not a big deal, but this could probably be replaced by `"".join(random.choices("abcdef", k=length))`

tests/integration/test_slurm.py
```diff
@@ -0,0 +1,94 @@
...
+
+    flow = Flow([add_parent_1, add_parent_2, add_children, write])
+    submit_flow(flow, worker=worker)
+
+    runner = Runner()
+    runner.run(ticks=10)
+
+    assert job_controller.count_jobs({}) == 4
+    assert len(job_controller.get_jobs({})) == 4
+    assert job_controller.count_flows({}) == 1
```
gpetretto: Can you add a check here that the jobs are actually completed? On my laptop these are in the REMOTE_ERROR state when running on the `test_remote_worker` (they are COMPLETED for the `test_local_worker` instead). See comment above above about the connection errors

tests/integration/test_slurm.py
```diff
@@ -0,0 +1,94 @@
...
+    assert len(job_controller.get_jobs({})) == 2
+    assert job_controller.count_flows({}) == 1
+
+    if worker == "test_local_worker":
+        assert job_controller.count_jobs(state=JobState.FAILED) == 2
+        assert job_controller.count_flows(state=FlowState.FAILED) == 1
+    else:
+        # The remote flow will be left in a running state until the remote errors can be rectified
+        assert job_controller.count_jobs(state=JobState.REMOTE_ERROR) == 2
+        assert job_controller.count_flows(state=FlowState.RUNNING) == 1
```
gpetretto: As in the comment above, in my case the REMOTE_ERROR is related to the fact that the system cannot connect to the slum container or execute the command correctly. 
This makes me suspect that the issue is not related to my setup. In fact, if a job fails *nicely* (i.e. the python process is not killed), it should still end up in a FAILED state. If not, there is a problem in jobflow-remote that I will need to investigate.

Maybe it will also be good to check that the `error` field in the job document contains the message from the exception "This job failed."

tests/integration/test_slurm.py
```diff
@@ -0,0 +1,94 @@
...
+    assert len(cm.projects) == 1
+    assert cm.projects[random_project_name]
+    project = cm.get_project()
+    assert len(project.workers) == 2
+
+
+def test_project_check(job_controller):
+    from jobflow_remote.cli.project import check
+
+    assert check(print_errors=True) is None
```
gpetretto: since `check` is a CLI command it always returns `None`. In my case this is failing, but the test passes. It would be possible to move the logic of the `check` function to a lower lever function, but then it would be difficult to show the progresses in the CLI. Maybe we can just add a `return` with the list of errors.

tests/integration/conftest.py
```diff
@@ -0,0 +1,259 @@
...
+            "test_remote_worker": dict(
+                type="remote",
+                host="localhost",
+                port=slurm_ssh_port,
+                scheduler_type="slurm",
+                work_dir="/home/jobflow/jfr",
+                user="jobflow",
+                password="jobflow",
+                pre_run="/home/jobflow/.venv/bin/activate",
+                resources={"partition": "debug", "ntasks": 1, "time": "00:01:00"},
```
gpetretto, line 228: In my case, not sure if it is common, I could not connect to the slurm host. I paritally solved by adding this to the dictionary:
```python
connect_kwargs={"allow_agent": False,"look_for_keys": False}
```
"partially" in the sense that before it would just say that no connection were available. With this it seems to be able to establish the connection, but the execution of all the commands fail. 

The weird thing is that, if running a command using the `RemoteHost` while the tests are running and the container is up using a script from outsite the tests the commands are executed correctly (if using the above `connect_kwargs`). After some testing I would first like to know if this is a problem of my laptop or is happening on other systems as well.

In any case I believe it may be that the error is there in other cases, since the tests were passing before I made any change, but just because they don't check the final state of the jobs (see comments below on the test).

gpetretto (commented):
Thanks a lot for finalising this! 
As detailed in the comments, I verified that on my laptop some errors were present even when the tests passed. I did not manage to completely track down the source of the issue. Can you please check if this is something limited to my system or is a general error?

If you want we can have a chat so I can give you more details on the tests that I did.

tests/integration/conftest.py
ml-evs: Hmmm, must admit I didn't check how portable this approach would be outside of Linux. I had to add the upper bound as some ports are "free" but restricted for use by docker in Linux. I'll try removing this bound for now.

tests/integration/conftest.py
ml-evs:
```suggestion
    return "".join(random.choice("abcdef", k=length))
```

tests/integration/test_slurm.py
ml-evs: I'll check the issues raised here and above. I naively assumed it would get stuck in the remote error phase as I am only running 10 "ticks" of the runner, i.e. not long enough for it to be retried enough times to class as a failure -- I haven't checked this though. I'll add some tests for the job doc as suggested.

tests/integration/test_slurm.py
ml-evs: I'll play around with `capsys` and see if checking stderr is enough for this test.

tests/integration/conftest.py
ml-evs: Beefing up the tests of the job state seems like the best way to approach this, I'll get it into a state where that works on my end/GH actions and can then ping you for review to test locally again.

.github/workflows/testing.yml
davidwaroquiers: > Yes, at the moment this will overwrite the coverage. Instead you can pass `--append` (or something) to get it to work, but as there are no other tests at the moment I think it's fine... Long-term I'm not sure how to fix this without more of a restructure, the problem is that the first jobflow_remote import loads the settings, which I need to overwrite to get the integration tests to work. So probably fine just to keep them separate forever and mark them not to run locally.
> 
> I'm still having some outstanding issues where the runner picks up the job but doesn't actually submit it in the queue (and does not make the queue database), hopefully will be able to iron these out soon...

I've seen this regarding grouping/separating coverage in codecov: https://docs.codecov.com/docs/flags
I think it would allow to say e.g. we want 100% coverage on unit tests and 95% on integration tests (or anything else, could even be 100 and 100 with specific hold out uncovered lines for each "type", unit/integration/..., of test).
Not urgent or absolutely necessary at this stage but could be useful.

davidwaroquiers:
Any news on this PR ? Gabriel is starting to try out jobflow remote as well. One thing he mentioned that did not work right now (from the documentation), is the installation from pypi (as there is no release yet). I think it would be good to make a first release (even if it hasn't been reviewed by others). What do you think ?

ml-evs:
> Any news on this PR ? Gabriel is starting to try out jobflow remote as well. One thing he mentioned that did not work right now (from the documentation), is the installation from pypi (as there is no release yet). I think it would be good to make a first release (even if it hasn't been reviewed by others). What do you think ?

This PR is now in a bit of a weird state (the reverse to usual) where the tests seem to work fine in the CI, but not locally for @gpetretto. I've been trying out various tweaks to the set up that don't seem to make a huge difference... The only failing aspect is the linter, which I am still confused about as it works outside of this PR. As these tests are not "user-facing" I think it would be fine to merge this before a PyPI release, but probably @gpetretto can have the final say!

gpetretto:
I still did not manage to make the tests pass on my laptop. Actually I have run out of ideas of what could be going wrong. However, as this seems to be a problem with my configuration, I think we can proceed to merge.

gpetretto (approved):

ml-evs:
Thanks @gpetretto, I don't like passing code reviews by exhaustion of the reviewer (and myself ;)) but maybe we can return to this at a later date. I'll spend a little while rebasing this PR into a few thematic commits and perform some final checks before merging later today.

gpetretto:
Sure. I will also try to check again my system later on.
Thanks for all the work!

davidwaroquiers (approved):

ml-evs:
The linter is still complaining about lots of files I haven't touched (that it doesn't complain about from other branches...) but I think this is now good to go. I added a few additional test cases that can be built upon in the future, and will come back around to trying to get this to run locally. Probably the best approach would be to mark these tests as skipped by default (i.e., locally), and then only have them run in the CI with a controlled environment? This would avoid all the issues with OS-dependent port privileges for docker, and local SSH configs getting in the way (I realised sometimes the tests would fail for me if I hadn't unlocked my SSH agent in the current session...). I'll ping you both once more for review in case you want to check the last commit (I squashed most of the others down) but feel free to merge if you are happy.

ml-evs:
This is the cause of the issues we've been running into with output capturing (https://github.com/pyinvoke/invoke/issues/710), unfortunately there seems to be no solution beyond running the tests with `-s`...

tests/integration/conftest.py
```diff
@@ -6,13 +6,22 @@
...
+import fabric
 import pytest
 from docker.models.containers import Container
 
 
+@pytest.fixture(autouse=True)
+def mock_fabric_run(monkeypatch):
+    monkeypatch.setattr(
+        fabric.Connection, "run", partialmethod(fabric.Connection.run, in_stream=False)
+    )
```
ml-evs, line 22: Okay, this seems to remedy the issue around the `jfr project check` test failing when pytest is capturing output.

gpetretto:
Some good news from my side. Although not all the test pass locally.
The tests now run and I don't have the connection problem anymore. However, I have seen that apparently the container is not immediately reachable even when `container.status` is `running`. I have solved it by adding an additional `time.sleep(1)` after the `while container.status != "running"` loop. Maybe we can add this in the code as well?

Unfortunately some tests are failing because in the slurm container an error appers when submitting a script to slurm. I tried to log in into the container and submit a simple script and I get the same error when it runs:

> slurmstepd-63b56928c860: error: setgroups: Operation not permitted

Submitting a script running `sudo sbatch submit.sh` successfully runs the script instead. Did you maybe encounter a similar problem? 

Aside from this, even if we cannot fix it for me locally you can proceed and merge when you see fit.

gpetretto (approved):

ml-evs:
> Some good news from my side. Although not all the test pass locally. The tests now run and I don't have the connection problem anymore. However, I have seen that apparently the container is not immediately reachable even when `container.status` is `running`. I have solved it by adding an additional `time.sleep(1)` after the `while container.status != "running"` loop. Maybe we can add this in the code as well?

I can add an extra sleep but I'm surprised its not ready in time for you (or perhaps it is ready too early so the existing 1s sleep is never triggered vs always triggered for me (not sure its worth delving into this).

> Unfortunately some tests are failing because in the slurm container an error appers when submitting a script to slurm. I tried to log in into the container and submit a simple script and I get the same error when it runs:
> 
> > slurmstepd-63b56928c860: error: setgroups: Operation not permitted
> 
> Submitting a script running `sudo sbatch submit.sh` successfully runs the script instead. Did you maybe encounter a similar problem?

Strange, I've never seen this. When you say logging in to the container do you mean ssh'ing as user jobflow or launching a shell with docker? It seems to work fairly reproducibly now for me between both Linux and Windows (via WSL)

> Aside from this, even if we cannot fix it for me locally you can proceed and merge when you see fit.

I'll try to trigger the same behaviour at my end but if it works locally and in the CI then I'll merge.

ml-evs:
> I'll try to trigger the same behaviour at my end but if it works locally and in the CI then I'll merge.

Yeah I definitely don't see this locally when doing `ssh jobflow@localhost -p <randomport>` (pw jobflow) and submitting a trivial slurm script without sudo, so I think I'll merge this now and we can try to debug later.

Also just to note, hardcoding the ports seemed to lead to more issues in the CI itself (each Python version probably needs a unique free port) so I ended up reverting it.

gpetretto:
> I can add an extra sleep but I'm surprised its not ready in time for you (or perhaps it is ready too early so the existing 1s sleep is never triggered vs always triggered for me (not sure its worth delving into this).

What I have seen is that even after the status is `running` the connection to the container is not readily available. Meaning that the first tests do not pass due to lack in connectivity, but they start to pass later on. One remark on your changes, is that I had added the additional sleep *after* the while loop, to guarantee that it sleeps a big longer. But even doing it before seems to do the trick.

> Strange, I've never seen this. When you say logging in to the container do you mean ssh'ing as user jobflow or launching a shell with docker? It seems to work fairly reproducibly now for me between both Linux and Windows (via WSL)

Yes, that is what I meant. I also tried removing entirely the image of the container to force it to recreate the container, but the problem persisted. We will see later if this proves to be an issue for other developers as well. It may be good to start a small section ofr developers in the documentation where we can write stuff like the requirements to run the tests and the option to run them with `CI=1`, alongside with these potentials issues I have been experiencing (although I have no solution to offer).

ml-evs:
> Yes, that is what I meant. I also tried removing entirely the image of the container to force it to recreate the container, but the problem persisted. We will see later if this proves to be an issue for other developers as well.

Frustrating for now though! Thanks for trying to persist with this, maybe we can attempt a debugging session together when we next have time.

> It may be good to start a small section ofr developers in the documentation where we can write stuff like the requirements to run the tests and the option to run them with `CI=1`, alongside with these potentials issues I have been experiencing (although I have no solution to offer).

Good point, I'll write this now.

PR #34, opened 2023-10-11
Linting fixes and CI config for multiple Python 3.9+

### LINKED ISSUE #33: Targeted Python versions
Quick Q for @gpetretto, which Python version are we targeting with jobflow-remote? Jobflow itself is 3.8+, which is what we have been testing until recently, but there is quite a lot of 3.10+ code (just nicer type hints really). I guess given that the normal way of using jobflow-remote is inside the submission scripts, which share an environment with the real "science" code, we should try to be as forgiving as possible... If we want to stick with 3.8+ (i.e., just follow jobflow from now on) I can make a PR that "downgrades" the code.

ml-evs:
Previously we had `fail-fast: true` (the default) which means that the first failing Python version would cancel all other tests. This PR changes it so that we can at least see which versions are failing and why (c.f. #33).

Also:

- [x] Removed the template docs workflow, which has never been working.
- [x] Use Python 3.9 specifically in pre-commit CI
- [x] Fix any outstanding linting issues
- [x] Fix dummy test on version number
- [x] Tweak CI config to cancel jobs on pushes to the same PR
- [x] Added a few cases of `from __future__ import annotations` that were missing for py39 compat
- [x] Add a test that checks that the top-level of the package can be imported successfully


ml-evs:
@gpetretto this PR addresses #33 and shows that we indeed currently fail on Python 3.8 (due to qtoolkit... so I've dropped it) and 3.9 (due to some weird typing issue with pydantic + __future__ annotations messing up -- see the test I added). With the other changes bundled here (enforcing Python 3.9 pre-commit, etc.), hopefully this means the CI will be useful once again at detecting these issues, ready to add some proper tests.

I'll try now to fix the 3.9 issue with pydantic, but if you already have some idea how, feel free...

utf:
As a side note, I found that I couldn't include the future annotations in modules defining pydantic models. So in atomate2 we remove the future import in those modules.

ml-evs:
> As a side note, I found that I couldn't include the future annotations in modules defining pydantic models. So in atomate2 we remove the future import in those modules.

Indeed, the issue as I understand it is that `from __future__ import annotations` is effectively just turning all annotations into postponed versions (i.e., strings), but when pydantic actually tries to use them it runs into issues. I think to support 3.9 we will have to downgrade the annotations in the config (as evidenced by the last test run).

ml-evs:
@gpetretto this is ready for you to look at I think, unfortunately in the process of fixing the CI there are a lot of whitespace/file ending fixes, so to quickly summarise the actual code changes:

- [x] use py39 compatible type hints for pydantic 88536804df6dec6f95687c3709a8746f43573d9b
- [x] add stacklevels to warnings (some flake8 plugin was complaining but it does make sense) 8d2490562ee49cad0e67ae414a82a98cdd42cb6e
- [x] add `from __future__ import annotations` in a few additional places -- if these still break then the next testing overhaul PR should catch them
- [x] add a test that checks that settings etc can be imported

gpetretto (approved):
looks good to me. Thanks!

PR #6, opened 2023-06-21
Go Live version

davidwaroquiers:
First production version.

PR #39, opened 2023-11-24
WIP: New queue

davidwaroquiers:


pyproject.toml
```diff
@@ -24,8 +23,21 @@ classifiers = [
     "Topic :: Other/Nonlisted Topic",
     "Topic :: Scientific/Engineering",
 ]
-requires-python = ">=3.8"
-dependencies =[]
+requires-python = ">=3.9"
+dependencies =[
+    "jobflow[strict]",
```
Andrew-S-Rosen: Can the strict be dropped? Makes it difficult to test out in a real production python env.

Andrew-S-Rosen (commented):
Looks exciting!

pyproject.toml
davidwaroquiers: I think so but I'll let @gpetretto confirm this and make the change.

gpetretto: It was introduced during the pydantic v1 -> v2 transition to keep everything consistent. I don't think it is needed anymore and I removed it.

davidwaroquiers:
I think this PR can be closed as all changes for the new queue have been merged into develop, and there is an open PR (#6 ) from develop to main to allow review of the code. @gpetretto

PR #46, opened 2023-12-20
Added GitHub release workflow and docs builds

davidwaroquiers:
Copied and modified the release workflow from qtoolkit.
Need @ml-evs to check it is fine.
Maybe it would be good to add the release procedure (when, how, who, ...) in the dev documentation some place.


.github/workflows/release.yml
```diff
@@ -0,0 +1,124 @@
...
+        python-version: '3.11'
+
+    - name: Install Python dependencies
+      run: |
+        python -m pip install -U pip
+        pip install -U setuptools wheel
+        pip install -e .[all]
+
+    - name: Update changelog
+      uses: CharMixer/auto-changelog-action@v1
```
ml-evs, line 45: Might have to delete the contents of the current changelog before this works usefully -- can also have a workflow in the tests that adds to the "Unreleased" section of the changelog every time you merge into `main`

.github/workflows/release.yml
```diff
@@ -0,0 +1,124 @@
...
+      with:
+        user: __token__
+        password: ${{ secrets.TEST_PYPI_PASSWORD }}
+        repository-url: https://test.pypi.org/legacy/
+
+    - name: Install docs dependencies
+      run: |
+        # Required to generate rst files from markdown
+        sudo apt install pandoc
+        pip install .[docs]
```
ml-evs: This is fine, but I would suggest building the docs before releasing and then only uploading after the release; this should prevent needing to re-release if the docs break in some unpredictable way

ml-evs (commented):
LGTM @davidwaroquiers, just a couple of comments below.

As is always the case with these things, the only problems will come from having the right permissions rather than anything else...

.github/workflows/release.yml
```diff
@@ -0,0 +1,124 @@
...
+      uses: CharMixer/auto-changelog-action@v1
+      with:
+        token: ${{ secrets.RELEASE_PAT_BOT }}
+        release_branch: ${{ env.PUBLISH_UPDATE_BRANCH }}
+        exclude_labels: "duplicate,question,invalid,wontfix,dependency_updates,skip_changelog"
+
+    - name: Update '${{ env.PUBLISH_UPDATE_BRANCH }}'
+      uses: CasperWA/push-protected@v2
+      with:
+        token: ${{ secrets.RELEASE_PAT_BOT }}
```
ml-evs, line 54: For example, I don't see this secret in the repo atm

.github/workflows/release.yml
```diff
@@ -0,0 +1,124 @@
...
+    - name: Build source distribution
+      run: |
+        pip install -U build
+        python -m build
+
+    - name: Publish package to Test PyPI first
+      uses: pypa/gh-action-pypi-publish@release/v1
+      with:
+        user: __token__
+        password: ${{ secrets.TEST_PYPI_PASSWORD }}
```
ml-evs, line 114: nor this one (sorry for breaking up the review)

.github/workflows/release.yml
davidwaroquiers: Ok I'll change that. I took it as such from qtoolkit, I will also modify it there then.

.github/workflows/release.yml
davidwaroquiers: Indeed, for qtoolkit, I think you added your own secrets there if I remember correctly ? Rationale was that e.g. pypi does not (yet, it is under "closed beta" as they say) organization accounts. Was it indeed what we did ? Are you ok to add your secret there ? (what does PAT mean btw ?)

.github/workflows/release.yml
davidwaroquiers: Same here, are you ok to add your secret ?

.github/workflows/release.yml
ml-evs: I can generate one and add it, sure!

PAT = Personal Access Token

ml-evs: Now is probably the time to also make a bot account for Matgenix, I will do that and use the token for it here. The email will be dev@matgenix.com, which presumably someone is watching?

.github/workflows/release.yml
ml-evs: Yep, just bear in mind that we don't have the "real" release workflow in this one yet

doc/source/conf.py
```diff
@@ -21,7 +21,7 @@
 
 # -- Project information -----------------------------------------------------
 
-project = "Jobflow-Remote"
+project = "Jobflow Remote"
 copyright = "2023, Matgenix SRL"
 author = "Guido Petretto, David Waroquiers"
```
davidwaroquiers: Matthew Evans

Issue #44, opened 2023-12-13
Question about supported data types


Andrew-S-Rosen:
Hi friends. What are the supported data types for dispatched functions in terms of the args/kwargs? Typical MSONable? I'm curious if passing a function as a keyword argument would be possible. **Disclaimer**: I haven't actually used jobflow-remote yet, so apologies if this is obvious once I do start to try it out.

Something like the following toy example (obviously one would never need to write it this way for such a simple example):

```python
# mymodule

def add(a: float, b: float) -> float:
    return a + b

def mult(a: float, b: float) -> float:
    return a * b
```

```python
from jobflow import job
from mymodule import add, mult

@job
def my_job(a: float, b: float, op: callable = add) -> float:
    return op(a, b)

my_job(a, b, op=mult)
```

davidwaroquiers:
Hi @Andrew-S-Rosen 

Indeed I believe it won't work (not necessarily due to jobflow-remote). It would indeed be a good addition but I think it should be within jobflow itself (or maybe monty MSONable ? not sure if that would be so easy ?). Typical args and kwargs are currently indeed MSONable objects.

Regarding jobflow-remote itself, if you'd like to have a call to have a quick demo, don't hesitate to reach out :)

utf:
There is already support for this in jobflow... I made callable functions msonable in monty in these PRs: 
- Builtin callables: https://github.com/materialsvirtuallab/monty/pull/251 
- General callables: https://github.com/materialsvirtuallab/monty/pull/239

davidwaroquiers:
Nice! I wasn't aware of that (or maybe forgot). Then it should also work in jobflow-remote but we should probably test this specifically just to be sure.

Andrew-S-Rosen:
This just made me so happy. Thank you both!! I had assumed it wasn't MSONable, but it looks like you already took care of that, @utf. Beautiful!!

PR #51, opened 2024-01-26
Add integration test for job with a callable as a kwarg

### LINKED ISSUE #45: Add tests for `Job`s with a function arg or kwarg
See discussion in #44. Just opening this as a separate issue so [@davidwaroquiers' comment](https://github.com/Matgenix/jobflow-remote/issues/44#issuecomment-1853445067) doesn't get lost.

ml-evs:
Related to #44 and builds atop #32, showing how callables can be used as keyword arguments to jobs. Note that this doesn't seem to work for `lambda`s (see the default kwarg for `jobflow_remote.testing.arithmetic`).

Closes #45.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -92,6 +92,7 @@ rst-roles = "class, func, ref, obj"
 [tool.mypy]
 ignore_missing_imports = true
 no_strict_optional = true
+follow_imports = "skip"
 
 [tool.pytest.ini_options]
 filterwarnings = [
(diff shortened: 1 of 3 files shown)


src/__init__.py
ml-evs, line 1: Okay, apparently deleting this file is what causes the linter to break between PRs... I will try to update this in a separate PR instead.

gpetretto (approved):

PR #54, opened 2024-01-28
Remove errant `__init__.py` that prevents mypy from working

ml-evs:
The file at `src/__init__.py` is not necessary and is somehow breaking mypy for the rest of the project. This PR removes that file, then adds a series of typing related fixes.

### DIFF
--- a/src/jobflow_remote/cli/job.py
+++ b/src/jobflow_remote/cli/job.py
@@ -1,5 +1,3 @@
-from __future__ import annotations
-
 import io
 from pathlib import Path
 from typing import Annotated
@@ -708,10 +706,10 @@ def resources(
     Set the resources for the selected Jobs. Only READY or WAITING Jobs.
     """
 
-    resources: dict | QResources = str_to_dict(resources_value)
+    resources = str_to_dict(resources_value)  # type: ignore
 
     if qresources:
-        resources = QResources(**resources)
+        resources = QResources(**resources)  # type: ignore[assignment]
 
     jc = get_job_controller()
     execute_multi_jobs_cmd(
--- a/src/jobflow_remote/cli/job.py
+++ b/src/jobflow_remote/cli/job.py
@@ -1,6 +1,8 @@
+from __future__ import annotations
+
 import io
 from pathlib import Path
-from typing import Annotated, Optional
+from typing import Annotated
 
 import typer
 from monty.json import jsanitize
@@ -685,7 +687,7 @@ def resources(
         typer.Option(
             "--qresources",
             "-qr",
-            help="If present the values will be interpreted as arguments for a QResources object",
+            help="If present the values in `resources_value` will be interpreted as arguments for a QResources object",
         ),
     ] = False,
     job_id: job_ids_indexes_opt = None,
@@ -705,10 +707,10 @@ def resources(
     Set the resources for the selected Jobs. Only READY or WAITING Jobs.
     """
 
-    resources_value = str_to_dict(resources_value)
+    resources: dict | QResources = str_to_dict(resources_value)
 
     if qresources:
-        resources_value = QResources(**resources_value)
+        resources = QResources(**resources)
 
     jc = get_job_controller()
     execute_multi_jobs_cmd(
@@ -791,7 +793,7 @@ def output(
     job_db_id: job_db_id_arg,
     job_index: job_index_arg = None,
     file_path: Annotated[
-        Optional[str],
+        str | None,
         typer.Option(
             "--path",
             "-p",
--- a/src/jobflow_remote/cli/job.py
+++ b/src/jobflow_remote/cli/job.py
@@ -706,10 +706,10 @@ def resources(
     Set the resources for the selected Jobs. Only READY or WAITING Jobs.
     """
 
-    resources = str_to_dict(resources_value)  # type: ignore
+    resources = str_to_dict(resources_value)
 
     if qresources:
-        resources = QResources(**resources)  # type: ignore[assignment]
+        resources = QResources(**resources)
 
     jc = get_job_controller()
     execute_multi_jobs_cmd(
--- a/src/jobflow_remote/cli/job.py
+++ b/src/jobflow_remote/cli/job.py
@@ -15,6 +15,7 @@ from jobflow_remote.cli.formatting import format_job_info, get_job_info_table
 from jobflow_remote.cli.jf import app
 from jobflow_remote.cli.jfr_typer import JFRTyper
 from jobflow_remote.cli.types import (
+    OptionalStr,
     break_lock_opt,
     days_opt,
     db_ids_opt,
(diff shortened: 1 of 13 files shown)


davidwaroquiers (approved):

gpetretto (approved):

ml-evs:
@gpetretto I had a go at fixing some of the typing issues suggested by mypy but I'm not entirely convinced by a lot of the annotations I am adding. Would you be able to take a look? (probably before merging this?)

gpetretto:
Sure. Do you want me to try to apply this change locally and check the typing issues? Or do you already have a list of problematic ones to be verified?

ml-evs:
> Sure. Do you want me to try to apply this change locally and check the typing issues? Or do you already have a list of problematic ones to be verified?

I'll push the changes I'm happy with to this PR and try to point out the ones that are weird

ml-evs:
> > Sure. Do you want me to try to apply this change locally and check the typing issues? Or do you already have a list of problematic ones to be verified?
> 
> I'll push the changes I'm happy with to this PR and try to point out the ones that are weird

I've just pushed these, the remaining lint errors are the weird ones (mostly lots of awkward typing around the db potentially returning `None`) -- hopefully the enum value changes are fine though

gpetretto:
I checked some of the errors related to the `None` and, while they are not real errors as the check on the object not being `None` is performed somewhere else, mypy is likely not able to deduce that. Not sure what is the best strategy here: either skip the check or adding some redundant check on the object not being None. I can do that, so I will also check that they are all indeed correct.

The other errors could probably be resolved by adding `__all__` to the `__init__.py` files that have imports.

I am not sure about replacing the values of the Enums with the Enums instance in the CLI. Does it work? Should I test it?

ml-evs:
I've finally got around to fixing the mypy issues in [b7b7f81](https://github.com/Matgenix/jobflow-remote/pull/54/commits/b7b7f8116ac4d0f3783020d6f0fba34593e6e83b), mostly by just adding checks for `None` where appropriate. Could you take another look @gpetretto?

src/jobflow_remote/cli/job.py
```diff
@@ -705,10 +707,10 @@ def resources(
     Set the resources for the selected Jobs. Only READY or WAITING Jobs.
     """
 
-    resources_value = str_to_dict(resources_value)
+    resources: dict | QResources = str_to_dict(resources_value)
 
     if qresources:
-        resources_value = QResources(**resources_value)
+        resources = QResources(**resources)
```
gpetretto, line 712: In the call to `execute_multi_jobs_cmd` below it keeps passing `resources_value`, so `resources` is never used.

src/jobflow_remote/cli/job.py
```diff
@@ -1,6 +1,8 @@
+from __future__ import annotations
```
gpetretto: This breaks the CLI. I always get an error 
> RuntimeError: Type not yet supported: <class 'jobflow_remote.cli.types.DictType'>

that goes away if remove this line and get back to `Optional` instead of `None`

src/jobflow_remote/cli/job.py
```diff
@@ -74,7 +76,7 @@ def jobs_list(
     hours: hours_opt = None,
     verbosity: verbosity_opt = 0,
     max_results: max_results_opt = 100,
-    sort: sort_opt = SortOption.UPDATED_ON.value,
+    sort: sort_opt = SortOption.UPDATED_ON,
```
gpetretto: Switching to the Enum also breaks the CLI. I think I had tried at the time. Typer expects a string (as it would from the user, and then converts it to the value. Can we revert back these to be the Enum values? Or does it work for you?

src/jobflow_remote/cli/project.py
```diff
@@ -96,7 +97,7 @@ def current_project(ctx: typer.Context):
 @app_project.command()
 def generate(
     name: Annotated[str, typer.Argument(help="Name of the project")],
-    file_format: serialize_file_format_opt = SerializeFileFormat.YAML.value,
+    file_format: serialize_file_format_opt = SerializeFileFormat.YAML,
```
gpetretto: Enum value as above

src/jobflow_remote/cli/runner.py
```diff
@@ -27,7 +27,7 @@
 
 @app_runner.command()
 def run(
-    log_level: log_level_opt = LogLevel.INFO.value,
+    log_level: log_level_opt = LogLevel.INFO,
```
gpetretto: Enum value as above

src/jobflow_remote/cli/runner.py
```diff
@@ -108,7 +108,7 @@ def start(
             help="Use a single process for the runner",
         ),
     ] = False,
-    log_level: log_level_opt = LogLevel.INFO.value,
+    log_level: log_level_opt = LogLevel.INFO,
```
gpetretto: Enum value as above

gpetretto (commented):
I am fine with most of the changes, but some of those to the CLI seem to break Typer. Does it work for you? Otherwise I would revert them and ignore any related linter comments.

src/jobflow_remote/cli/job.py
ml-evs: Hmmm, that's annoying, as this change is made automatically by pyupgrade... I'll investigate further

src/jobflow_remote/cli/job.py
ml-evs: I think I've caught and reverted all of these, just had to add various `# type: ignore`s

src/jobflow_remote/cli/job.py
ml-evs: Here are the related issues: https://github.com/tiangolo/typer/issues/533 and https://github.com/tiangolo/typer/issues/348 -- for now we can disable this kind of rewrite in pyupgrade (I will try to go through and check that there are no other occurences)

src/jobflow_remote/cli/job.py
ml-evs: I have updated this but cannot easily test

ml-evs:
Annoyingly this is now back in the state where the tests work locally but not in the CI; I'll try to debug it when I get a chance but for now do not let this PR hold up the other ones that actually add features.

gpetretto:
> Annoyingly this is now back in the state where the tests work locally but not in the CI; I'll try to debug it when I get a chance but for now do not let this PR hold up the other ones that actually add features.

After your last commit it seems that only a lint issue remains in the tests. Depending on your availability, I suppose these can be fixed and then we can merge this before the other PRs.

ml-evs:
> After your last commit it seems that only a lint issue remains in the tests. Depending on your availability, I suppose these can be fixed and then we can merge this before the other PRs.

Think that's got it -- I won't be able to take a look at the other PRs again until Friday so feel free to do with this as you please!

PR #59, opened 2024-01-30
Add testing and runtime checks for additional stores

### LINKED ISSUE #76: Add guard for missing `additional_store` before executing job
As tested in #59, if a job requires an additional store for its output, and that store is not defined in the JFR config, then the job will only fail at the point of trying to access that store.

I think we should add a guard in the runner to prevent this job from being submitted. I'm not sure what `JobState` it should be left with (perhaps a new one, `HELD` or some such, so it can be easily rerun once the runner config is updated?) Thoughts @gpetretto?

ml-evs:
This PR adds some integration tests that make use of `additional_stores`. In the process, I also added a check that makes sure that any `additional_stores` requested by a job are checked before the job executes, rather than failing afterwards.

We could also add a check at submission time for this, which should be a future PR.

Closes #76

### DIFF
--- a/src/jobflow_remote/remote/data.py
+++ b/src/jobflow_remote/remote/data.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import inspect
 import io
 import logging
 import os
@@ -7,11 +8,18 @@ from pathlib import Path
 from typing import Any
 
 import orjson
+from jobflow.core.job import Job
 from jobflow.core.store import JobStore
 from maggma.stores.mongolike import JSONStore
 from monty.json import jsanitize
 
+from jobflow_remote.jobs.data import RemoteError
 from jobflow_remote.utils.data import uuid_to_path
+
+JOB_INIT_ARGS = {k for k in inspect.signature(Job).parameters.keys() if k != "kwargs"}
+"""A set of the arguments of the Job constructor which
+can be used to detect additional custom arguments
+"""
 
 
 def get_job_path(
@@ -165,4 +173,16 @@ def resolve_job_dict_args(job_dict: dict, store: JobStore) -> dict:
     # substitution is in place
     job_dict["function_args"] = resolved_args
     job_dict["function_kwargs"] = resolved_kwargs
+
+    additional_store_names = set(job_dict.keys()) - JOB_INIT_ARGS
+    for store_name in additional_store_names:
+        # Exclude MSON fields
+        if store_name.startswith("@"):
+            continue
+        if store_name not in store.additional_stores:
+            raise RemoteError(
+                f"Additional store {store_name!r} is not configured for this project.",
+                no_retry=True,
+            )
+
     return job_dict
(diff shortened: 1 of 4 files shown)


src/jobflow_remote/remote/data.py
```diff
@@ -4,16 +4,34 @@
...
+from jobflow.core.job import Job, JobConfig
 from jobflow.core.store import JobStore
 from maggma.stores.mongolike import JSONStore
 from monty.json import jsanitize
+from pydantic import BaseModel
 
 from jobflow_remote.utils.data import uuid_to_path
 
 
+class JobWrapper(Job):
```
gpetretto: I am not sure I understand the need for this class. If it is just to have a list of arguments of `Job`'s `__init__` can't it be a list of strings instead? This will not offer any automatic update in case the `Job` API is changed, right? If we want automation maybe we can use `inspect` on the `__init__` method?

I am mainly concerned that a new developer (or me in 3 months :) )  arriving here would think this wrapper may have some functional use, while it seems to be equivalent to a list of attributes.

src/jobflow_remote/remote/data.py
```diff
@@ -165,4 +183,15 @@ def resolve_job_dict_args(job_dict: dict, store: JobStore) -> dict:
...
     job_dict["function_kwargs"] = resolved_kwargs
+
+    expected_fields = set(JobWrapper.__annotations__)
+    additional_store_names = set(job_dict.keys()) - expected_fields
+    for additional_store in additional_store_names:
+        if additional_store.startswith("@"):
+            continue
+        if additional_store not in store.additional_stores:
+            raise ValueError(
+                f"Additional store {additional_store!r} is not configured."
```
gpetretto: I would raise a `RemoteError` with `no_retry=True` here. This would avoid that the runner retries multiple times. See also https://github.com/Matgenix/jobflow-remote/issues/76.

src/jobflow_remote/remote/data.py
ml-evs: The idea would be to contribute it back up to Jobflow, with the benefit being the automatic updates. That is, if we think this is useful!

gpetretto: Do you mean the you would move all the properties to `Job` so to access them with `__annotations__`? Otherwise I still don't see the advantage of having a class rather than simply a list of strings. Maybe the whole check function could be moved to jobflow? For example a `Job.required_additional_stores` property? I still need to check if this can always be determined only looking at the `Job` and not at the content of the job's `function`.

ml-evs: I've just replaced the wrapper with `inspect.signature(Job)` which seems to work quite well.

Happy to migrate this to jobflow if you think that's the better place for it!

gpetretto: It may be interesting to move it to jobflow, but I think we can leave it here for the time being and see if any other option or check may be needed.

gpetretto (approved):

Issue #58, opened 2024-01-30
Handling Multi-Factor Authentication for workers


davidwaroquiers:
Following question from @JaGeo, opening here an issue on the MFA topic. Let's gather ideas, info, existing solutions, problems, ... related to the fact that clusters are slowly (or maybe rapidly ?) moving to MFA authentication.
@gpetretto  did some tests (could you maybe summarize insights here ?)

ml-evs:
One "simple" approach that might work here (and that I have seen suggested on some machines) is to use SSH multiplexing, so that once an authenticated connection is created (by the user, with TOTP or whatever), then the connection is kept open and all further connections within the session go through it. This is handled with a simple socket file, so Paramiko/Fabric etc. should just seamlessly work too. This would require some structure whereby JFR notifies the user that a new TOTP code is required to keep monitoring jobs, and will depend on how stringently machines enforce these timeouts (in practice, the timeout can be infinite...). The relevant SSH config parameters are `ControlPersist` and `ControlMaster` (see https://www.man7.org/linux/man-pages/man5/ssh_config.5.html and e.g. the suggestions in the Cambridge docs https://docs.hpc.cam.ac.uk/hpc/user-guide/mfa.html#reducing-the-effort-of-mfa-connection-sharing).

ml-evs:
Also related is the previous discussion at https://github.com/Matgenix/qtoolkit/issues/14

JaGeo:
Would this also require a setup without password? We have both, password and MFA at the moment. Using a key pair does not help with the password.

gpetretto:
I think in general it would not require a password (at least from my tests). The problem is that unfortunately it seems that multiplexing is not (yet) supported by paramiko: https://github.com/paramiko/paramiko/issues/852.
Until support is addeed I am afraid we should deal with this issue in some other way. 

The fact that jobflow-remote *should* keep the connection with the host open would allow to pass the OTP when the Runner starts. 
The connection can still be closed and indeed it would be good if the user could be notified about that. However, I am not sure if there is a convenient way of doing that.

ml-evs:
As an aside, I've just pushed #60 which can be used as a test bed for some of these approaches (both by manually building and launching the MFA-enabled Slurm container and testing locally, and by the eventual full JFR automation...). For now we should at least add clear error messages and a docs page about this until we have a real solution.

ml-evs:
[...]

PR #60, opened 2024-01-31
Add integration tests for MFA-enabled SSH/Slurm container

ml-evs:
Related to #58.

This PR adds a test environment for an MFA-enabled SSH/Slurm worker. It uses the [google-authenticator-libpam](https://github.com/google/google-authenticator-libpam) wrapper for Ubuntu to set up time-based MFA on the default `jobflow` user, with a pre-defined emergency backup code (which can only be used for one login without container rebuild).

Currently, the tests fail, but should probably be upgraded to instead ensure that sensible errors are returned for the MFA enabled worker.

This is hopefully a **somewhat** realistic emulation of an MFA-enabled supercomputer (perhaps excluding the backup codes), and can be used to test potential future workarounds involving multiplexing or otherwise.

Note, as stated in the Dockerfile.slurm.mfa header, although the two integration test dockerfiles are the same except for the final build stage, as the Docker Python SDK does not support BuildKit (https://github.com/docker/docker-py/issues/2230), we cannot use multi-stage builds here.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -57,6 +57,12 @@ def slurm_ssh_port():
 
 
 @pytest.fixture(scope="session")
+def slurm_ssh_mfa_port():
+    """The exposed local port for SSH connections to the Slurm + MFA container."""
+    yield _get_free_port()
+
+
+@pytest.fixture(scope="session")
 def db_port():
     """The exposed local port for connections to the MongoDB stores."""
     yield _get_free_port()
@@ -72,6 +78,7 @@ def build_and_launch_container(
     dockerfile: Path | None = None,
     image_name: str | None = None,
     ports: dict[str, int] | None = None,
+    target: str | None = None,
 ):
     """Builds and/or launches a container, returning the container object.
 
@@ -81,6 +88,7 @@ def build_and_launch_container(
         image_name: Either the tag to attach to the built image, or an image
             name to pull from the web (may require authenticated docker client).
         ports: A port specification to use for the launched container.
+        target: The docker build target to use.
 
     Yields:
         The launched container object, then stops the container after use.
@@ -95,6 +103,7 @@ def build_and_launch_container(
             tag=image_name,
             rm=True,
             quiet=False,
+            target=target,
         )
 
         for step in logs:
@@ -139,7 +148,22 @@ def slurm_container(docker_client, slurm_ssh_port):
     yield from build_and_launch_container(
         docker_client,
         Path("./tests/integration/dockerfiles/Dockerfile.slurm"),
-        "jobflow-slurm:latest",
+        image_name="jobflow-slurm:latest",
+        ports=ports,
+    )
+
+
+@pytest.fixture(scope="session", autouse=True)
+def slurm_container_mfa(docker_client, slurm_ssh_mfa_port):
+    """Build and launch a container running Slurm + SSH with multi-factor authentication,
+    exposed on a random available port.
+
+    """
+    ports = {"22/tcp": slurm_ssh_mfa_port}
+    yield from build_and_launch_container(
+        docker_client,
+        Path("./tests/integration/dockerfiles/Dockerfile.slurm.mfa"),
+        image_name="jobflow-slurm-mfa:latest",
         ports=ports,
     )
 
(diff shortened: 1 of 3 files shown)


ml-evs:
Results of some testing with paramiko and oath:

1. oath can successfully programmatically generate TOTP codes given the initial secret key/QR code you are provided (e.g., I was able to ssh into jobflow-mfa container using an oath code rather than a backup code). This would require storing the secret key provided by the server (which presumably is exactly what apps like the Google Authenticator do; obviously this would need to be done securely but perhaps this isn't any worse than allowing jobflow to use ssh keys?)
2. Paramiko does not support `password` + 2FA verification, instead only relying on pubkey + 2FA. With some gross hacking, I was able to connect to the jobflow-mfa container using a backup code with paramiko, but could not keep the connection open long enough to run commands. I managed to get it to this point by overriding the detection of "partial" auth to work with passwords:

  ```python
    def _parse_userauth_failure(self, m):
        authlist = m.get_list()
        # TODO 4.0: we aren't giving callers access to authlist _unless_ it's
        # partial authentication, so eg authtype=none can't work unless we
        # tweak this.
        partial = m.get_boolean()
        if partial:
            self._log(INFO, "Authentication continues...")
            self._log(DEBUG, "Methods: " + str(authlist))
            self.transport.saved_exception = PartialAuthentication(authlist)
  ```

  (https://github.com/paramiko/paramiko/blob/eb470b129ce0651102aec6340a08190fe393b94e/paramiko)/auth_handler.py#L740-L744)
  
  It might be possible with the help of paramiko devs to enable this route, otherwise we could make a specialised `SSHClient` for this case if `password` + 2FA token is going to be a common thing.


If 2. was implemented, I think this would give us everything required to use 2FA (as long as its based on OATH), as jobflow could mimic the user running an app on a mobile phone for auth. Of course, the problem with this is is that it presumably breaks the usage terms of the supercomputers that use 2FA (which in some cases might be actually illegal too? :sweat_smile:)

gpetretto:
I managed to connect with password+OPT using paramiko with this code: 
```python
sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
sock.connect(("mfahost.example.com", 22))
ts = paramiko.Transport(sock)
ts.start_client(timeout=10)
ts.auth_interactive_dumb(user)
```
from [here](https://stackoverflow.com/a/64318194). This prompts for both. I did not test it, but in principle it should be possible to replace `auth_interactive_dumb` with `auth_interactive` and define a handler that inserts password and OTP (as suggested in [another answer to the same question](https://stackoverflow.com/a/43943658)). I was investigating this to reduce the constraints of the solution I mentioned in the issue.

However, I would rather not offer a standardized way of automatically generating the OTP by storing the secret locally, unless this is a policy generally approved by the clusters. I imagine that such an approach will make the OTP generation pointless from the point of view of the administrators. While we cannot prevent individual users from hacking their way through, the administrators will likely disapprove if we actively document and support tools that break their policies. Maybe we can get in touch with the administrators of a few computing centers and check before proceeding with this approach?

ml-evs:
> I managed to connect with password+OPT using paramiko with this code: 
> ```python
> sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
> sock.connect(("mfahost.example.com", 22))
> ts = paramiko.Transport(sock)
> ts.start_client(timeout=10)
> ts.auth_interactive_dumb(user)
> ```

Nice, my attempts were to essentially trigger the call to the "dumb" route using the normal SSHClient but I didn't really get anywhere. 
 
> However, I would rather not offer a standardized way of automatically generating the OTP by storing the secret locally, unless this is a policy generally approved by the clusters. I imagine that such an approach will make the OTP generation pointless from the point of view of the administrators. While we cannot prevent individual users from hacking their way through, the administrators will likely disapprove if we actively document and support tools that break their policies. Maybe we can get in touch with the administrators of a few computing centers and check before proceeding with this approach?

I think reaching out is a good idea. Whilst this should never be the default, I'm sure some enterprising individuals would try this anyway so it might be good to have a proper encrypted store version built-in to jobflow. If the first login when launching a runner still required the TOTP then this is somehow spiritually the same as trusting Google or whoever with your secret key...

ml-evs:
This has gone stale

PR #78, opened 2024-02-22
handle multiple nested gateways in config

gpetretto:
This PR increases the options for configuring the connection to a remote worker. Giving more flexibility to set up a fabric connection with one or more gateways directly in the config file. In principle this could already be handled by properly defining the connections in the `~/.ssh/config` file, but this could cover a few more use cases, if the user has limited control on the machine hosting the runner.

The change should be backward compatible and not require updates to an existing config file.


gpetretto:
Closing as the changes have been merged in #83

PR #83, opened 2024-02-28
Interactive login for MFA

gpetretto:
Following the discussion in #58, and after some testing, this seems to be a working solution for computing centers requirering MFA.
Even if experimental I would start to merge this, so that it would be easier for others to test.

Note that the current implementation could also be expanded to cover more cases of interactive login. For example to let the user insert the passphrase of the ssh key. However, I would proceed with this kind of improvements only if a request comes in.

### DIFF
--- a/doc/source/user/advancedoptions.rst
+++ b/doc/source/user/advancedoptions.rst
@@ -38,10 +38,24 @@ procedure should not expose any sensitive information, you should consider if it
 against the security policies of the computing center you are working with.
 
 In order to deal with this kind of configuration, first set up the worker that
-requires an OTP in the standard way, as shown in the : ref:`projectconf worker` section.
+requires an OTP in the standard way, as shown in the :ref:`projectconf worker` section.
 Then, set the option ``interactive_login: true`` in the worker configuration.
 This will signal to the system that interaction from the user is required during the
 opening of the connection.
+
+In addition to the ``interactive_login`` option, there are two different use cases
+that should be handled separately by the user in the configuration:
+
+#. The connection requires **ssh key+OTP**: the ssh key **should be defined explicitly**
+   either in the ``~/.ssh/config`` file that defines the connection to the host, or
+   in the configuration of the worker (using the ``key_filename`` option).
+#. The connection requires **password+OTP**: no ssh key should be present in the
+   configuration files.
+
+.. warning::
+
+    The rest of the procedure is independent of which option is being used, but it
+    is crucial to properly define the options for the ssh key, depending on the use case.
 
 When starting the runner, the daemon should also be told that the user needs to
 input some information during the connection. The runner should thus be started with::
--- a/doc/source/user/introduction.rst
+++ b/doc/source/user/introduction.rst
@@ -64,10 +64,11 @@ Once the daemon is started, the runner loops over the different actions that it 
 perform and updates the state of Jobs in the database performing some actions on them.
 
 * After a Job has been ``CHECKED_OUT``, the Runner will proceed to upload the information required to run a Job. This includes:
-      - resolving all the references of the Job from the database (including everything in additional stores)
-      - using those data to generate a JSON representation of the Job without external references
-      - uploading a JSON file with this information on the runner
-  Once this is done, the state of the Job is ``UPLOADED``
+
+  - resolving all the references of the Job from the database (including everything in additional stores)
+  - using those data to generate a JSON representation of the Job without external references
+  - uploading a JSON file with this information on the runner
+  Once this is done, the state of the Job is ``UPLOADED``.
 * The runner generates a submission script suitable for the type of chosen worker.
   Uploads it and submits the job. The Job is now ``SUBMITTED``.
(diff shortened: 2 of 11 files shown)


gpetretto:
This PR now also includes the changes from #78, since had large overlaps for the changes in the remote host.
From the description of #78:
> This PR increases the options for configuring the connection to a remote worker. Giving more flexibility to set up a fabric connection with one or more gateways directly in the config file. In principle this could already be handled by properly defining the connections in the ~/.ssh/config file, but this could cover a few more use cases, if the user has limited control on the machine hosting the runner.
> The change should be backward compatible and not require updates to an existing config file.

ml-evs (approved):
I have tested this empirically with a machine that requires OTP + password, but as discussed I think there are still issues with SSH key + OTP that may not be easily solvable. Specific error I get is:

```
[18:35:29] WARNING  Error while trying the authentication with all the public keys available: ('Could not
                    deserialize key data. The data may be in an incorrect format, it may be encrypted with an
                    unsupported algorithm, or it may be an unsupported key type (e.g. EC curves with explicit
                    parameters).', [<OpenSSLError(code=503841036, lib=60, reason=524556,
                    reason_text=unsupported)>]). This may be due to the format of one of the keys. Authentication
                    will proceed with interactive prompts
```
with keys that otherwise work fine through SSH.

If you're happy to merge this PR as is, then so am I, but it might be worth adding some nicer error message wrapper and linking to the OTP docs where more detail can be provided about what we expect to work/fail. Perhaps this will only be clearer once merged and tested by others :upside_down_face:

gpetretto:
The "Error while trying the authentication with all the public key available" section was commented out in the last commit: https://github.com/Matgenix/jobflow-remote/blob/05a9e22f1d05a8288f20f07d74486c939ef98ab3/src/jobflow_remote/remote/host/remote.py#L410. May I ask you to try again with the latest version, as this could solve the ssh key issue? In that case the ssh key to be used should be indicated explicitly either in the ~/.ssh/config or in the worker settings.

ml-evs:
> The "Error while trying the authentication with all the public key available" section was commented out in the last commit:
> 
> https://github.com/Matgenix/jobflow-remote/blob/05a9e22f1d05a8288f20f07d74486c939ef98ab3/src/jobflow_remote/remote/host/remote.py#L410
> . May I ask you to try again with the latest version, as this could solve the ssh key issue? In that case the ssh key to be used should be indicated explicitly either in the ~/.ssh/config or in the worker settings.

Ah yes, I just tried it but explicitly setting the key in `.ssh/config` and it works well! Weird that it doesn't work with the same key used as a default! I would say this definitely needs to be added to the docs...

gpetretto:
> Ah yes, I just tried it but explicitly setting the key in `.ssh/config` and it works well! Weird that it doesn't work with the same key used as a default! I would say this definitely needs to be added to the docs...

I will definitely add this to the docs. I just wanted to check whether this solution was working or not.

As for the fact that it requires the key to be set explicitly it is due to how I have coded it. 
To summarize: paramiko/fabric can more or less deal on their own with the case of ssh-key+OTP. On the other hand they cannot deal at all with the case password+OTP (or at least I did not find any way). So I subclassed fabric's `OpenSSHAuthStrategy` and changed the `get_sources` method https://github.com/fabric/fabric/blob/988dd0fd05db47331cb43d0ea9787908ef33219c/fabric/auth.py#L184 hoping to handle both cases. However, as `OpenSSHAuthStrategy`'s documentation says, this is still experimental and as you have witnessed does not work well in our case. 
My current solution is that if you don't pass explicitly the ssh-key, I expect that you want to use password+OTP and the code uses the `InteractiveAuthStrategy`. If the ssh-key is instead explicitly passed, the code goes for the standard paramiko approach. Given how fabric's `Connection` works, it seems to me that the only way to avoid this would be to import quite a bit of code from paramiko to the `InteractiveAuthStrategy` and try to handle the different cases there. However, this would be quite difficult to maintain and I would rather avoid it.

However, it does not strictly need to be the way I did. For example the `interactive_login` configuration parameter could be a 3-way option: `None`, `"password"`, `"key"`. Do you think this would be better? 
I would say that the main issue at the moment is that we don't know if this covers enough use cases. So, it is not clear what would be the best solution in general.

ml-evs:
> However, it does not strictly need to be the way I did. For example the `interactive_login` configuration parameter could be a 3-way option: `None`, `"password"`, `"key"`. Do you think this would be better? I would say that the main issue at the moment is that we don't know if this covers enough use cases. So, it is not clear what would be the best solution in general.

I think that would be quite nice, if its not too much extra effort, but maybe that could wait for a future PR? It would probably not be too hard in future to allow `interactive_login` to be True, False/None, `password` or `key` so it wouldn't need to introduce a breaking change.  I can't imagine there would be that many more alternatives for HPC systems to implement? It would also make it much easier to document this feature, I think.

gpetretto:
Yes, I think we can address this in a subsequent PR, once it is a bit clearer what are the options needed.
I will update the documentation and merge. Thanks for the review.

Issue #79, opened 2024-02-23
Overhead from filewriting


JaGeo:
Hi all, 

I have been playing around with jobflow-remote for the last weeks now. I really like it. 

One thing, however, that leads to some issues was the following: I am trying to create a huge number of structures within the phonon workflow (>>4000 atoms, no symmetry, thus 24,000 structures). Of course, one could do MD; however, similar things could happen if one aims to calculate thermal conductivity. When I run this workflow with jobflow-remote, the jobs fail locally and remotely due to memory issues. There are no issues if I just run the same job in jobflow. Thus, I assume that the JSONStore and the saving process might lead to problems.

Have you encountered similar problems? Would one need to adapt how atomate2 workflows are written or are there possible other solutions? Locally, in my setup, for example, the MongoDB database could be used directly, and a JSONStore would not be needed.

gpetretto:
Hi @JaGeo,

thanks for reporting this. We have not used jobflow-remote for such large calculations yet, but I was somehow expecting that new issues would be triggered by starting using it for different use cases.

Can you be more explicit about where the issues happen? Where do the jobs fail? Do you have a stacktrace? Can you also share a simple script that would allow to reproduce the problem? 

> Have you encountered similar problems? Would one need to adapt how atomate2 workflows are written or are there possible other solutions? Locally, in my setup, for example, the MongoDB database could be used directly, and a JSONStore would not be needed.

You mean that MongoDB is reachable from the cluster nodes, and thus getting inputs and inserting outputs directly from the Jobs is possible? While I think it could be technically possible, I am a bit reluctant to go down this road, since with things happening in different ways may lead to some bugs and the code could be more difficult to maintain and expand. Let us first see if there is anything we can do to optimize the usage of the Store.

JaGeo:
I will make an example script to reproduce later on today.

JaGeo:
@gpetretto , here it is. 

The structure I uploaded is nonsense but should illustrate the problem as it creates 24000 displacements (thus, leave out the optimization part of the phonon run). I get failures when the supercells are created in jobflow-remote but not in jobflow. The reported issues via slurm are memory errors.

[to_guido.zip](https://github.com/Matgenix/jobflow-remote/files/14384069/to_guido.zip)

gpetretto:
Thanks! I will have a look at it. Just to be sure, when you say running with jobflow, you mean using `run_locally` or with fireworks?

JaGeo:
[...]

PR #88, opened 2024-03-04
Add options for remote JobStore

gpetretto:
Following #79, allow different options for the `JobStore` used remotely to dump the output. 
Can be configured with the `remote_jobstore` option to test different possibilities. The default is left to `None` to avoid constraining the structure of the configuration, as it is not clear what would be the best option in the future and to avoid issues with old configuration files in the future. The default corresponds to the old option. a maggma JSONStore with json files not zipped. The additional options are the json format based on [orjson](https://github.com/ijl/orjson) and [msgspec](https://jcristharif.com/msgspec/index.html), or [msgpack](https://github.com/msgpack/msgpack-python). Files can now optionally be zipped as well.

### DIFF
--- a/src/jobflow_remote/config/base.py
+++ b/src/jobflow_remote/config/base.py
@@ -540,9 +540,14 @@ class Project(BaseModel):
     )
     jobstore: dict = Field(
         default_factory=lambda: dict(DEFAULT_JOBSTORE),
-        description="The JobStore used for the input. Can contain the monty "
-        "serialized dictionary or the Store int the Jobflow format",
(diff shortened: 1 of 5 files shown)


ml-evs (approved):
Hi @gpetretto, the code here looks fine to me but I'm aware its not necessarily being hit by the tests (and right now I cannot easily test myself [train station wifi...]).

So I "approve" but would suggest that something is added to the docs about this, as it is not clear to me from the config description alone when I would want to configure a `remote_jobstore` (I think I understand it from the code itself but if this is meant to be user-facing then it needs more explanation).

PR #122, opened 2024-05-31
Migrate linting from `flake8`, `isort`, `autoflake` to `ruff`

janosh:
this PR brings linting in line with current `atomate2` and `jobflow` setup


src/jobflow_remote/jobs/runner.py
```diff
@@ -1073,9 +1073,9 @@ def update_batch_jobs(self) -> None:
                 )
 
                 if submit_result.status == SubmissionStatus.FAILED:
-                    logger.error(f"submission failed. {submit_result!r}")
+                    logger.exception(f"submission failed. {submit_result!r}")
```
ml-evs: I don't think this change is valid; my understanding is that logger.exception needs to be inside an exception handler but nothing is being raised here (there will be many such cases)

src/jobflow_remote/remote/host/remote.py
```diff
@@ -354,20 +351,16 @@ def inter_handler(title, instructions, prompt_list):
     Handler function for interactive prompts from the server.
     Used by Interactive AuthSource.
     """
-
     if title:
-        print(title.strip())
+        pass
```
ml-evs: Does ruff have a rule to remove print? I don't think these are load-bearing but if removed they should be replaced with appropriate debugging aids/logging (we don't always have a debugger available when running this code)

src/jobflow_remote/remote/data.py
```diff
@@ -96,22 +93,22 @@ def get_single_store(
...
+    if store_type == "msgpack":
         return MinimalMsgpackStore(file_path)
-    elif isinstance(store_type, dict):
+    if isinstance(store_type, dict):
         store_type = dict(store_type)
         store_type["path"] = file_path
         store = MontyDecoder().process_decoded(store_type)
         if not isinstance(store, Store):
-            raise ValueError(
+            raise TypeError(
```
ml-evs: I don't think this makes sense at all, and we are not testing if we rely on this returning a `ValueError` else where in the code.

src/jobflow_remote/cli/flow.py
```diff
@@ -254,9 +245,7 @@ def graph(
         exit_with_error_msg("No data matching the request")
 
     if print_mermaid:
-        from jobflow_remote.jobs.graph import get_mermaid
-
-        print(get_mermaid(flows_info[0]))
+        pass
```
ml-evs: This change seems to have removed some functionality. I think several of the rules being applied are not appropriate for a CLI tool

ml-evs (commented):
Thanks for this @janosh! Lots of helpful changes here. I'm onboard with the idea of migrating over to ruff but think we need to be a lot more careful introducing wholesale changes like this (especially as our test coverage is quite low at the moment).

Would you mind splitting this into two PRs: 

- 1680593ca8afcdc17130c39972fe213a4fcbe6fe -- using ruff for the "basic" linting we already do,
- everything else?

I'll use the comments of this review to pick out a few changes that I don't like that we need to ignore, but it is difficult to do an exhaustive review here.

janosh:
> Would you mind splitting this into two PRs:

of course! thought about doing it that way after starting the PR but then figured i'd wait to hear your preference. i kept only the safe auto-fixes in this PR and ignored all other existing `ruff` errors. i'll push the commits that addressed those into 2nd PR once this one is accepted

src/jobflow_remote/cli/flow.py
janosh: good catch. will be reverted in the follow-up PR

src/jobflow_remote/remote/host/remote.py
janosh: `ruff` does have such a rule: `T201`. should have ignored that rule

src/jobflow_remote/jobs/runner.py
janosh: thanks, i'll look into this for the 2nd PR

ml-evs (commented):
Thanks @janosh, this is looking great to me now! I'll let @gpetretto take the final decision

gpetretto:
Thanks @janosh for the PR and @ml-evs for going though it. These changes are fine with me as well.

gpetretto (approved):

Issue #125, opened 2024-06-11
How to update failed jobs input parameters ?


naik-aakash:
Hi @davidwaroquiers  and @gpetretto ,

Is there any way to query and update failed jobs input parameters in the Jobflow remote? 

Example use case, what I mean is as follows

For example, some vasp jobs failed in a flow using atomate2, and I now need to update the jobs' INCAR parameters, for example `ALGO `/ `NPAR`, specifically for this failed jobs not complete flow. I cannot just rerun the job; I first need to update these parameters in the DB.

I looked into the code; the only thing I could edit so far was the execution config for already added jobs using `JobController`. (This is very handy already)

Maybe I am overlooking something. If there exists anything for this, can you please let me know?

gpetretto:
Hi @naik-aakash,
an option that is available at the moment, altough not very good, is the private method `JobController._set_job_properties()`. This is what is used to set execution config by the method that you mentioned.
This should probably be made public. If you think it might be useful I will do this right away. It could also be complemented by an additional method that would directly set values in the `job` subsection of the document in the DB. Maybe `_set_job_properties` could be renamed to `set_job_doc_properties` and add a `set_job_values` method.

As far as I am concerned, for atomate2 workflows I think it would be useful to have a function like `update_job_incar_settings` that would directly update the settings of the input set generator for a Vasp Job in the DB. It would be difficult for the generic user to know that one needs to update the `job.function.@bound.input_set_generator.user_incar_settings` part of the document. I was planning to write such a function myself on the first occasion I would have needed it. Of course such a function cannot sit in jobflow-remote, but I am not sure if it would be fine to have that directly in atomate2 either.

naik-aakash:
> Hi @naik-aakash, an option that is available at the moment, altough not very good, is the private method `JobController._set_job_properties()`. This is what is used to set execution config by the method that you mentioned. This should probably be made public. If you think it might be useful I will do this right away. It could also be complemented by an additional method that would directly set values in the `job` subsection of the document in the DB. Maybe `_set_job_properties` could be renamed to `set_job_doc_properties` and add a `set_job_values` method.
> 
[...]

PR #132, opened 2024-06-21
Small updates to JobController and CLI

### LINKED ISSUE #129: Feature request: optionally delete outputs when doing jf flow delete
When deleting a flow using jf flow delete command, only the queue related databases are deleted.
Add an option to also delete the outputs from the outputs database (if it has run of course ...).

### LINKED ISSUE #130: Feature request: show more information about the flow(s) that will be deleted
When doing jf flow delete ..., add more information about which flow id(s) and the job that will be deleted when asking for confirmation.

gpetretto:
Addressing a few points discusses in issues:
*  `set_job_doc_properties` is a public method, deprecated `_set_job_properties` #125
* Added option `-v` and `-o` to `jf flow delete` to print the list of flows that will be deleted and delete also the outputs from the `JobStore` #129 #130

### DIFF
--- a/src/jobflow_remote/jobs/runner.py
+++ b/src/jobflow_remote/jobs/runner.py
@@ -412,7 +412,6 @@ class Runner:
         if job_id:
             query["uuid"] = job_id[0]
             query["index"] = job_id[1]
-        print(query)
         job_data = self.job_controller.checkout_job(query=query)
         if not job_data:
             if not db_id and not job_id:
--- a/src/jobflow_remote/jobs/runner.py
+++ b/src/jobflow_remote/jobs/runner.py
@@ -412,6 +412,7 @@ class Runner:
         if job_id:
             query["uuid"] = job_id[0]
             query["index"] = job_id[1]
+        print(query)
         job_data = self.job_controller.checkout_job(query=query)
         if not job_data:
             if not db_id and not job_id:
--- a/tests/db/cli/test_flow.py
+++ b/tests/db/cli/test_flow.py
@@ -1,7 +1,7 @@
 import pytest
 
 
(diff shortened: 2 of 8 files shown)


src/jobflow_remote/jobs/runner.py
```diff
@@ -412,6 +412,7 @@ def run_one_job(
         if job_id:
             query["uuid"] = job_id[0]
             query["index"] = job_id[1]
+        print(query)
```
davidwaroquiers: To be removed ?

tests/db/cli/test_flow.py
```diff
@@ -18,24 +21,60 @@ def test_flows_list(job_controller, four_jobs):
...
 
+    # don't confirm and verbose option
+    # only check the first characters of the uuid because it may be cut in the output
+    outputs = ["This operation will delete the following 1 Flow", four_jobs[1].uuid[:5]]
+    run_check_cli(
+        ["flow", "delete", "-fid", four_jobs[1].uuid, "-v"],
+        required_out=outputs,
+        cli_input="n",
+    )
+    assert job_controller.count_flows() == 1
```
davidwaroquiers, line 62: I was puzzled why it was giving one, as I thought four_jobs was just one flow so I went to look into the conftest to undestand that it's actually 2 flows. Maybe it would be good to just rename the fixture so that it is more clear ?

davidwaroquiers (approved):
All good to me! Just one print to be removed and maybe name change for a fixture to make things more clear ? (see comments)

tests/db/cli/test_flow.py
gpetretto: done

Issue #124, opened 2024-06-10
How to add metadata to `flows` docs?


janosh:
There currently appears to be no way of adding metadata like material IDs, formulas, structure provenance and the like to documents added into the `flows` collection.

i think it would make sense to e.g. look for a metadata attribute on a `jobflow.Flow` and if found, add that to the `flow_doc` prior to DB insertion in `JobController.add_flow`:

```py
if hasattr(flow, 'metadata'):
     flow_doc['metadata'] = flow.metadata
```

https://github.com/Matgenix/jobflow-remote/blob/967e7c512f230105b1a82c2227fb101d8d4acb3d/src/jobflow_remote/jobs/jobcontroller.py#L2591

happy to submit a PR for this if there's interest

gpetretto:
Hi @janosh, indeed there is no way of setting the metadata in the flow document, thanks for reporting that. 
I am not 100% convinced by the solution though. I can see a few potential minor issues:
1) the fact that jobflow's `Flow` object does not have a `metadata` could make this a bit hackish
2) There may be a confusion between this `metadata` attribute and the `update_metadata` method in Flow. One calling this method on a Flow may be tricked in thinking that also the Flow metadata is set.
3) In principle a user could run a code like this:
    ```python
    flow1 = Maker1().make()
    flow1.metadata = {"a": 1}
    flow2 = Maker2().make(flow1.output)
    flow = Flow([flow1, flow2])
    submit_flow(flow)
   ```
   and no Flow metadata will be added to the DB. Since only the top layer Flow is added in the DB it would be ambiguous what to do with those metadata.
4) If jobflow introduces a `metadata` attribute to `Flow` in the future this may break something.

Adding the metadata attribute to `Flow`now  could help, although will leave point 3 open and point 2 could still be tricky, but at least the behaviour of `update_metadata` could be documented with respect to the `Flow`'s `metadata`.

An alternative solution could be to to pass a `metadata` (or `flow_metadata`, to be more explicit) argument to `submit_flow`. This would solve the points above, but would probably feel a bit more clunky. 

What do you think?

davidwaroquiers:
[...]

PR #134, opened 2024-06-25
Fix legacy `ruff` errors and enable corresponding rules for future linting

janosh:
follow up to #122.

excludes all the inadvertent changes @ml-evs pointed out in #122 and i went through the rest more carefully this time. so much higher level of confidence that these changes are safe

### DIFF
--- a/src/jobflow_remote/config/base.py
+++ b/src/jobflow_remote/config/base.py
@@ -180,14 +180,16 @@ class WorkerBase(BaseModel):
     model_config = ConfigDict(extra="forbid")
 
     @field_validator("scheduler_type")
-    def check_scheduler_type(self, scheduler_type: str) -> str:
+    @classmethod
+    def check_scheduler_type(cls, scheduler_type: str) -> str:
         """Validator to set the default of scheduler_type."""
         if scheduler_type not in scheduler_mapping:
             raise ValueError(f"Unknown scheduler type {scheduler_type}")
         return scheduler_type
 
     @field_validator("work_dir")
-    def check_work_dir(self, v) -> Path:
+    @classmethod
+    def check_work_dir(cls, v) -> Path:
         if not v.is_absolute():
             raise ValueError("`work_dir` must be an absolute path")
         return v
@@ -461,7 +463,8 @@ class QueueConfig(BaseModel):
     )
 
     @field_validator("store")
-    def check_store(self, store: dict) -> dict:
+    @classmethod
+    def check_store(cls, store: dict) -> dict:
         """Check that the queue configuration could be converted to a Store."""
         if store:
             try:
@@ -569,7 +572,8 @@ class Project(BaseModel):
         return JobController.from_project(self)
 
     @field_validator("base_dir")
-    def check_base_dir(self, base_dir: str, info: ValidationInfo) -> str:
+    @classmethod
+    def check_base_dir(cls, base_dir: str, info: ValidationInfo) -> str:
         """Validator to set the default of base_dir based on the project name."""
         if not base_dir:
             from jobflow_remote import SETTINGS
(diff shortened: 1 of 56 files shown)


src/jobflow_remote/config/base.py
```diff
@@ -643,12 +625,8 @@ def check_jobstore(cls, jobstore: dict) -> dict:
...
-    A generic Exception related to the configuration
-    """
+    """A generic Exception related to the configuration."""
 
 
-class ProjectUndefined(ConfigError):
-    """
-    Exception raised if the Project has not been defined or could not be determined.
-    """
+class ProjectUndefinedError(ConfigError):
```
ml-evs, line 631: This is one of the few code behavior changes I can see, and think it's fine; probably no-one was relying on this exception name

src/jobflow_remote/remote/host/remote.py
```diff
@@ -353,18 +352,15 @@ def inter_handler(title, instructions, prompt_list):
     Used by Interactive AuthSource.
     """
     if title:
-        print(title.strip())
+        pass
     if instructions:
-        print(instructions.strip())
+        pass
```
ml-evs: I'm not sure that these particular print statements are useful, but probably this whole chunk can be removed if not

ml-evs (approved):
I've given this a once over and nothing looks untoward (just a couple of minor comments), thanks @janosh

src/jobflow_remote/remote/host/remote.py
janosh: oops, i missed this one. meant to revert all the `ruff` print removals. not sure if they're good to keep but that can be decided at a later time

janosh:
fwiw, i think fixing [`FBT001`](https://docs.astral.sh/ruff/rules/boolean-type-hint-positional-argument/) [`FBT002`](https://docs.astral.sh/ruff/rules/boolean-default-value-positional-argument/) would be good to do at some point

gpetretto:
Hi @janosh, thanks for opening this PR and addressing the issues raised previously by @ml-evs.
From a quick look it seems indeed that there are not many changes affecting the code, but since this has a large impact on the code base my plan would be:
* merge #132 and #127
* make a new release (as we it need relatively quickly).
* take a bit of time to look at this PR in detail.

I will do this in the next days.

janosh:
sounds good, will resolve the merge conflicts here.

i was going to take a swing at addressing https://github.com/Matgenix/jobflow-remote/issues/124 once this PR is merged (to avoid further conflicts)

src/jobflow_remote/remote/data.py
```diff
@@ -96,22 +93,22 @@ def get_single_store(
...
         store_type["path"] = file_path
         store = MontyDecoder().process_decoded(store_type)
         if not isinstance(store, Store):
-            raise ValueError(
+            raise TypeError(
                 f"Could not instantiate a proper store from remote config dict {store_type}"
             )
-    else:
-        raise ValueError(f"remote store type not supported: {store_type}")
+        return None
```
gpetretto: This was a bug in the original code... Can you make this to `return store`? Otherwise I am afraid I will forget to fix it afterwards.

src/jobflow_remote/utils/db.py
```diff
@@ -149,25 +150,19 @@ def __init__(
...
-    def acquire(self):
-        """
-        Acquire the lock
-        """
+    def acquire(self) -> None:
+        """Acquire the lock."""
         # Set the lock expiration time
         now = datetime.utcnow()
-        db_filter = copy.deepcopy(self.filter)
+        db_filter = copy.deepcopy(dict(self.filter))
```
gpetretto, line 165: I don't think it has a big impact, but why adding a `dict` here?

janosh:
@gpetretto bump

gpetretto (commented):
Hi @janosh, sorry, I had already done my review, but forgot to finally publish it. I just have one question and one bug fix that was already there and could be fixed directly here.

src/jobflow_remote/utils/db.py
janosh: this fixes a `mypy` issue:
```
src/jobflow_remote/utils/db.py:176: error: "Mapping[str, Any]" has no attribute "update"  [attr-defined]
```

can also add a `type: ignore` but really `mypy` is right here, this could cause a run time error

src/jobflow_remote/remote/data.py
janosh: will do

src/jobflow_remote/utils/db.py
gpetretto: The question was mainly for my understanding of the issue. Thanks for the explanation. I think it is fine like this.

gpetretto (approved):

gpetretto:
Thanks!

PR #142, opened 2024-07-09
CLI add `--pid` option to `jf job info`

janosh:
to query job details by `slurm` job ID

i added a test and verified this to work on our cluster

### DIFF
--- a/src/jobflow_remote/cli/job.py
+++ b/src/jobflow_remote/cli/job.py
@@ -191,13 +191,13 @@ def job_info(
     verbosity: verbosity_opt = 0,
 ) -> None:
     """Detailed information on a specific job."""
-    if pid is not None and (job_db_id is not None or job_index is not None):
+    if pid is not None and job_db_id is not None:
         raise typer.BadParameter("Cannot specify both job ID/index and process ID")
 
     with loading_spinner():
         jc = get_job_controller()
 
-        if job_db_id is not None or job_index is not None:
+        if job_db_id is not None:
             if verbosity > 0:
                 job_data = jc.get_job_doc(
                     job_id=job_db_id,
--- a/src/jobflow_remote/cli/job.py
+++ b/src/jobflow_remote/cli/job.py
@@ -198,15 +198,19 @@ def job_info(
         jc = get_job_controller()
 
         if job_db_id is not None:
+            db_id, job_id = get_job_db_ids(job_db_id, job_index)
+
             if verbosity > 0:
                 job_data = jc.get_job_doc(
-                    job_id=job_db_id,
+                    job_id=job_id,
                     job_index=job_index,
+                    db_id=db_id,
                 )
             else:
                 job_data = jc.get_job_info(
-                    job_id=job_db_id,
+                    job_id=job_id,
                     job_index=job_index,
+                    db_id=db_id,
                 )
         elif pid is not None:
             job_data = jc.get_job_info_by_pid(pid)
--- a/src/jobflow_remote/cli/job.py
+++ b/src/jobflow_remote/cli/job.py
@@ -165,7 +165,7 @@ def job_info(
     job_db_id: job_db_id_arg = None,
     job_index: job_index_arg = None,
     pid: Annotated[
-        int,
+        str,
         typer.Option(
             "--pid",
             help="The process ID of the job in the queue system (e.g. Slurm job ID)",
--- a/src/jobflow_remote/cli/job.py
+++ b/src/jobflow_remote/cli/job.py
@@ -162,8 +162,15 @@ def jobs_list(
 
 @app_job.command(name="info")
 def job_info(
-    job_db_id: job_db_id_arg,
+    job_db_id: job_db_id_arg = None,
     job_index: job_index_arg = None,
+    pid: Annotated[
+        int,
+        typer.Option(
+            "--pid",
+            help="The process ID of the job in the queue system (e.g. Slurm job ID)",
+        ),
+    ] = None,
     with_error: Annotated[
         bool,
         typer.Option(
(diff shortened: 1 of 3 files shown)


src/jobflow_remote/cli/job.py
```diff
@@ -162,8 +162,15 @@ def jobs_list(
 
 @app_job.command(name="info")
 def job_info(
-    job_db_id: job_db_id_arg,
+    job_db_id: job_db_id_arg = None,
     job_index: job_index_arg = None,
+    pid: Annotated[
+        int,
```
gpetretto: In the DB the `process_id` is stored as a string because I thought that I cannot really guarantee that the id would be necessarily an integer. I think it would make sense to have a string here as well. In any case it should not alter the functionality

src/jobflow_remote/cli/job.py
```diff
@@ -182,30 +189,28 @@ def job_info(
...
-    """Detail information on a specific job."""
-    db_id, job_id = get_job_db_ids(job_db_id, job_index)
-
-    if with_error:
-        out_console.print(
-            "The --with-error option is deprecated and not needed anymore to show error messages",
-            style="yellow",
-        )
+    """Detailed information on a specific job."""
+    if pid is not None and (job_db_id is not None or job_index is not None):
```
gpetretto: The check on the `job_index` is redundant. Since they are sequential arguments, It cannot happen that `job_index` is defined if `job_db_id` is not.

src/jobflow_remote/cli/job.py
```diff
@@ -182,30 +189,28 @@ def job_info(
...
     with loading_spinner():
         jc = get_job_controller()
 
-        if verbosity > 0:
-            job_data = jc.get_job_doc(
-                job_id=job_id,
-                job_index=job_index,
-                db_id=db_id,
-            )
+        if job_db_id is not None or job_index is not None:
```
gpetretto: Same as above, the check on `job_index` is redundant.

src/jobflow_remote/cli/job.py
```diff
@@ -182,30 +189,28 @@ def job_info(
...
+            if verbosity > 0:
+                job_data = jc.get_job_doc(
```
gpetretto, line 204: Here I think you missed the logic of the the code. The input for `job_db_id` can be either the `db_id` or the `uuid` of the job. This is possible thanks to the call `db_id, job_id = get_job_db_ids(job_db_id, job_index)`. The function decides which it is and returns the proper `db_id` and `job_id` values (one of which is None). Those should then be passed to the `get_job_doc` and `get_job_info`. Can you restore the call to the `get_job_db_ids` function and pass both the ids?

src/jobflow_remote/cli/job.py
```diff
@@ -182,30 +189,28 @@ def job_info(
...
+                    job_id=job_db_id,
+                    job_index=job_index,
+                )
+            else:
+                job_data = jc.get_job_info(
+                    job_id=job_db_id,
+                    job_index=job_index,
+                )
+        elif pid is not None:
+            job_data = jc.get_job_info_by_pid(pid)
```
gpetretto, line 216: This means that if `--pid` is used the `-v` option is ignored. I don't think it is worth adding a `get_job_doc_by_pid` method to `JobController` as I suppose this would not be a very common occurrence. Can you maybe mention that the `-v` option would be ignored in the `help` of the `--pid` option?

tests/db/cli/test_job.py
```diff
@@ -51,6 +51,47 @@ def test_job_info(job_controller, two_flows_four_jobs) -> None:
...
+
+    # Test that we can't use --pid with other job identification options
+    run_check_cli(
+        ["job", "info", "1", "--pid", str(test_pid)],
+        error=True,
+        required_out="Cannot specify both job ID/index and process ID",
+    )
+
+    # Test with verbosity
+    run_check_cli(
```
gpetretto: not sure what it is testing here, since the `-vvv` option is ineffective with  '--pid'.

tests/db/cli/test_job.py
```diff
@@ -51,6 +51,47 @@ def test_job_info(job_controller, two_flows_four_jobs) -> None:
...
+        ["job", "info", "--pid", str(test_pid), "-vvv"],
+        required_out=[f"'step_attempts': 0, 'process_id': '{test_pid}'"],
+    )
+
+    run_check_cli(
+        ["job", "info", "--pidd", str(test_pid)],
```
gpetretto: I see this is making the tests fail. 
I think this check is more testing Typer itself rather then the `--pid` option. I think it can be removed.

gpetretto (changes requested):
Hi @janosh,
thanks adding a new functionality! 
I made a few comments, can you check them and update the code accordingly.

src/jobflow_remote/cli/job.py
janosh: > I don't think it is worth adding a get_job_doc_by_pid method to JobController as I suppose this would not be a very common occurrence.

i'll slightly push back on this. i think it makes sense to have that function even if just for consistency of impementation when comparing with [`JobController.get_job_info`](https://github.com/janosh/jobflow-remote/blob/ddded83bb48574b071e13e52a928c5423c283987/src/jobflow_remote/jobs/jobcontroller.py#L571). it's nice to be able to directly translate a CLI command to the Python API which would not be possible if we remove `JobController.get_job_info_by_pid`. let me know if you prefer it be removed anyway

src/jobflow_remote/cli/job.py
janosh: apologies, this was hastily done. didn't actually mean to change this code

tests/db/cli/test_job.py
janosh: good point!

janosh:
@gpetretto this is ready to go from my side

src/jobflow_remote/cli/job.py
gpetretto: I agree that it is nice to have this kind of mapping. I am just afraid that in the future this could lead to a lot of similar methods based on different query options. However, since this is not the case at the moment I think it can stay like this and be reevaluated if it becomes an issue in the future.

gpetretto (approved):
Thanks!

PR #150, opened 2024-07-17
WIP: Check only one runner

### LINKED ISSUE #140: Check there is not already a runner running

davidwaroquiers:
This PR adds info to the auxiliary collection when the runner is started. The idea would be to warn the user that there may be another runner running already, in which case bad things could happen.

Currently just inserts the info when starting the runner and removes it when stopping it.

Still to be done:
- Deal with the locked document (if two people try to start the runner at the exact same time, guess it would not be very common but still)
- Deal with the specific goal of this PR: what do we do when this happens
- Deal with when the document with the runner info in the auxiliary collection is not present


davidwaroquiers:
Hi @gpetretto 

Still a few things to be done but I think before going on, it would be better to have the expert's eye on this ;)

src/jobflow_remote/cli/admin.py
```diff
@@ -230,3 +266,34 @@ def unlock_flow(
...
+
+        num_docs, num_unlocked = jc.unlock_runner()
+
+    if num_docs == 0:
+        out_console.print(
+            "No runner document... "
+            "Consider upgrading your database using 'jf admin upgrade'"
+        )
+    else:
+        if num_docs > 1:
```
gpetretto: I wonder if we really need to consider this case. For example, what if there are two job documents with the same uuid and index? There are potentially a lot of cases that should not happen but you cannot possibly check all of them. Is there any reason why this is special and needs to be cheked? 
I rather wonder if there is a way to enforce this at the db level.

src/jobflow_remote/jobs/daemon.py
```diff
@@ -169,6 +177,12 @@ def sock_filepath(self) -> Path:
             raise DaemonError(msg)
         return path
 
+    @lazy_property
+    def job_controller(self):
+        if self._job_controller is None:
+            return JobController.from_project(self.project)
+        return self._job_controller
```
gpetretto, line 260: This is duplicating the caching functionality. The `lazy_property` already does this for you. Either just `return JobController.from_project(self.project)` or remove `lazy_property`

src/jobflow_remote/jobs/daemon.py
```diff
@@ -405,61 +419,102 @@ def start(
...
-                connect_interactive=connect_interactive,
-            )
-            # else:
-            #     error = self.start_processes()
-        elif status == DaemonStatus.STOPPING:
-            error = "Daemon process are stopping. Cannot start."
-        else:
-            error = f"Daemon status {status} could not be handled"
+        db_filter = {"running_runner": {"$exists": True}}
+        with self.job_controller.lock_auxiliary(filter=db_filter) as lock:
```
gpetretto: If understand the code properly, here you need to pass `get_locked_doc=True`. Otherwise if the document is present but locked `lock.unavailable_document` will always be empty

src/jobflow_remote/jobs/daemon.py
```diff
@@ -405,61 +419,102 @@ def start(
...
+                    f"- user: {drr['user']}"
+                )
 
-        if error is not None:
-            if raise_on_error:
-                raise DaemonError(error)
-            logger.error(error)
-            return False
-        return True
+            elif status == DaemonStatus.RUNNING:
```
gpetretto: I wonder if it may be worth doing this check as a first point to avoid interacting with the DB if the runner is already running on this machine.
Also, if I undestand correctly, if a runner is normally already running and you try again the `jf runner run` you will get the error because the document in the db is not empty, and you will never get to this point. Is that right?

src/jobflow_remote/jobs/daemon.py
```diff
@@ -405,61 +419,102 @@ def start(
...
-        if status in (
-            DaemonStatus.STOPPED,
-            DaemonStatus.STOPPING,
-            DaemonStatus.SHUT_DOWN,
-        ):
-            return True
+        db_filter = {"running_runner": {"$exists": True}}
+        with self.job_controller.lock_auxiliary(filter=db_filter) as lock:
+            if lock is None:
+                # print('Handle case where the document is not present!')
```
gpetretto: To avoid repeating the same pieces of code, you could add a method to `DaemonManager` that is a context manager (e.g. `lock_runner_doc`) that contains all the logic that needs to be checked. Will make it easier to update if something changes.

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -3594,3 +3763,72 @@ def get_flow_leafs(job_docs: list[dict]) -> list[dict]:
...
+        _update_version_information(self.job_controller.auxiliary)
+        logger.info(
+            f"Finished upgrade of the database from version {previous_version} to version {this_version}"
+        )
+        return True
+
+    def _upgrade_to_0_1_3(self):
+        self.job_controller.auxiliary.find_one_and_update(
+            filter={"running_runner": {"$exists": True}},
+            update={"$set": {"running_runner": None}},
```
gpetretto: what if the runner is running? not sure there is anything that can be done here...

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -2357,9 +2384,125 @@ def reset(self, reset_output: bool = False, max_limit: int = 25) -> bool:
...
+            from the current environment. Explicitly setting it is mainly for
+            testing purposes.
+
+        Returns
+        -------
+        bool
+            True if the database was upgraded, False otherwise.
+        """
+        this_version = this_version or jobflow_remote.__version__
+        environment_doc = self.auxiliary.find_one(
```
gpetretto: shouldn't this be inside the `if previous_version is None`?

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -2357,9 +2384,125 @@ def reset(self, reset_output: bool = False, max_limit: int = 25) -> bool:
...
+        if environment_doc is None:
+            msg = (
+                "No information about jobflow version in the database.\n"
+                "The database is likely from before version 0.1.3 of jobflow-remote.\n"
+            )
+            logger.warning(msg)
+            return msg
+        previous_jobflow_version = environment_doc["jobflow_version"]
+        current_jobflow_version = jobflow.__version__
+        if jobflow.__version__ != previous_jobflow_version:
```
gpetretto: This check is not really clear. Isn't the upgrade procedure supposed to be applied if the version is different?

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -2357,9 +2384,125 @@ def reset(self, reset_output: bool = False, max_limit: int = 25) -> bool:
...
+        )
+        if environment_doc is None:
+            msg = (
+                "No information about environment (all packages versions) in the database.\n"
+                "The database is likely from before version 0.1.3 of jobflow-remote.\n"
+            )
+            logger.warning(msg)
+            return msg
+        previous_package_versions = environment_doc["full_environment"]
+        installed_packages = importlib.metadata.distributions()
```
gpetretto: Is all this check really needed? If anyone installed a newer version pip should have already taken care of updating the packges to the correct versions and add new packages, if needed. This will be useful only if someone forces pip to install the package without the dependencies. 
I don't see a real advantage of repeating the check in the code.
After our discussion, maybe this is if the version is manually specified for the tests? I will leave the comment anyway.

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -3172,6 +3315,12 @@ def get_flow_jobs_data(
 
         return list(self.flows.aggregate(pipeline))
 
+    def get_running_runner(self):
+        """Get the running runner information from the auxiliary collection."""
+        return self.auxiliary.find_one({"running_runner": {"$exists": True}})[
```
gpetretto: if there is no document this will raise an error. And this could still happen if one does not upgrade, right?. Maybe just use `get("running_runner")`

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -3594,3 +3763,72 @@ def get_flow_leafs(job_docs: list[dict]) -> list[dict]:
                 d.pop(parent_id, None)
 
     return list(d.values())
+
+
+class Upgrader:
```
gpetretto: We did not discuss this before, but here all the focus is to have an upgrade for the DB. However I hope we will have few updates of the DB structure. On the other hand, changes to the project configuration files seem more likely. 

I would consider reverting the logic: instead of calling `JobController.upgrade`, that then call `Upgrade.upgrade`, move the `Upgrader` to its own module and call `Upgrader.ugrade`, that can call `JobController.upgrade_db`. In this way the upgrader may also act to upgrade old configuration files that may not be compatible anymore. What do you think?
Or maybe this could be a side function? Like the `Upgrader` has an `upgrade_db` option that upgrades the DB for the currently loaded project and `upgrade_config` to upgrade a configuration file that is not the currently selected one(and it cannot be the currently selected one, because it may be incompatible with the `Project` model, and thus cannot be loaded at all)?
It may be worth to think this in more details.

tests/db/cli/test_admin.py
```diff
@@ -77,3 +77,86 @@ def test_unlock_flow(job_controller, one_job) -> None:
         required_out="No data matching the request",
         error=True,
     )
+
+
+def test_running_runner(job_controller) -> None:
```
gpetretto: shouldn't this be called `test_unlock_runner`?

src/jobflow_remote/cli/admin.py
davidwaroquiers: True, I was just worried of my own upgrade implementation, and that would be some fail safe, but probably it's not worth it. I'll remove.

src/jobflow_remote/jobs/daemon.py
davidwaroquiers: Actually not because the job_controller is optional in the init. If it has been passed to init, it should be that one, if not then it should be instantiated with from_project. Now maybe it's not useful to be have it in the init ... we could maybe always use the property. What do you think ?

src/jobflow_remote/jobs/daemon.py
davidwaroquiers: Correct. Fixed.

src/jobflow_remote/jobs/daemon.py
davidwaroquiers: Indeed, so inverting these two first ifs right ? Yes if a runner is already running it will not get to this point

src/jobflow_remote/jobs/daemon.py
davidwaroquiers: Indeed, I was planning to.

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: Correct, but the upgrade is supposed to be done in the cli and the cli checks (locally, maybe it could/should check in the db also ?) if it is running before doing the upgrade (in the same way a reset could be done in python while a runner is running for example).

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: Here I check jobflow version, not jobflow-remote. We may upgrade jobflow-remote but not jobflow.

src/jobflow_remote/jobs/daemon.py
gpetretto: ok, then I think you can remove the lazy_property, becase in any case you are already reusing the one that is in `self._job_controller`

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: ok, fine with me

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: ah, I missed that, but why is that a problem for the upgrade procedure? Jobflow does not define the content of the DB

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: It's more informative. It is deactivated by default btw. I was considering having this check for production environments for example, just to make sure other packages are also in line with what is needed.

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: Maybe I would put get("running_runner", "NO_DOCUMENT") or something like that to distinguish from no running runner ? What do you think ?

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: Yes, after discussion, I will do that :)

src/jobflow_remote/jobs/daemon.py
davidwaroquiers: Ok then in that case I'd just set the _job_controller in the if and return it after.

def job_controller(self):
        if self._job_controller is None:
            self._job_controller = JobController.from_project(self.project)
        return self._job_controller

Like this

src/jobflow_remote/jobs/daemon.py
davidwaroquiers: Also, this makes me think that I did the check for jf runner start but not for jf runner run ... I'll add this also

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: Done.

src/jobflow_remote/cli/admin.py
```diff
@@ -25,13 +25,51 @@
...
+    test_version_upgrade: Annotated[
+        Optional[str],
+        typer.Option(
+            help="Explicitly sets the target version for upgrade. For testing purposes only",
+            hidden=True,
+        ),
+    ] = None,
+) -> None:
+    """
+    Upgrade the jobflow database.
```
gpetretto, line 73: Maybe clarify what the "upgrade" means here: it can/should be run when updgrading the jobflow-remote version.

Also suggesting to create a backup of the db before proceeding? Not sure if needed or possible at all... Maybe we should have a `backup` command as well? Using MongoDB to dump the collections to json files?

src/jobflow_remote/jobs/daemon.py
```diff
@@ -405,61 +418,104 @@ def start(
...
-        status = self.check_status()
-        if status in (
-            DaemonStatus.STOPPED,
-            DaemonStatus.STOPPING,
-            DaemonStatus.SHUT_DOWN,
-        ):
-            return True
+        db_filter = {"running_runner": {"$exists": True}}
+        with self.job_controller.lock_auxiliary(filter=db_filter) as lock:
+            if lock is None:
```
gpetretto: Probably here you meant to check if lock.locked_document?

src/jobflow_remote/jobs/daemon.py
```diff
@@ -405,61 +418,104 @@ def start(
...
+                # print('Handle case where the document is not present!')
+                pass
+            if lock.is_locked:
```
gpetretto: if it gets here lock.is_locked is probably always True. Need to check unavailable_document (and locking with `get_locked_doc=True`?

src/jobflow_remote/jobs/daemon.py
```diff
@@ -405,61 +418,104 @@ def start(
...
+                # print('Handle case where the document is locked')
+                pass
+            status = self.check_status()
+            if status in (
+                DaemonStatus.STOPPED,
+                DaemonStatus.STOPPING,
+                DaemonStatus.SHUT_DOWN,
+            ):
+                lock.update_on_release = {"$set": {"running_runner": None}}
+                return True
```
gpetretto, line 714: Could there be a fault in the logic here? If the runner is started on a machine and then run "stop" from another, the DaemonStatus will be SHUT_DOWN, but I suppose the `running_runner` should not be set to None in that case.

src/jobflow_remote/jobs/daemon.py
```diff
@@ -483,52 +539,62 @@ def _verify_call_result(
...
+        with self.job_controller.lock_auxiliary(filter=db_filter) as lock:
+            # If the daemon is shutting down supervisord may not be able to identify
+            # the state. Try proceeding in that case, since we really want to kill
+            # the process
+            status = None
+            try:
+                status = self.check_status()
+                if status == DaemonStatus.SHUT_DOWN:
+                    logger.info("supervisord is not running. No process is running")
+                    lock.update_on_release = {"$set": {"running_runner": None}}
```
gpetretto, line 786: As in the comment above, if this is executed on a machine where the runner is SHUT_DOWN the `running_runner` will be set to None even if it is actually running o another machine.

src/jobflow_remote/jobs/daemon.py
```diff
@@ -483,52 +539,62 @@ def _verify_call_result(
...
+                    return True
+                if status == DaemonStatus.STOPPED:
+                    logger.info("Processes are already stopped.")
+                    lock.update_on_release = {"$set": {"running_runner": None}}
```
gpetretto, line 790: Same as above?

src/jobflow_remote/jobs/daemon.py
```diff
@@ -483,52 +539,62 @@ def _verify_call_result(
...
     def shut_down(self, raise_on_error: bool = False) -> bool:
-        status = self.check_status()
-        if status == DaemonStatus.SHUT_DOWN:
-            logger.info("supervisord is already shut down.")
+        db_filter = {"running_runner": {"$exists": True}}
+        with self.job_controller.lock_auxiliary(filter=db_filter) as lock:
+            status = self.check_status()
+            if status == DaemonStatus.SHUT_DOWN:
+                logger.info("supervisord is already shut down.")
+                lock.update_on_release = {"$set": {"running_runner": None}}
```
gpetretto, line 839: Same as above?

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -3172,6 +3201,13 @@ def get_flow_jobs_data(
 
         return list(self.flows.aggregate(pipeline))
 
+    def get_running_runner(self):
```
gpetretto: Specify the return type.

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -3570,10 +3626,128 @@ def remove_batch_process(self, process_id: str, worker: str) -> dict:
...
+            logger.warning(msg)
+            return msg
+        previous_jobflow_version = environment_doc["jobflow_version"]
+        current_jobflow_version = jobflow.__version__
+        if jobflow.__version__ != previous_jobflow_version:
+            msg = (
+                f"The previous jobflow version ({previous_jobflow_version}) "
+                f"is not the same as the current one ({current_jobflow_version}).\n"
+            )
+            logger.warning(msg)
```
gpetretto: I think this message will not be really clear. It will be called from the `jf admin upgrade` but the user will not provide any "previous version" or "current version". Maybe it should say something like "the jobflow version registered in the DB is different from the currently installed one". 
Moreover this will likely not be an issue 99% of the cases, so maybe it should be clear that this may not be an issue.

One last point: I think that this should either return the message or log the warning, otherwise in the CLI the warning will appear twice (is that correct?). And the same for the msg above.

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -3570,10 +3626,128 @@ def remove_batch_process(self, process_id: str, worker: str) -> dict:
...
+                        )
+                    ]
+                )
+                msg += (
+                    f"The following {len(missing_packages)} packages "
+                    f"have different versions:\n "
+                    f"{different_str}\n"
+                )
+            logger.warning(msg)
+            return msg
```
gpetretto, line 4613: Again this differences are going to be harmless in 99% of the cases. Maybe should be clarified?
And maybe remove the logger.warning?

Also why is the jobflow version checked in a different function?
And is the difference in the version of jobflow-remote also shown here?

src/jobflow_remote/cli/admin.py
```diff
@@ -25,13 +25,51 @@
...
+    if not force:
+        jobflow_check = jc.upgrade_check_jobflow()
+        full_check = jc.upgrade_full_check()
+        if jobflow_check or full_check:
+            text = Text.from_markup(
+                "[bold]Possible issues in upgrade of the database.[/bold]\n"
+                + jobflow_check
+                + full_check
+                + "Carefully check this is ok and use 'jf admin upgrade --force'"
+            )
```
gpetretto, line 99: I already raised a doubt in the previous review, but I still have doubt about this part of the code. In the previous message you said that this check is disabled by default, but to me it will always run if the `-force` option will not be set.
Moreover the message pushes the user to check the upgrades but I doubt any user will have any clue to determine if this could be an issue. It will be really difficult to say if the difference in some package will impact the upgrade of the jobflow-remote DB/configuration. And even if there is any kind of issue, this alaready present in the fact that the packages in the python environment have already been changed.
I would find more useful to have instead a sort of dry-run here that will tell which operations are going to be performed. At least one can have a better idea of what to backup and what is going to happen. Would that be feasible?

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -3570,10 +3626,128 @@ def remove_batch_process(self, process_id: str, worker: str) -> dict:
             upsert=True,
         )
 
+    def upgrade_check_jobflow(self):
```
gpetretto, line 4527: Why does jobflow has its own specific function? Can't this be in `upgrade_full_check` like the other dependencies?

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -3570,10 +3626,128 @@ def remove_batch_process(self, process_id: str, worker: str) -> dict:
...
+                msg += (
+                    f"The following {len(missing_packages)} packages "
+                    f"have different versions:\n "
+                    f"{different_str}\n"
+                )
+            logger.warning(msg)
+            return msg
+        return ""
+
+    def upgrade(self, previous_version, this_version):
```
gpetretto: Specify the return type. Is there any case where this returns `False`?

src/jobflow_remote/cli/admin.py
```diff
@@ -25,13 +25,51 @@
...
+    """
+    Upgrade the jobflow database.
+    WARNING: can modify all the data. Previous version could not be retrieved anymore.
+    """
+    check_stopped_runner(error=True)
+
+    jc = get_job_controller()
+    if not force:
+        jobflow_check = jc.upgrade_check_jobflow()
+        full_check = jc.upgrade_full_check()
```
gpetretto, line 93: maybe the checks should go through the upgrader, rather than directly to the JobController? Or since these are just check on the db it goes directly through the Jobcontroller?

tests/db/jobs/test_daemon.py
```diff
@@ -135,3 +153,54 @@ def test_kill_one_process(job_controller, daemon_manager, caplog) -> None:
...
+
+def test_start_two_runners(daemon_manager, mocker) -> None:
+    from jobflow_remote.jobs.daemon import DaemonError, DaemonStatus
+
+    assert daemon_manager.check_status() == DaemonStatus.SHUT_DOWN
+    assert daemon_manager.start(raise_on_error=True, single=False)
+    _wait_daemon_started(daemon_manager)
+
+    mocker.patch.object(
+        daemon_manager, "check_status", return_value="FAKE_NOT_RUNNING_LOCALLY"
```
gpetretto: wouldn't it be better to return a proper DaemonStatus value? e.g. SHUT_DOWN?

gpetretto (commented):
I have added some more comments to the latest changes

src/jobflow_remote/cli/admin.py
davidwaroquiers: Sure, I'll clarify what it means.
About the backup command, good idea. I propose we put that for a different PR. I'll open an issue for it.

src/jobflow_remote/jobs/daemon.py
davidwaroquiers: No I think it was indeed lock is None, in the case where the document is not found, the lock itself returns None.

gpetretto: In principle lock should always be the instance of `MongoLock`:
https://github.com/Matgenix/jobflow-remote/blob/45fd5ec40d530d0c4f3dc549cc6055769202b455/src/jobflow_remote/utils/db.py#L266
If not I would say it is a bug.

davidwaroquiers: Hmm, maybe I was not awake enough. I'll check and see if it should indeed be lock.locked_document (probably should).

src/jobflow_remote/jobs/daemon.py
davidwaroquiers: Indeed ... I guess it should only set it to None if the status was not stopped before. Or I should make a matching between the information in the running runner and the information about the runner on the machine ?

davidwaroquiers: Or actually just not do anything ...

gpetretto: I don't know. What happens if I have only one machine that runs the Runner (which is likely 99% of the use cases) . It is started but the machine goes down abruptly (e.g. power cut). The document will not be updated but the deamon is shut down. What happens then? How am I supposed to deal with that situation? I likely don't want to go through an annoying procedure of having to reset the document manually before I can restart the db. Could it make sense that there is a global check that whenever the runner is not active and the document matches the local information it will clean up the content of the document?

Acutally this is also a linked to how the `start` method works. Currently in a scenario as the one I just described I would never be able to start the runner again, right?

tests/db/conftest.py
```diff
@@ -230,3 +231,27 @@ def two_flows_four_jobs(random_project_name):
     submit_flow(flow2, worker="test_local_worker")
 
     return [flow, flow2]
+
+
+@pytest.fixture()
+def version_candidate():
+    """Next version candidate. Used for tests of upgrade methods."""
+    return "0.1.3"
```
davidwaroquiers: Is this correct ? Next version candidate should be 0.1.5 no ?

tests/db/conftest.py
```diff
@@ -230,3 +231,27 @@ def two_flows_four_jobs(random_project_name):
...
+
+
+@pytest.fixture(scope="session")
+def wait_daemon_started():
+    def _wait_daemon_started(daemon_manager, max_wait: int = 10) -> bool:
```
davidwaroquiers: Is max_wait really tunable here ?

tests/db/cli/test_admin.py
```diff
@@ -99,6 +99,82 @@ def test_unlock_flow(job_controller, one_job) -> None:
...
 
 
+def test_unlock_runner(job_controller) -> None:
+    from jobflow_remote.testing.cli import run_check_cli
+
+    with job_controller.lock_auxiliary(filter={"running_runner": {"$exists": True}}):
+        run_check_cli(
+            ["admin", "unlock-runner"],
+            required_out="The runner document has been unlocked",
+        )
```
davidwaroquiers: Should it check the state of the auxiliary document just before run_check_cli ? i.e. manually getting the running_runner document and checking the lock keys are there ? Then checking it just after that it's not there anymore ?

src/jobflow_remote/cli/admin.py
```diff
@@ -30,13 +32,92 @@
...
+            "--check-env",
+            "-ce",
+            help="Report the changes in the package installed in the python environment before "
+            "proceeding with the upgrade",
+        ),
+    ] = False,
+) -> None:
+    """
+    Upgrade the jobflow database.
+    WARNING: can modify all the data. Previous version could not be retrieved anymore.
```
davidwaroquiers:
```suggestion
    WARNING: can modify all the data. Previous version cannot be retrieved anymore.
```

src/jobflow_remote/cli/admin.py
```diff
@@ -30,13 +32,92 @@
...
+            )
+            out_console.print(text)
+
+    upgrader = DatabaseUpgrader(jc)
+    if not no_dry_run:
+        target_version = test_version_upgrade or upgrader.current_version
+        actions = upgrader.dry_run(target_version=test_version_upgrade)
+        if not actions:
+            out_console.print(
+                f"No actions will be required for upgrading to version {target_version}"
```
davidwaroquiers, line 106: Can this happen ? It should at least set the new version in the db no ?

src/jobflow_remote/jobs/daemon.py
```diff
@@ -380,6 +567,14 @@ def start_supervisord(
         return None
 
     def start_processes(self) -> str | None:
+        """
+        Start all the processes of the daemon.
```
davidwaroquiers: Maybe add a Note to mention that the supervisord process should be up and running ?

src/jobflow_remote/jobs/daemon.py
```diff
@@ -355,6 +519,29 @@ def start_supervisord(
         nodaemon: bool = False,
         connect_interactive: bool = False,
     ) -> str | None:
+        """
+        Start the supervisord daemon.
```
davidwaroquiers: Correct me if I'm wrong. This starts the supervisord process + the processes right ? I got confused a bit about this in the start method below when the status of the daemon is SHUTDOWN, it does self.start_supervisord but it does not do start_processes afterwards, so I guess start_supervisord starts supervisord and it's processes from the config.

davidwaroquiers: In that case, maybe add a Note to explain that the processes are started (and that there is no need to do self.start_processes). It was not completely clear to me but maybe it's because I'm not fully aware of how supervisord works exactly.

src/jobflow_remote/jobs/daemon.py
```diff
@@ -380,6 +567,14 @@ def start_supervisord(
...
+        Start all the processes of the daemon.
+
+        Returns
+        -------
+        str | None
+            An error message if not all the processes started correctly, otherwise None.
+        """
         interface = self.get_interface()
         result = interface.supervisor.startAllProcesses()
         if not result:
```
davidwaroquiers, line 581: This is a comment for a line below (for which github does not allow me to do a comment in this PR). I am wondering if and how the very last line of this method can be reached.

src/jobflow_remote/jobs/daemon.py
```diff
@@ -405,65 +600,156 @@ def start(
...
+                logger.error(doc_error)
+                return False
+            status = self.check_status()
+            if status in (
+                DaemonStatus.STOPPED,
+                DaemonStatus.STOPPING,
+                DaemonStatus.SHUT_DOWN,
+            ):
+                lock.update_on_release = {"$set": {"running_runner": None}}
+                return True
```
davidwaroquiers, line 714: When it's "STOPPING", are we sure that all processes will indeed be stopped upon releasing the lock ? Could there be any problem here ?

davidwaroquiers (commented):
I think it's all good. I did not test myself the functionality though. I would say it might be something to mention in the next release that if people encounter "weird things" when starting/stopping, they should open an issue. What do you think ?
I can't approve my own PR even if it has more become yours ;-)

tests/db/conftest.py
gpetretto: It is wrong, but it is never used, as the tests for the upgrade specify manually the target version. Maybe should be removed?

tests/db/conftest.py
gpetretto: Why not? The fixture returns the function. Inside the test the function is called and in principle the `max_wait` could be set.

tests/db/conftest.py
gpetretto: I removed it

tests/db/cli/test_admin.py
gpetretto: done

src/jobflow_remote/cli/admin.py
gpetretto: Indeed it should not happen, but I thought of leaving the "if" just in case things will change in the future. Anyway I can remove it if you think it better.

src/jobflow_remote/jobs/daemon.py
gpetretto: done

src/jobflow_remote/jobs/daemon.py
gpetretto: done

src/jobflow_remote/jobs/daemon.py
gpetretto: I revised it

src/jobflow_remote/jobs/daemon.py
gpetretto: Indeed it cannot be guaranteed. However, following what we discussed about the fact that also when running the `jf runner stop/shutdown` it cannot be guaranteed that the runner will be actually stopped immediately, I think it should be fine to leave it this way also for the `STOPPING` state.

tests/db/conftest.py
davidwaroquiers: Yes, my mistake..

src/jobflow_remote/cli/admin.py
davidwaroquiers: Ok lets keep it

src/jobflow_remote/jobs/daemon.py
davidwaroquiers: Yes, ok

doc/source/user/runner.rst
```diff
@@ -0,0 +1,217 @@
...
+=====
+
+As explained in the :ref:`setup options` section, and exemplified in the figure below,
+the Runner process must have **access to all required resources** as specified in the project
+configuration. In particular all the workers, the MongoDB database defined in the
+``queue`` section and the output ``JobStore``.
+
+.. image:: ../_static/img/daemon_schema.svg
+   :width: 50%
+   :alt: All-in-one configuration
```
davidwaroquiers: Is this alternate text correct?

doc/source/user/runner.rst
```diff
@@ -0,0 +1,217 @@
...
+The number of runner processes can only be managed at start time. The ``--single``
+option will run all the actions described in the previous section in a single
+process, instead that in multiple ones, which is the default. The ``--transfer``
+and ``--complete`` options allow to increase the number of processes dedicated to
+the steps 3 and 4.
+
+.. warning::
+    The ``Runner`` reads the project configurations when each of the processes is
+    started and does not attempt to refresh them during the execution. Whenever
+    changing
```
davidwaroquiers: Missing rest of the sentence here

doc/source/user/runner.rst
```diff
@@ -0,0 +1,217 @@
...
+    jf runner stop
+
+relies on Supervisor to send a ``SIGTERM`` signal (a termination signal that allows
+the process to exit cleanly) to all the ``Runner`` processes.
+In this case the supervisor process will remain active. Unless the ``--wait`` option
+is specified, the completion of the command will not imply that all the ``Runner``
+processes have been terminated.
+
+.. warning::
+    The ``Runner`` is designed to recognize the signal and **wait the completion of
```
davidwaroquiers:
```suggestion
    The ``Runner`` is designed to recognize the signal and **wait for the completion of
```

doc/source/user/runner.rst
```diff
@@ -0,0 +1,217 @@
...
+    the action being performed**, before actually exiting.
+
+.. note::
+    Since the supervisor processes remains active, when starting the runner again after
```
davidwaroquiers:
```suggestion
    Since the supervisor processes remain active, when starting the runner again after
```

doc/source/user/runner.rst
```diff
@@ -0,0 +1,217 @@
...
+providing the state of each individual daemon process and its system
+process ID.
+
+Running daemon check
+====================
+
+There are no strict limitations to which machine should be used to execute the ``Runner``
+as a daemon, and, as explained in the :ref:`setup options` section, there are several
+possible configurations. It is thus possible for a user to mistakenly start
+the runner daemon on two different locations. While this should not corrupt
```
davidwaroquiers, line 166: I thought some jobs could fail because they are checked out by one runner and submitted by another, or downloaded by one and completed by another. No?

tests/db/jobs/test_daemon.py
```diff
@@ -3,31 +3,44 @@
...
     for _i in range(max_wait):
         time.sleep(1)
-        state = daemon_manager.check_status()
-        assert state in (DaemonStatus.STARTING, DaemonStatus.RUNNING)
-        if state == DaemonStatus.RUNNING:
+        state = None
+        try:
+            state = daemon_manager.check_status()
+        except DaemonError:
+            pass
```
davidwaroquiers: Here you pass while in the similar fixture, you continue. Is there a reason ?
Also, is this function needed (see below comment as to why not using the fixture instead) ?

tests/db/jobs/test_daemon.py
```diff
@@ -64,16 +77,23 @@ def test_start_stop(job_controller, single, daemon_manager) -> None:
...
+        warn_msg = (
+            "Daemon is STOPPED, but the options single differ from the values used "
+            "to activate supervisor. The daemon will start with the initial configurations"
+        )
+        assert warn_msg in caplog.text
+
+    wait_daemon_started(daemon_manager)
     assert daemon_manager.shut_down(raise_on_error=True)
-    time.sleep(1)
+    _wait_daemon_shutdown(daemon_manager)
```
davidwaroquiers: Why not using the wait_daemon_shutdown fixture ?

davidwaroquiers (commented):
Added last few comments or suggestions.

doc/source/user/runner.rst
gpetretto: no, from a copy paste. thanks

doc/source/user/runner.rst
gpetretto: fixed

doc/source/user/runner.rst
gpetretto: Indeed, I corrected the description.

tests/db/jobs/test_daemon.py
gpetretto: I switched to the fixture. In the other case I use a continue because it checks that current state belongs to a list

tests/db/jobs/test_daemon.py
gpetretto: switched to the fixture

davidwaroquiers (commented):
All good! Can't approve as it was initially my PR. Bit approved 😀

gpetretto (approved):

PR #154, opened 2024-07-25
delete_job functionality

### LINKED ISSUE #141: Delete job functionality

### LINKED ISSUE #169: Cannot force delete more than 10 flows
I am unable to delete more than 10 flows at a time, even with explicit confirmation and setting the `--force` flag. Is this the intended behavior or a bug?

image

image

gpetretto:
A new option to delete a single Job of a Flow.
No checks for consistency are performed, so it is the user's responsibility to leave the Flow in a consistent state.

### DIFF
--- a/src/jobflow_remote/cli/job.py
+++ b/src/jobflow_remote/cli/job.py
@@ -497,6 +497,61 @@ def stop(
 
 
 @app_job.command()
+def delete(
+    job_db_id: job_db_id_arg = None,
+    job_index: job_index_arg = None,
+    job_id: job_ids_indexes_opt = None,
+    db_id: db_ids_opt = None,
+    flow_id: flow_ids_opt = None,
+    state: job_state_opt = None,
+    start_date: start_date_opt = None,
+    end_date: end_date_opt = None,
+    name: name_opt = None,
+    metadata: metadata_opt = None,
+    days: days_opt = None,
+    hours: hours_opt = None,
+    verbosity: verbosity_opt = 0,
+    wait: wait_lock_opt = None,
+    raise_on_error: raise_on_error_opt = False,
+    delete_output: Annotated[
+        bool,
+        typer.Option(
+            "--output",
+            "-o",
+            help="Also delete the outputs of the Jobs in the output Store",
+        ),
+    ] = False,
+) -> None:
+    """
+    Delete Jobs individually. The Flow document will be updated accordingly but
+    no consistency check is performed. The Flow may be left in an inconsistent state.
+    For advanced users only.
+    """
+    jc = get_job_controller()
+
+    execute_multi_jobs_cmd(
+        single_cmd=jc.delete_job,
+        multi_cmd=jc.delete_jobs,
+        job_db_id=job_db_id,
+        job_index=job_index,
+        job_ids=job_id,
+        db_ids=db_id,
+        flow_ids=flow_id,
+        states=state,
+        start_date=start_date,
+        end_date=end_date,
+        name=name,
+        metadata=metadata,
+        days=days,
+        hours=hours,
+        verbosity=verbosity,
+        wait=wait,
+        raise_on_error=raise_on_error,
+        delete_output=delete_output,
+    )
+
+
+@app_job.command()
 def queue_out(
     job_db_id: job_db_id_arg,
     job_index: job_index_arg = None,
--- a/src/jobflow_remote/cli/job.py
+++ b/src/jobflow_remote/cli/job.py
@@ -17,6 +17,9 @@ from jobflow_remote.cli.types import (
     break_lock_opt,
     days_opt,
     db_ids_opt,
+    delete_all_opt,
+    delete_files_opt,
+    delete_output_opt,
     end_date_opt,
     flow_ids_opt,
     hours_opt,
@@ -513,20 +516,19 @@ def delete(
     verbosity: verbosity_opt = 0,
     wait: wait_lock_opt = None,
     raise_on_error: raise_on_error_opt = False,
-    delete_output: Annotated[
-        bool,
-        typer.Option(
-            "--output",
-            "-o",
-            help="Also delete the outputs of the Jobs in the output Store",
-        ),
-    ] = False,
+    delete_output: delete_output_opt = False,
+    delete_files: delete_files_opt = False,
+    delete_all: delete_all_opt = False,
 ) -> None:
     """
     Delete Jobs individually. The Flow document will be updated accordingly but
     no consistency check is performed. The Flow may be left in an inconsistent state.
     For advanced users only.
     """
+
+    if delete_all:
+        delete_files = delete_output = True
+
     jc = get_job_controller()
 
     execute_multi_jobs_cmd(
@@ -548,6 +550,7 @@ def delete(
         wait=wait,
         raise_on_error=raise_on_error,
         delete_output=delete_output,
+        delete_files=delete_files,
     )
 
 
--- a/src/jobflow_remote/jobs/jobcontroller.py
+++ b/src/jobflow_remote/jobs/jobcontroller.py
@@ -37,6 +37,7 @@ from jobflow_remote.jobs.data import (
     projection_job_info,
 )
 from jobflow_remote.jobs.state import (
+    DELETABLE_STATES,
     PAUSABLE_STATES,
     RESETTABLE_STATES,
     RUNNING_STATES,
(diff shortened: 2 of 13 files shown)


src/jobflow_remote/utils/db.py
```diff
@@ -142,11 +142,11 @@ def __init__(
         self.lock_id = lock_id or suuid()
         self.kwargs = kwargs
         self.update_on_release: dict | list = {}
+        self.delete_on_release: bool = False
```
davidwaroquiers: Maybe add to the docstring in the main characteristics that it allows to delete the document upon release, and add an example (similar to update_on_release). Maybe one thing that could be good (even if it is more verbose, but I had to look into the code to know how to to an update on release) is to add an "Attributes" section in the class docstring (somehow in "addition" to the __init__ "Parameters" section), so that update_on_release and delete_on_release (maybe a few others that are not in the arguments ?) are clearly documented. What do you think ?

src/jobflow_remote/utils/db.py
```diff
@@ -243,21 +243,36 @@ def release(self, exc_type, exc_val, exc_tb) -> None:
...
-        )
-
-        # Check if the lock was successfully released
-        if result.modified_count == 0:
-            msg = f"Could not release lock for document {self.locked_document['_id']}"
-            warnings.warn(msg, stacklevel=2)
+
+        if self.delete_on_release:
+            if self.update_on_release:
+                raise ValueError(
```
davidwaroquiers: Maybe this check could be done when trying to set update_on_release and/or delete_on_release ? It would require to modify the code a bit. Probably make update_on_release and delete_on_release properties with getter/setter methods and having this check in the setter methods. This way the raise is not at release time (which then may leave the document in a locked state because it never gets to the update of locks to None). What do you think ?

src/jobflow_remote/utils/db.py
davidwaroquiers: Would also be good to add some tests just for MongoLock's update_on_release and delete_on_release (outside of its usage elsewhere). I can do that in another PR.

src/jobflow_remote/cli/job.py
```diff
@@ -494,6 +494,61 @@ def stop(
...
+    start_date: start_date_opt = None,
+    end_date: end_date_opt = None,
+    name: name_opt = None,
+    metadata: metadata_opt = None,
+    days: days_opt = None,
+    hours: hours_opt = None,
+    verbosity: verbosity_opt = 0,
+    wait: wait_lock_opt = None,
+    raise_on_error: raise_on_error_opt = False,
+    delete_output: Annotated[
```
davidwaroquiers: This is the same as the one in flows. Should it be shared ? (and go in types.py)

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -3189,6 +3190,7 @@ def update_flow_state(
         updated_states
             A dictionary with the updated states of Jobs that have not been
             stored in the DB yet. In the form {job_uuid: JobState value}.
+            If the value is None the Job is considered deleted.
```
davidwaroquiers: Just to make sure I understand properly, the value of the state of a deleted Job might be None while doing the delete (and inside the lock_job_flow  context in the delete_job method) right ? But a Job can never have a None state in the database correct ? Maybe it might be good to explain this as a comment in the delete_job method (and here). I was initially "puzzled" by the fact that a job state could be None and I was thinking then are these (or this, in this case) jobs still in the database but "set" as "deleted" ("using" a None state).

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -3570,6 +3576,190 @@ def remove_batch_process(self, process_id: str, worker: str) -> dict:
...
+            # Remove job from ids list
+            flow_doc.ids = [
+                id_tuple
+                for id_tuple in flow_doc.ids
+                if id_tuple[1] != job_uuid or id_tuple[2] != job_index
+            ]
+
+            # Remove job from jobs list if no job with that id remains in the flow
+            if not any(job_uuid == id_tuple[1] for id_tuple in flow_doc.ids):
+                flow_doc.jobs.remove(job_uuid)
```
davidwaroquiers: Maybe I'm not enough familiar with the structure of the FlowDoc but I thought the jobs list had multiple occurrences of the same job uuid if there has been some replace. If that's the case shouldn't it always remove the job_uuid ? (i.e. it would remove one of them). Then related to that, if it's the case, is there a matching between the jobs list and the ids list ? Maybe I just missed the fact that the jobs attribute in FlowDoc contains the unique uuids.

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -3570,6 +3576,190 @@ def remove_batch_process(self, process_id: str, worker: str) -> dict:
...
+
+            # Remove job from parents
+            flow_doc.parents.pop(job_uuid, None)
+            for parent_dict in flow_doc.parents.values():
+                for index_list in parent_dict.values():
+                    if job_uuid in index_list:
+                        index_list.remove(job_uuid)
```
davidwaroquiers: What if there has been a replace and we delete the job with index 2 ? Shouldn't the job with index 1 still be a parent ? Maybe I'm also missing something in the logic.

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -3570,6 +3576,190 @@ def remove_batch_process(self, process_id: str, worker: str) -> dict:
...
+                try:
+                    if delete_output:
+                        self.jobstore.remove_docs({"uuid": job_id, "index": job_index})
+                except Exception:
+                    warnings.warn(
+                        f"Error while delete the output of job {job_id} {job_index}",
+                        stacklevel=2,
+                    )
+
+            return [job_doc["db_id"]]
```
davidwaroquiers: Why does it return a list ? (it's actually also a question for all other kinds of job-related methods)

davidwaroquiers (commented):
Thanks for this!
I have some comments, questions and maybe thoughts for improvements but maybe some (or most/all :p) of these are because I am not fully aware of everything in the code, and I did not have time to completely go through the rest and fully grasp all the details.

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: The Job is never really in the None state. The point here is just to signal that the state of the Job should be ignored, since it will be deleted in the update. I will try to clarify the usage.

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: My idea was that the `jobs` attribute should only contain one entry for each uuid, but I see that indeed when a replace is present it contains the same uuid multiple times. Aside from this specific case it should not cause issues, since it should always be used just to find a Flow that contains a Job with a specific uuid. So it does not matter that it is present multuple times. Nonetheless I would consider this a bug and fix it. 
In that case this would be correct, but maybe I will change this to 
```python 
flow_doc.jobs = [jid for jid in flow_doc.jobs if jid != job_uuid]
```
to be backward compatible.
What do you think?

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: Yes. Thanks for spotting this!

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: Because all the methods acting on multiple jobs use the `_many_jobs_action` to apply the same action on a list of jobs and call the methods that act on a single job. This one expects a list of jobs as output. I thought it would be easier to keep the same API for all these methods (also in the beginning there were much less such mehtods and the first one was `rerun_job`, that indeed return a list of multiple ids). However it would also be possible to handle this in `_many_jobs_action` and let methods that only modify one job to return a single entry. Do you think it would be better?
In the case of `delete_job` it may be dubious, since maybe in the future we may introduce the option to also delete linked jobs?

src/jobflow_remote/utils/db.py
gpetretto: Seems a good idea.

gpetretto:
Some changes with the last commits.
* added the option to delete the files when deleting the Job.
* moved the logic to delete the files in `JobController` (also when deleting Flows)
* changed the signature of the functions that deal with a single job in `JobController`: now if the function acts only on a single Job returns the single db_id, not a list with a single element. The change is handled in `_many_jobs_action`.
* Fix for #169 by allowing to set the maximum number of flows to be deleted.
* add the same option for job deletion as well.

davidwaroquiers (approved):
Great work! Thanks!

Issue #149, opened 2024-07-13
Is there a way to visualize the workflow based on jobflow-remote info?


QuantumChemist:
... and if not, would you mind if I add some code to generate a Mermaid flowchart based on jf job info (or similar)?

I know there is to_mermaid in jobflow that can visualize your Flow object but it doesn't contain the dynamic jobs.

I could think of making a Mermaid flowchart that can be output in the terminal (if that's possible at least) and as any kind of image format.

davidwaroquiers:
You can look at the jf flow graph command. Not (yet) full flavored but still nice (mermaid and dash-cytoscape).

QuantumChemist:
> You can look at the jf flow graph command. Not (yet) full flavored but still nice (mermaid and dash-cytoscape).

Cool thank you! I'll try that!

QuantumChemist:
Do you have any idea how I could solve this problem?

```
(auto) certural@...:~$ jf flow graph -m f11725bd-595b-45ba-95cb-0f2a5dab4feb
The selected project is auto from config file /home/certural/.jfremote/auto.yaml
An Error occurred during the command execution: OperationFailure PlanExecutor error during aggregation :: caused by :: Total size of documents in jobs matching pipeline's $lookup 
stage exceeds 104857600 bytes, full error: {'ok': 0.0, 'errmsg': "PlanExecutor error during aggregation :: caused by :: Total size of documents in jobs matching pipeline's $lookup 
stage exceeds 104857600 bytes", 'code': 4568, 'codeName': 'Location4568'}

```

gpetretto:
Hi, the error seems to come from the MongoDB query. That one fetches most of the data of the Jobs also from the jobs collection. Is the Flow very large?

QuantumChemist:
> Hi, the error seems to come from the MongoDB query. That one fetches most of the data of the Jobs also from the jobs collection. Is the Flow very large?

Yes, it has 1055 jobs.

gpetretto:
Hi @QuantumChemist,
the graph command (as well as some other commands) fetched more information than that actually needed. This was not a problem for relatively small Job documents and number of Jobs, but clearly is for your use case.
I made an update in this branch: https://github.com/Matgenix/jobflow-remote/tree/gp/minor_updates
Would you be available to test it to check if this solves your issue?

QuantumChemist:
Hi @gpetretto !

[...]

PR #156, opened 2024-07-26
Minor updates

gpetretto:
A few minor updates.
* Minimize the amount of data fetched during the aggregation joining the flow and job documents. To avoid exceeding memmory limits (see discussion in #149)
* small fix in cli and daemon
* add the option to delete the job execution folders on the worker when deleting a Flow.

A few points to be discussed concerning the last functionality:
* the deletion is implemented in the CLI and thus only available there. In general I would prefer not to overload the JobController with different functionalities, but would it be better to move the file deletion there?
* I added a safeguard before deleting the file to avoid the risk of mistakenly deleting other data: the folder to be deleted should contain the `jfremote_in.json` file, singaling that it is indeed a folder where jobflow-remote was executed. This requires an additional request to the host of the worker for each folder to be deleted. Maybe could be replaced by a check on the folder name (i.e. should match the `UUID_INDEX` format? This would be faster and probably as safe as the other check.
* Paramiko does not expose a method to delete recursively a folder and all its content, so the `rmtree` for the remote host relies on the unix `rm -r`. In the remote host it is declared and used that some commands expect the remote system to be unix. But could this an issue? A potential solution using only paramiko is [this](https://stackoverflow.com/questions/20507055/recursive-remove-directory-using-sftp/20507586#20507586), but I am not convinced that it would be much better.

Apparently the tests did not like the first change. They run fine on my laptop, so I suppose the problem is the mongoDB version. We are using a a mongodb 4.4 for the testings. Mine is 6.0.6 and the most recent is 7.0.12. Can we consider increasing the MongoDB version for the tests? Or should it preserve more backward compatibility?

### DIFF
--- a/src/jobflow_remote/cli/flow.py
+++ b/src/jobflow_remote/cli/flow.py
@@ -227,6 +227,12 @@ def delete(
                             f"since it may not contain a jobflow-remote execution",
                             stacklevel=2,
                         )
+            # explicitly close the connection to all the hosts
+            for host in hosts.values():
+                try:
+                    host.close()
+                except Exception:
+                    pass
 
         progress.add_task(description="Deleting data...", total=None)
 
--- a/src/jobflow_remote/cli/flow.py
+++ b/src/jobflow_remote/cli/flow.py
@@ -1,4 +1,5 @@
-from typing import Annotated, Optional
+import warnings
+from typing import TYPE_CHECKING, Annotated, Optional
 
 import typer
 from jobflow.utils.graph import draw_graph
@@ -39,6 +40,9 @@ from jobflow_remote.cli.utils import (
     out_console,
 )
 from jobflow_remote.jobs.graph import get_graph, plot_dash
+
+if TYPE_CHECKING:
+    from jobflow_remote.remote.host import BaseHost
 
 app_flow = JFRTyper(
     name="flow", help="Commands for managing the flows", no_args_is_help=True
@@ -127,10 +131,29 @@ def delete(
             help="Also delete the outputs of the Jobs in the output Store",
         ),
     ] = False,
+    delete_files: Annotated[
+        bool,
+        typer.Option(
+            "--files",
+            "-fi",
+            help="Also delete the files on the worker",
+        ),
+    ] = False,
+    delete_all: Annotated[
+        bool,
+        typer.Option(
+            "--all",
+            "-a",
+            help="enable --output and --files",
+        ),
+    ] = False,
 ) -> None:
     """Permanently delete Flows from the database"""
     check_incompatible_opt({"start_date": start_date, "days": days, "hours": hours})
     check_incompatible_opt({"end_date": end_date, "days": days, "hours": hours})
+
+    if delete_all:
+        delete_files = delete_output = True
 
     start_date = get_start_date(start_date, days, hours)
 
@@ -176,7 +199,36 @@ def delete(
 
     to_delete = [fi.flow_id for fi in flows_info]
     with loading_spinner(processing=False) as progress:
-        progress.add_task(description="Deleting...", total=None)
+        if delete_files:
+            progress.add_task(description="Deleting files...", total=None)
+            # get all jobs to delete
+            jobs_info = jc.get_jobs_info(flow_ids=to_delete)
+            # keep a dictionary of the hosts to avoid connecting multiple times
+            hosts: dict[str, BaseHost] = {}
+            for job_info in jobs_info:
+                if job_info.run_dir:
+                    if job_info.worker in hosts:
+                        host = hosts[job_info.worker]
+                    else:
+                        host = jc.project.workers[job_info.worker].get_host()
+                        hosts[job_info.worker] = host
+                        host.connect()
+                    remote_files = host.listdir(job_info.run_dir)
+                    # safety measure to avoid mistakenly deleting other folders
+                    # maybe too much?
+                    if any(
+                        fn in remote_files
+                        for fn in ("jfremote_in.json", "jfremote_in.json.gz")
+                    ):
+                        host.rmtree(path=job_info.run_dir, raise_on_error=False)
+                    else:
+                        warnings.warn(
+                            f"Did not delete folder {job_info.run_dir} "
+                            f"since it may not contain a jobflow-remote execution",
+                            stacklevel=2,
+                        )
+
+        progress.add_task(description="Deleting data...", total=None)
 
         jc.delete_flows(flow_ids=to_delete, delete_output=delete_output)
 
@@ -254,6 +306,7 @@ def graph(
     ] = False,
 ) -> None:
     """Provide detailed information on a Flow."""
+    check_incompatible_opt({"dash": dash_plot, "mermaid": print_mermaid})
     db_id, jf_id = get_job_db_ids(flow_db_id, None)
     db_ids = job_ids = flow_ids = None
     if db_id is not None:
@@ -281,7 +334,7 @@ def graph(
 
         print(get_mermaid(flows_info[0]))
 
-    if dash_plot:
+    elif dash_plot:
         plot_dash(flows_info[0])
     else:
         plt = draw_graph(get_graph(flows_info[0], label=label))
--- a/src/jobflow_remote/cli/utils.py
+++ b/src/jobflow_remote/cli/utils.py
@@ -221,7 +221,7 @@ def get_job_db_ids(job_db_id: str, job_index: int | None):
 
     if job_index and db_id is not None:
         out_console.print(
-            "The index is defined even if a db_id is passed as an ID. Will be ignored",
+            "The index is defined while a db_id is passed as an ID. Will be ignored",
             style="yellow",
         )
 
--- a/src/jobflow_remote/cli/utils.py
+++ b/src/jobflow_remote/cli/utils.py
@@ -221,7 +221,7 @@ def get_job_db_ids(job_db_id: str, job_index: int | None):
 
     if job_index and db_id is not None:
         out_console.print(
-            "The index is defined even if an integer is passed as an ID. Will be ignored",
+            "The index is defined even if a db_id is passed as an ID. Will be ignored",
             style="yellow",
         )
 
--- a/src/jobflow_remote/jobs/jobcontroller.py
+++ b/src/jobflow_remote/jobs/jobcontroller.py
@@ -33,6 +33,7 @@ from jobflow_remote.jobs.data import (
     get_initial_flow_doc_dict,
     get_initial_job_doc_dict,
     get_reset_job_base_dict,
+    projection_flow_info_jobs,
     projection_job_info,
 )
 from jobflow_remote.jobs.state import (
@@ -2002,7 +2003,8 @@ class JobController:
     def get_flow_job_aggreg(
         self,
         query: dict | None = None,
-        projection: dict | None = None,
+        projection_flow: dict | None = None,
+        projection_job: dict | None = None,
         sort: list[tuple] | None = None,
         limit: int = 0,
     ) -> list[dict]:
@@ -2015,8 +2017,10 @@ class JobController:
         ----------
         query
             A dictionary representing the filter.
-        projection
-            Projection of the fields passed to the aggregation.
+        projection_flow
+            Projection of the fields for the Flow document passed to the aggregation.
+        projection_job
+            Projection of the fields for the Job document used in the $lookup.
         sort
             A list of (key, direction) pairs specifying the sort order for this
             query. Follows pymongo conventions.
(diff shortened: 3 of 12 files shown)


src/jobflow_remote/cli/utils.py
```diff
@@ -221,7 +221,7 @@ def get_job_db_ids(job_db_id: str, job_index: int | None):
 
     if job_index and db_id is not None:
         out_console.print(
-            "The index is defined even if an integer is passed as an ID. Will be ignored",
+            "The index is defined even if a db_id is passed as an ID. Will be ignored",
```
davidwaroquiers:
```suggestion
            "The index is defined while a db_id is passed as an ID. Will be ignored",
```

src/jobflow_remote/remote/host/local.py
```diff
@@ -114,3 +115,15 @@ def listdir(self, path: str | Path) -> list[str]:
...
+
+    def rmtree(self, path: str | Path, raise_on_error: bool = False) -> bool:
+        removed = True
+
+        def onerror(func, dir_path, err):
+            nonlocal removed
+            removed = False
+            warnings.warn(f"Error while deleting folder {path}: {err[1]}", stacklevel=2)
+
+        shutil.rmtree(path, onerror=onerror if not raise_on_error else None)
```
davidwaroquiers, line 148: Seems like onerror is deprecated, in favor of onexc (https://docs.python.org/3/library/shutil.html#shutil.rmtree)

src/jobflow_remote/remote/host/base.py
```diff
@@ -90,6 +90,10 @@ def listdir(self, path: str | Path) -> list[str]:
     def remove(self, path: str | Path):
         raise NotImplementedError
 
+    @abc.abstractmethod
+    def rmtree(self, path: str | Path, raise_on_error: bool = False) -> bool:
+        raise NotImplementedError
+
```
davidwaroquiers: Add docstring (see suggestion). Maybe docstrings to other methods of the basehost could also be added. Maybe it's a good time to test github copilot ?

```suggestion
    @abc.abstractmethod
    def rmtree(self, path: str | Path, raise_on_error: bool = False) -> bool:
        """Recursively delete a directory tree on a host.
        
        This method must be implemented by subclasses of  `BaseHost`.
        It is intended to remove an entire directory tree, including all files
        and subdirectories, on the host represented by the subclass.
        
        Parameters:
        -----------
        path : str or Path
            The path to the directory tree to be removed.

        raise_on_error : bool, optional
            If set to `False` (default), errors will be ignored, and the method will
            attempt to continue removing remaining files and directories.
            Otherwise, any errors encountered during the removal process
            will raise an exception.
        """
        raise NotImplementedError

```

src/jobflow_remote/remote/host/local.py
```diff
@@ -114,3 +115,15 @@ def listdir(self, path: str | Path) -> list[str]:
 
     def remove(self, path: str | Path) -> None:
         os.remove(path)
+
+    def rmtree(self, path: str | Path, raise_on_error: bool = False) -> bool:
```
davidwaroquiers, line 119: Docstring ? (test and use copilot?)

src/jobflow_remote/remote/host/remote.py
```diff
@@ -321,6 +322,16 @@ def remove(self, path: str | Path) -> None:
 
         self._execute_remote_func(self.connection.sftp().remove, str(path))
 
+    def rmtree(self, path: str | Path, raise_on_error: bool = False) -> bool:
```
davidwaroquiers, line 325: Docstring (copilot ?)

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -2042,8 +2046,21 @@ def get_flow_job_aggreg(
...
+            # to reduce the impact of the size of the documents.
+            # This can help reducing the size of the fetched documents and
+            # avoid exceeding the maximum size allowed. Adding the projection
+            # in the general pipeline does not have the same effect.
+            pipeline[0]["$lookup"]["pipeline"] = [{"$project": projection_job}]
+            # if the additional projection is set, the keys need to be specified
+            # in that part of the pipeline as well.
+            if projection_flow:
+                for k in projection_job:
+                    pipeline[-1]["$project"][f"jobs_list.{k}"] = 1
```
davidwaroquiers, line 2063: Wondering if it would not be more efficient to put the query and/or projection before the lookup ? Or maybe it changes the result ? What do you think ?

src/jobflow_remote/cli/flow.py
```diff
@@ -176,7 +199,36 @@ def delete(
...
+            # keep a dictionary of the hosts to avoid connecting multiple times
+            hosts: dict[str, BaseHost] = {}
+            for job_info in jobs_info:
+                if job_info.run_dir:
+                    if job_info.worker in hosts:
+                        host = hosts[job_info.worker]
+                    else:
+                        host = jc.project.workers[job_info.worker].get_host()
+                        hosts[job_info.worker] = host
+                        host.connect()
```
davidwaroquiers, line 215: Should'nt it be closed after ? I guess a context manager will be difficult to use here as the hosts are connected dynamically while jobs are iterated over (so probably a big try except ? not very nice but maybe the less bad option ?) ? Wondering what happens if a host connection is not closed, is that a big problem ?

src/jobflow_remote/cli/flow.py
```diff
@@ -127,11 +131,30 @@ def delete(
...
+            help="enable --output and --files",
+        ),
+    ] = False,
 ) -> None:
     """Permanently delete Flows from the database"""
     check_incompatible_opt({"start_date": start_date, "days": days, "hours": hours})
     check_incompatible_opt({"end_date": end_date, "days": days, "hours": hours})
 
+    if delete_all:
+        delete_files = delete_output = True
```
davidwaroquiers, line 156: One thing I'm wondering is: if we delete the outputs, but not the files, we somehow have no way to access (easily) these files ... Should we either
- always delete the files when we delete the outputs (maybe with an option to keep the files?), but maybe it's dangerous ..
- delete most of the content of the dict in the database, but keep uuid, index, db_id and run directory somehow ?
- do nothing

In any case, I think we should document this carefully, whichever option we choose above.

davidwaroquiers (commented):
Thanks for the implementation. Gave a few comments on the delete functionality that we can discuss when you get back. A few other minor comments here and there.

src/jobflow_remote/remote/host/local.py
gpetretto: I have seen that, but `onexec` was introduced in python 3.12, so we cannot use it unless we drop support for all previous versions.

or maybe there is a `__future__` import for that?

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: I am not sure what is the suggestion here. Making two different queries in python and match the results? This is probably not very efficient. Otherwise this behaves roughly like a `JOIN` in SQL, so it first needs to perform the matches in order to make the filter. I don't think the order can be altered here.

src/jobflow_remote/cli/flow.py
gpetretto: I think that as long as it runs just from the CLI it will just be closed by paramiko itself before terminating the execution. However it would be safer to enforce it. I will add a close

src/jobflow_remote/cli/flow.py
gpetretto: Good point. I think the preferred way would depend a lot on users preferences, so probably we should have the least invasive one as default.

src/jobflow_remote/remote/host/local.py
davidwaroquiers: Ah indeed, I did not check when it was added sorry. I don't know if there is a `__future__` import for it. We can leave it this way for now anyway.

src/jobflow_remote/cli/flow.py
davidwaroquiers: So maybe something like the second option ? i.e. "delete most of the content of the dict in the database, but keep uuid, index, db_id and run directory somehow ?"

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: no, I meant to change the pipeline order. But again, I'm not completely sure about this :)

src/jobflow_remote/remote/host/local.py
```diff
@@ -117,6 +117,28 @@ def remove(self, path: str | Path) -> None:
         os.remove(path)
 
     def rmtree(self, path: str | Path, raise_on_error: bool = False) -> bool:
+        """Recursively delete a directory tree on a host.
+
+        This method must be implemented by subclasses of  `BaseHost`.
```
davidwaroquiers: Here its not anymore the baseclass so I guess this paragraph should be removed, maybe the first sentence could be "Recursively delete a directory tree on this local host."

src/jobflow_remote/remote/host/remote.py
```diff
@@ -323,6 +323,28 @@ def remove(self, path: str | Path) -> None:
         self._execute_remote_func(self.connection.sftp().remove, str(path))
 
     def rmtree(self, path: str | Path, raise_on_error: bool = False) -> bool:
+        """Recursively delete a directory tree on a host.
+
+        This method must be implemented by subclasses of  `BaseHost`.
```
davidwaroquiers: Same as for the local host, here its not anymore the baseclass so I guess this paragraph should be removed, maybe the first sentence could be "Recursively delete a directory tree on this remote host."

davidwaroquiers (approved):

src/jobflow_remote/cli/flow.py
gpetretto: Looking more into this, the problem with that approach is that the changes to the output documents are done through the JobStore (and cannot be otherwise). Since the maggma Store only allows full deletion or updates, it means that to leave only a minimal version of the outputs the documents should all be fetched from the Store, cleaned up and resinserted into the Store. If there are many this may be an expensive operation and I am not sure if it is worth.

src/jobflow_remote/remote/host/local.py
```diff
@@ -114,3 +115,36 @@ def listdir(self, path: str | Path) -> list[str]:
 
     def remove(self, path: str | Path) -> None:
         os.remove(path)
+
+    def rmtree(self, path: str | Path, raise_on_error: bool = False) -> bool:
+        """Recursively delete a directory tree on a host.
```
davidwaroquiers:
```suggestion
        """Recursively delete a directory tree on a local host.
```

src/jobflow_remote/remote/host/local.py
```diff
@@ -114,3 +115,36 @@ def listdir(self, path: str | Path) -> list[str]:
...
+
+        It is intended to remove an entire directory tree, including all files
+        and subdirectories, on the host represented by the subclass.
```
davidwaroquiers:
```suggestion
        and subdirectories, on this local host.
```

src/jobflow_remote/remote/host/remote.py
```diff
@@ -321,6 +322,37 @@ def remove(self, path: str | Path) -> None:
 
         self._execute_remote_func(self.connection.sftp().remove, str(path))
 
+    def rmtree(self, path: str | Path, raise_on_error: bool = False) -> bool:
+        """Recursively delete a directory tree on a host.
```
davidwaroquiers:
```suggestion
        """Recursively delete a directory tree on a remote host.
```

src/jobflow_remote/remote/host/remote.py
```diff
@@ -321,6 +322,37 @@ def remove(self, path: str | Path) -> None:
...
+
+        It is intended to remove an entire directory tree, including all files
+        and subdirectories, on the host represented by the subclass.
```
davidwaroquiers:
```suggestion
        and subdirectories, on this remote host.
```

src/jobflow_remote/cli/flow.py
davidwaroquiers: I propose to keep this as it is for now and have an issue to keep track of it and see if we should do something.

src/jobflow_remote/cli/flow.py
davidwaroquiers: Only comment left on my side I think (you said you'd add a close but I did not see it ? if you decided not to add it in the end, ok for me too)

davidwaroquiers (approved):

src/jobflow_remote/cli/flow.py
gpetretto: Done. I forgot to push the change.

Issue #157, opened 2024-07-30
Jobs fail with error `Remote error: file path/to/job_dir/remote_job_data.json for job xxx does not exist`


QuantumChemist:
Hey!

so on the weekend we had some network issues and connection to the remote cluster was disrupted. Therefore all the jobs failed with `REMOTE_ERROR` (showing authentication error). Yesterday that problem was fixed and for a few hours, I could run some jobs again. But for whatever reason I can't figure out, in the evening the jobs started to fail with  `Remote error: file path/to/job_dir/remote_job_data.json for job xxx does not exist` and if I copy the file (that only contains "[]" for comparable jobs)  there, it will complain that  `Remote error: file path/to/job_dir/additional_store_data.json for job xxx does not exist` and if I put that file there (which also only contains "[]" for other finished jobs), the job will still fail with `REMOTE_ERROR` saying:
```
│      error = Traceback (most recent call last):                                                                                                         │
│                File "/home/user-on-remote-cluster/miniconda3/envs/auto/lib/python3.10/site-packages/jobflow_remote/jobs/run.py", line 41, in run_remote_job │
│                  in_data = loadfn(IN_FILENAME)                                                                                                          │
│                File "/home/user-on-remote-cluster/miniconda3/envs/auto/lib/python3.10/site-packages/monty/serialization.py", line 82, in loadfn             │
│                  return json.load(fp, *args, **kwargs)                                                                                                  │
│                File "/home/user-on-remote-cluster/miniconda3/envs/auto/lib/python3.10/json/__init__.py", line 293, in load                                  │
│                  return loads(fp.read(),                                                                                                                │
│                File "/home/user-on-remote-cluster/miniconda3/envs/auto/lib/python3.10/json/__init__.py", line 359, in loads                                 │
│                  return cls(**kw).decode(s)                                                                                                             │
│                File "/home/user-on-remote-cluster/miniconda3/envs/auto/lib/python3.10/site-packages/monty/json.py", line 843, in decode                     │
│                  return self.process_decoded(d)                                                                                                         │
│                File "/home/user-on-remote-cluster/miniconda3/envs/auto/lib/python3.10/site-packages/monty/json.py", line 820, in process_decoded            │
│                  return {                                                                                                                               │
│                File "/home/user-on-remote-cluster/miniconda3/envs/auto/lib/python3.10/site-packages/monty/json.py", line 821, in <dictcomp>                 │
[...]

PR #158, opened 2024-08-01
Project check

davidwaroquiers:
Adding check on version compatibility between the runner and the workers.
This would have probably allowed to prevent #157 from happening (at least finding out the reason maybe ?).

### DIFF
--- a/src/jobflow_remote/cli/project.py
+++ b/src/jobflow_remote/cli/project.py
@@ -163,6 +163,13 @@ def check(
             help="Print the errors at the end of the checks",
         ),
     ] = False,
+    full: Annotated[
+        bool,
+        typer.Option(
+            "--full",
+            help="Perform a full check",
+        ),
+    ] = False,
 ) -> None:
     """Check that the connection to the different elements of the projects are working."""
     check_incompatible_opt({"jobstore": jobstore, "queue": queue, "worker": worker})
@@ -192,9 +199,9 @@ def check(
             worker_to_test = project.workers[worker_name]
             if worker_to_test.get_host().interactive_login:
                 with hide_progress(progress):
-                    err = check_worker(worker_to_test)
+                    err = check_worker(worker_to_test, full_check=full)
             else:
-                err = check_worker(worker_to_test)
+                err = check_worker(worker_to_test, full_check=full)
             header = tick
             if err:
                 errors.append((f"Worker {worker_name}", err))
--- a/src/jobflow_remote/cli/project.py
+++ b/src/jobflow_remote/cli/project.py
@@ -167,6 +167,7 @@ def check(
         bool,
         typer.Option(
             "--full",
+            "-f",
             help="Perform a full check",
         ),
     ] = False,
@@ -190,6 +191,7 @@ def check(
         workers_to_test = [worker]
 
     tick = "[bold green]✓[/] "
+    tick_warn = "[bold yellow]✓[/] "
     cross = "[bold red]x[/] "
     errors = []
     with loading_spinner(processing=False) as progress:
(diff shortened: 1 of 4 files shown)


src/jobflow_remote/cli/project.py
```diff
@@ -155,6 +155,13 @@ def check(
             help="Print the errors at the end of the checks",
         ),
     ] = False,
+    full: Annotated[
+        bool,
+        typer.Option(
+            "--full",
```
gpetretto, line 169: maybe also add the `-f` option?

src/jobflow_remote/config/helper.py
```diff
@@ -206,3 +209,57 @@ def check_jobstore(jobstore: JobStore) -> str | None:
...
+    """
+    # TODO: not sure about this test here but I based this check function on the _check_worker function
+    #  which does the same.
+    try:
+        host_error = host.test()
+        if host_error:
+            return host_error
+    except Exception:
+        exc = traceback.format_exc()
+        return f"Error while testing worker:\n {exc}"
```
gpetretto: I think this is redundant, since this check test is already done in `_check_worker`. So here it is just repeating it.

src/jobflow_remote/config/helper.py
```diff
@@ -165,6 +167,7 @@ def check_worker(worker: WorkerBase) -> str | None:
         qm.get_jobs_list()
 
         _check_workdir(worker=worker, host=host)
```
gpetretto: Here there is probably a bug. `_check_workdir` may return a string and this means there was an error. But it is never reported to the outside of this function. Can you fix it in this PR? either always raise in `_check_workdir` or handle the returned string here. Probably easier to do the former.
Then do the same with `_check_environment`

src/jobflow_remote/config/helper.py
```diff
@@ -206,3 +209,57 @@ def check_jobstore(jobstore: JobStore) -> str | None:
...
+        if local_package_versions[package] != host_package_versions[package]:
+            mismatch.append(
+                (
+                    package,
+                    local_package_versions[package],
+                    host_package_versions[package],
+                )
+            )
+    if missing or mismatch:
+        raise ValueError("Version mismatch")
```
gpetretto: Here will be mentioned the differences in the versions?

gpetretto (commented):
I have added some comments on the changes. I think there is a bug in the original version of the check. Maybe it can be fixed in this PR.

src/jobflow_remote/cli/project.py
```diff
@@ -192,13 +201,16 @@ def check(
...
-                err = check_worker(worker_to_test)
+                err, worker_warn = check_worker(worker_to_test, full_check=full)
             header = tick
             if err:
-                errors.append((f"Worker {worker_name}", err))
+                errors.append((f"Worker {worker_name} ", err))
                 header = cross
+            elif worker_warn:
+                errors.append((f"Worker {worker_name} warning ", worker_warn))
+                header = tick_warn
```
davidwaroquiers: From the check_worker function, it's not possible to have an error AND a warning (at least currently I think). Could that be possible in the future ? I had to look in the check_worker about this, maybe I'd either put a comment in the code here to explain that ?

davidwaroquiers (commented):
All good to me (just one small comment/question regarding the check_worker but I leave it up to you to take the decision to do anything or not, be it a comment or changing the code). Thanks! Can't approve my own PR :D

src/jobflow_remote/cli/project.py
gpetretto: I changed it to manage in case both are returned and also added a short comment.

gpetretto (approved):

Issue #110, opened 2024-04-24
SUGGESTION: `--count` option for `jf flow list` and `jf job list`


utf:
It can be useful to know how many jobs you have remaining. I'm potentially interested in implementing this if I have some time.

davidwaroquiers:
Hi @utf . That could indeed be useful. In the jobcontroller, there is already the count_jobs and count_flows methods (https://github.com/Matgenix/jobflow-remote/blob/develop/src/jobflow_remote/jobs/jobcontroller.py#L2397 and https://github.com/Matgenix/jobflow-remote/blob/develop/src/jobflow_remote/jobs/jobcontroller.py#L2460) to do it so it should not be too complex to add it to the cli. Maybe it could be "jf job count" and "jf flow count" ? instead of "jf job list --count" and "jf flow list --count" ?

gpetretto:
I think I can quickly add it, it should be quite easy to do. Let me know if you instead prefer using this as a way to start working your way thorugh jobflow-remote's code.

utf:
Thanks for the suggestions. I'd be quite interested to implement it myself as an opportunity to dig further into the code!

gpetretto:
Great, thanks! Let me know if you need any clarification. I believe that the CLI section is not the most documented.

gpetretto:
Hi @utf, sorry but I needed this feature, so I ended up implementing it in #307

PR #171, opened 2024-09-04
Add --worker and --query option in CLI

### LINKED ISSUE #170: Add  `--query` (or `--worker`) flag to `jf job rerun/retry`
i would like to rerun jobs selectively by worker. this reruns/retries _all_ jobs:

```sh
jf --project <proj-name> job retry --state REMOTE_ERROR
jf --project <proj-name> job rerun --state FAILED
```

there's the `--metadata` flag but that only allows filtering by `job.metadata.<any-sub-key>` (which in my case is an empty dict for all jobs across multiple workers). how about renaming `--metadata` to `--query` (as in `fireworks`) to allow filtering by any document key?

gpetretto:
Closes #170 adding the `--worker` and `--query` options in the CLI to filter based on worker name and with a custom query.

I considered adding the `--query` for filtering on flows as well, but it seems that the current options basically cover the content of the Flow document. However it can be added if it seems useful.

### DIFF
--- a/src/jobflow_remote/cli/utils.py
+++ b/src/jobflow_remote/cli/utils.py
@@ -149,6 +149,13 @@ def check_incompatible_opt(d: dict) -> None:
         exit_with_error_msg(f"Options {options_list} are incompatible")
 
 
+def check_query_incompatibility(query, incompatible_options):
+    if query and any(opt is not None for opt in incompatible_options):
+        exit_with_error_msg(
+            "The --query option is incompatible with all the other filtering options"
+        )
+
+
 def check_at_least_one_opt(d: dict) -> None:
     not_none = []
     for k, v in d.items():
@@ -332,9 +339,11 @@ def execute_multi_jobs_cmd(
     start_date: datetime | None = None,
     end_date: datetime | None = None,
     name: str | None = None,
-    metadata: str | None = None,
+    metadata: dict | None = None,
     days: int | None = None,
     hours: int | None = None,
+    workers: list[str] | None = None,
+    custom_query: dict | None = None,
     verbosity: int = 0,
     raise_on_error: bool = False,
     **kwargs,
@@ -350,6 +359,8 @@ def execute_multi_jobs_cmd(
         metadata,
         days,
         hours,
+        workers,
+        custom_query,
     ]
     try:
         if job_db_id is not None:
@@ -370,7 +381,22 @@ def execute_multi_jobs_cmd(
                 {"start_date": start_date, "days": days, "hours": hours}
             )
             check_incompatible_opt({"end_date": end_date, "days": days, "hours": hours})
-            metadata_dict = str_to_dict(metadata)
+            check_query_incompatibility(
+                custom_query,
+                [
+                    job_ids,
+                    db_ids,
+                    flow_ids,
+                    states,
+                    start_date,
+                    end_date,
+                    name,
+                    metadata,
+                    days,
+                    hours,
+                    workers,
+                ],
+            )
 
             job_ids_indexes = get_job_ids_indexes(job_ids)
             start_date = get_start_date(start_date, days, hours)
@@ -385,6 +411,8 @@ def execute_multi_jobs_cmd(
                     end_date,
                     name,
                     metadata,
+                    workers,
+                    custom_query,
                 )
             ):
                 text = Text.from_markup(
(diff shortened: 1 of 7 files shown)


src/jobflow_remote/cli/utils.py
```diff
@@ -370,7 +381,22 @@ def execute_multi_jobs_cmd(
                 {"start_date": start_date, "days": days, "hours": hours}
             )
             check_incompatible_opt({"end_date": end_date, "days": days, "hours": hours})
-            metadata_dict = str_to_dict(metadata)
+            check_query_incompatibility(
```
davidwaroquiers, line 384: Fine for now. Just a question. Do you envision making some of these compatible with the custom query ?

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -670,16 +693,35 @@ def _many_jobs_action(
...
+            start_date,
+            end_date,
+            name,
+            metadata,
+            workers,
+        ]
+        if custom_query and any(opt is not None for opt in filtering_options):
+            raise ValueError(
+                "The custom query option is incompatible with all the other filtering options"
+            )
```
davidwaroquiers, line 709: why not using the check_query_incompatibility function from utils here ?

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -264,6 +269,10 @@ def _build_query_job(
         if metadata:
             metadata_dict = {f"job.metadata.{k}": v for k, v in metadata.items()}
             query.update(metadata_dict)
+            print(query)
```
davidwaroquiers: Leftover ?

davidwaroquiers (approved):
LGTM! Just a few small comments. Maybe @janosh you can also have a look and test to see if it fits your needs ?

janosh:
thanks a lot @gpetretto for implementing! looks great 👍 

maybe worth mentioning the new flags in the [dealing with errors docs](https://matgenix.github.io/jobflow-remote/user/errors.html#dealing-with-errors)?
perhaps `--worker` could be mentioned after this

> If, for example, this was due to a problem with the connection to the worker that has already been resolved, it is possible to avoid waiting for the job to be retried. This can be done by running:
> ...
> [To rerun all `REMOTE_ERROR` jobs for a given worker, use ...]

src/jobflow_remote/cli/utils.py
gpetretto: Honestly not. Mainly to avoid having to deal with potentially clashing or involved queries. For example cases where one might need to handle `"$and"`, `$or` or the same field. I would consider doing the effort of adding the feature only if someone asks for it.

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: In principle the `check_query_incompatibility` is a functionality specific for the CLI as it does not raise, but exists from the CLI with `exit_with_error_msg`. I could try to adapt it to cover both use cases, but I am not sure it is worth.

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: yes, thanks

gpetretto:
Hi @janosh, thanks for the suggestion. I do not fully agree with it though. I feel like that the error section of the documentation should try to stay focused on the different cases that could happen concerning the errors. This would indeed add additional information, but more on a use case of the CLI rather than explaining how the errors are different from one another and which specific command to use. Maybe a new section of the documentation showing the more commonly used options for the CLI could be a better place to mention this? (I would need to leave this for a future PR though)

janosh:
> Maybe a new section of the documentation showing the more commonly used options for the CLI could be a better place to mention this?

sounds good! i think such a page could make extra sense once the coming `--count` feature from https://github.com/Matgenix/jobflow-remote/issues/110 is implemented

src/jobflow_remote/cli/utils.py
davidwaroquiers: sure I wasn't suggesting to do it now anyway, jsut wondering :) resolving this

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: ah yes I see. No let's keep it like this. Resolving

PR #172, opened 2024-09-05
Parallel batch submission

### LINKED ISSUE #86: Does jobflow-remote support the pilot job model?
I still haven't used jobflow-remote yet (my apologies), but I had a question I wanted to ask.

The conventional way people used FireWorks was to have each `@job` be a Slurm job. This can be incredibly inefficient in practice due to queuing policies on the HPC cluster. Lately, many workflow tools built for HPC have adopted a ["pilot job" model](https://en.wikipedia.org/wiki/Pilot_job). In this framework, the daemon will request a Slurm allocation that will run many `@job`s concurrently, continually pulling in new `@job`s as old ones finish (until the walltime is reached). For example, if you are running VASP with 4 nodes per `@job`, you might have the daemon request one or more 20-node Slurm jobs, each of which would run at most 5 concurrent VASP `@job`s (and when one finishes, a new one is fetched from the database). In Materials Project land, we have called this "jobpacking" in FireWorks, although it was always very hacky.

Does jobflow-remote support such a model for workflow execution? I saw @JaGeo's [issue](https://github.com/Matgenix/jobflow-remote/issues/79) about 10s of thousands of calculations, so I figure it might be the case since no HPC center would be happy with you launching 24,000 Slurm jobs, but I wanted to inquire.

### LINKED ISSUE #96: Missing docs: batch mode
Just raising an issue about the batch mode, which is not documented. In fact, the batch config model is just serving the docstring for `BaseModel` at https://matgenix.github.io/jobflow-remote/api/jobflow_remote.config.base.html#jobflow_remote.config.base.BatchConfig so it is very well hidden.

gpetretto:
As a long standing request (see #86), I have started implementing the option to allow the execution of parallel jobs in the batch mode. Up to now only one job at the time could run in a batch job (see [docs](https://matgenix.github.io/jobflow-remote/user/advancedoptions.html#batch-submission)).
With this implementation I managed to run multiple atomate2 vasp Jobs in parallel in a single job submitted to SLURM.

I have two main doubts about the implementation and I would be interested to have some comments from people that have used this functionality with fireworks.

1) In fireworks it is possible to specify the name of the `nodefile`, so that a list of nodes/cores can be generated
    https://github.com/materialsproject/fireworks/blob/96300617c8a3471f83e1b8aed978552e6ac6a15d/fireworks/scripts/mlaunch_run.py#L57
    https://github.com/materialsproject/fireworks/blob/e9f2004d3d861e2de288f0023ecdb5ccd9114939/fireworks/features/multi_launcher.py#L235
    However, it seems that this information is never used in atomate/atomate2. So, if this is not really useful, I would avoid adding this and eventually save this in case an explicit request comes in in the future. Can anybody confirm that this is not used for typical fireworks multi launch executions?
2) At the moment there is no way to monitor the SLURM/PBS/... jobs (i.e., the jobs that will run multiple jobflow `Job`s) that have been submitted and have finished running. These do not fit in the regular `jobs` documents, since there is not a 1->1 correspondence between the SLURM jobs and the jobflow Job. It is instead possible to get a list of those that are currently running. This could be mainly a problem if some issues happen inside the batch execution and one has to track down its logs (the logs of the SLURM job, not the jobflow one. The latter will still be easily accessible) or check that the batch submission is actually working fine. The questions here are: how important would be to have the full history of all the batch jobs submitted to SLURM/PBS? The most straightforward way would be to add a new collection to contain these data, is it worth adding this? Any alternative suggestion? 

I think @Andrew-S-Rosen, @JaGeo, @utf may be interested in the feature.

Fixes #86 
Fixes #96

# TODO

* update the documentation.
* Tests? I still have to figure out if it would be possible to effectively test that the jobs really run in parallel. Maybe just adding test so that the code is tested and avoid trivial regressions.


Andrew-S-Rosen:
Thank you, @gpetretto, for this very exciting contribution!! Unfortunately, I won't be able to provide much input at this time, but I will be monitoring the progress here.

JaGeo:
I am surely very interested. I have to see when I/we can test it. 

I have no real experience with 1. 
I think 2 is likely only important for debugging. I don't think we would need a full history of the slurm jobs that ran.

Andrew-S-Rosen:
> With this implementation I managed to run multiple atomate2 vasp Jobs in parallel in a single job submitted to SLURM.

One question I have regarding implementation: a common scenario on MP's side is to request a huge allocation (e.g. 1024 nodes) and run concurrent VASP jobs on that allocation (e.g. 4 nodes per VASP job, so 256 max concurrent jobs in one Slurm allocation). When one of those jobs finishes, we don't necessarily (in this use case) want to leave the nodes empty. We want to pull in a new job so all resources are fully utilized until the walltime is hit. Based on the description, I wasn't sure if the "continue pulling in new jobs" aspect was being covered or not.

Is this something within scope for this PR?

gpetretto:
In principle this should allow more or less what is available in fireworks. After one Job finishes another can be started. You can specify a maximum number of jobs to be executed or a timeout after which no more jobs will be started, to avoid hitting the walltime.

Having collected a few more comments, I think that at this point the best strategy would be to proceed with the review and merge of the PR. It would be easier for people to try it and give feedback once it is in an official release. In the documentation it is still specified that this shoule be considered an experimental option, as it has not been extensively tested.

Andrew-S-Rosen:
That sounds wonderful! I think releasing as experimental is probably the best way to get input. 

We will definitely try it out here but sadly not for a few months. Waiting on graduate students to join the group and our dedicated MongoDB server to be installed this month. But we will for sure be stress testing things once that's taken care of.

JaGeo:
@gpetretto I agree! We will test this as soon as it is in the main or in am official version

davidwaroquiers (approved):
As discussed, I think we can go on with this mentioning it is currently experimental as we need feedback from users. Only thing I would still change is the name of max_jobs of a given batch submission to max_jobs_per_batch as we discussed also.

PR #177, opened 2024-09-13
Bugfix: improve collection indexes to prevent duplications

gpetretto:
As pointed out by @fraricci, at the moment a code like this allows to insert multiple flows and jobs with duplicated UUIDs:
```python
j = add(1,2)
flow = Flow([j])
 
for i in range(5):
    submit_flow(flow)
```
This should not happen, because the code assumes in some points that the uuid+index should be unique for the jobs collection and uuid for the flows collection.

To prevent this, unique indexes have been added to the collections.
This also includes functionalitites to rebuild the indexes from the CLI, as users are encouraged to run `jf admin index rebuild` to prevent such issues in already existing DBs.

### DIFF
--- a/src/jobflow_remote/cli/admin.py
+++ b/src/jobflow_remote/cli/admin.py
@@ -1,4 +1,4 @@
-from typing import Annotated
+from typing import Annotated, Optional
 
 import typer
 from rich.prompt import Confirm
@@ -12,11 +12,15 @@ from jobflow_remote.cli.types import (
     flow_ids_opt,
     flow_state_opt,
     force_opt,
+    foreground_index_opt,
+    index_direction_arg,
+    index_key_arg,
     job_ids_indexes_opt,
     job_state_opt,
     start_date_opt,
 )
 from jobflow_remote.cli.utils import (
+    IndexDirection,
     check_stopped_runner,
     exit_with_error_msg,
     get_config_manager,
@@ -25,6 +29,7 @@ from jobflow_remote.cli.utils import (
     loading_spinner,
     out_console,
 )
+from jobflow_remote.jobs.data import DbCollection
 
 app_admin = JFRTyper(
     name="admin", help="Commands for administering the database", no_args_is_help=True
@@ -230,3 +235,70 @@ def unlock_flow(
         )
 
     out_console.print(f"{num_unlocked} flows were unlocked")
+
+
+app_index = JFRTyper(
+    name="index",
+    help="Commands for managing the indexes of the queue database",
+    no_args_is_help=True,
+)
+app_admin.add_typer(app_index)
+
+
+@app_index.command()
+def rebuild(
+    foreground: foreground_index_opt = False,
+):
+    """
+    Rebuild all the standard indexes.
+    Old indexes will be dropped, including the custom ones.
+    """
+    with loading_spinner(processing=False) as progress:
+        progress.add_task(description="Resetting the indexes...", total=None)
+        jc = get_job_controller()
+        jc.build_indexes(background=not foreground, drop=True)
+
+    if foreground:
+        out_console.print("Indexes rebuilt")
+    else:
+        out_console.print("Indexes rebuild started in background")
+
+
+@app_index.command()
+def create(
+    key: index_key_arg,
+    direction: index_direction_arg = IndexDirection.ASC,
+    foreground: foreground_index_opt = False,
+    collection: Annotated[
+        Optional[DbCollection],
+        typer.Option(
+            "--collection",
+            "-c",
+            help="The collection where the index should be added",
+        ),
+    ] = DbCollection.JOBS.value,  # type: ignore[assignment]
+    unique: Annotated[
+        bool,
+        typer.Option(
+            "--unique",
+            "-u",
+            help="The index will be unique",
+        ),
+    ] = False,
+):
+    """
+    Add an index to one of the queue collections
+    """
+    with loading_spinner(processing=False) as progress:
+        progress.add_task(description="Resetting the indexes...", total=None)
+        jc = get_job_controller()
+        jc.create_indexes(
+            indexes=[[(key, direction.as_pymongo)]],
+            unique=unique,
+            collection=collection,
+            background=not foreground,
+        )
+    if foreground:
+        out_console.print("Index created")
+    else:
+        out_console.print("Index creation started in background")
--- a/src/jobflow_remote/cli/types.py
+++ b/src/jobflow_remote/cli/types.py
@@ -4,7 +4,12 @@ from typing import Annotated, Optional
 import click
 import typer
 
-from jobflow_remote.cli.utils import IndexDirection, SerializeFileFormat, SortOption, str_to_dict
+from jobflow_remote.cli.utils import (
+    IndexDirection,
+    SerializeFileFormat,
+    SortOption,
+    str_to_dict,
+)
 from jobflow_remote.config.base import LogLevel
 from jobflow_remote.jobs.state import FlowState, JobState
 
--- a/src/jobflow_remote/cli/types.py
+++ b/src/jobflow_remote/cli/types.py
@@ -4,7 +4,7 @@ from typing import Annotated, Optional
 import click
 import typer
 
-from jobflow_remote.cli.utils import SerializeFileFormat, SortOption, str_to_dict
+from jobflow_remote.cli.utils import IndexDirection, SerializeFileFormat, SortOption, str_to_dict
 from jobflow_remote.config.base import LogLevel
 from jobflow_remote.jobs.state import FlowState, JobState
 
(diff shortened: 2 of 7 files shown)


src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -2482,8 +2484,21 @@ def build_indexes(
...
+            If True all existing indexes in the collections will be dropped.
         """
+
+        if drop:
+            self.jobs.drop_indexes()
+            self.flows.drop_indexes()
+            self.auxiliary.drop_indexes()
+
         self.jobs.create_index("db_id", unique=True, background=background)
+        self.jobs.create_index(
```
davidwaroquiers, line 2590: Does it actually need an index on both uuid+index here ? Maybe in case of dynamical workflows ? At least I would have thought that the example you had before would not need that. In any case, I guess it does not have much of an effect on performance to have this additional double index ?

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -2518,6 +2533,76 @@ def build_indexes(
...
+            to pymongo's create_index, thus following those conventions.
+        collection
+            The collection where the index will be created.
+        unique
+
+        background
+            If True, the indexes should be created in the background.
+        """
+        coll = self.get_collection(collection)
+        coll.create_indexes(indexes, background=background, unique=unique)
```
davidwaroquiers: This does not drop indexes right ? Docstring needs update and the method should be updated. Is this used somewhere btw ?

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -2755,8 +2840,14 @@ def add_flow(
...
         # the jobs is checked out in the meanwhile. The opposite could lead
         # to errors.
-        self.flows.insert_one(flow_doc)
-        self.jobs.insert_many(job_dicts)
+        try:
+            self.flows.insert_one(flow_doc)
+            self.jobs.insert_many(job_dicts)
+        except pymongo.errors.DuplicateKeyError as exc:
+            raise ValueError(
+                "A duplicate key error happened while inserting the flow. If this is not an unlikely "
```
davidwaroquiers: Maybe I wouldn't even mention the unlikely event of uuid overlap as it may be not clear to the average user to know what that means. What do you think ? Maybe just one comment in the code instead to mention that this is still a very unlikely possibility ? What do you think ?

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -2518,6 +2533,76 @@ def build_indexes(
             for idx in flow_custom_indexes:
                 self.flows.create_index(idx, background=background)
 
+    def create_indexes(
+        self,
+        indexes: list[str | list],
+        collection: DbCollection = DbCollection.JOBS,
+        unique: bool = False,
+        background: bool = True,
```
davidwaroquiers, line 2634: This is only done for admin/maintaining. Do we need to do that in the background ? Maybe it's better to have that always in the foreground ? What do you think ?

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -2755,8 +2840,14 @@ def add_flow(
         # should not lead to inconsistencies in the states, even if one of
         # the jobs is checked out in the meanwhile. The opposite could lead
         # to errors.
-        self.flows.insert_one(flow_doc)
-        self.jobs.insert_many(job_dicts)
+        try:
+            self.flows.insert_one(flow_doc)
+            self.jobs.insert_many(job_dicts)
```
davidwaroquiers, line 2917: If the flow can be inserted but not one of the jobs, we end up with one flow in the db without its jobs. Correct ? Could this be problematic ?

src/jobflow_remote/cli/types.py
```diff
@@ -337,6 +337,31 @@
     ),
 ]
 
+foreground_index_opt = Annotated[
```
davidwaroquiers, line 340: See my comment in the job controller. Do we need this option ? Maybe we can always have this in the foreground. What do you think ?

src/jobflow_remote/cli/admin.py
```diff
@@ -230,3 +235,70 @@ def unlock_flow(
...
+    """
+    with loading_spinner(processing=False) as progress:
+        progress.add_task(description="Resetting the indexes...", total=None)
+        jc = get_job_controller()
+        jc.build_indexes(background=not foreground, drop=True)
+
+    if foreground:
+        out_console.print("Indexes rebuilt")
+    else:
+        out_console.print("Indexes rebuild started in background")
```
davidwaroquiers, line 264: Do we need the background option ?

davidwaroquiers (commented):
Thanks for fixing this quickly!
Few comments, questions and one piece of code I think is wrong and not used (drop_index method in jobcontroller).

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: Actually this is the main point of the bug fix 😄 
I cannot put a "unique" index on uuid alone, because jobs with index>1 will have the same uuid. So the "unique" index that prevents having duplicated Jobs is this composed uuid+index.

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: I wanted to implement this, then I thought it was not worth, but forgot to remove the initial copy/paste.
I would remove unless you think it might be useful to have.
In any case these functions in `JobController` are more to expose the functionality to the CLI. A user that is expert enough to know what is an index and to get to this part of the code will likely be able to just run `job_controller.jobs.drop_index()`.

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: ok. I was also unsure about what to write exactly in this message.

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: I think that for a big DB it may take a while to build an index. I admit I made it like this mostly referring to what was done in fireworks, but I don't see a particular advantage doing it in the foreground.

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: yes, that is one case that I thought about. However this cannot happen because an issue in the submission script. 
```python
j = add(1,2)

for i in range(5):
    flow = Flow([j])
    submit_flow(flow)
```
This script will not even get to `submit_flow` in the second loop, because jobflow will directly prevent to insert the same Job in two different Flows.
I imagined that the only possible occurrence of an inconsistent state is if there is a random clash of uuid. (or if a user deliberately sets the uuid in two jobs to be equal). Can you foresee other cases where the flow uuid is different and one of the job ids is duplicated?

In that case I would really need to manually check the db for clashes or perform all the action as a MongoDB transaction. The latter might not be so bed for this action. And it is not something done in the runner that will take much time. I would maybe need to make some tests to see if this can get much slower. What do you think?

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: I see, I was only thinking about workflow creation time.

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: Yup I think you can remove

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: so you'd always do it in the background anyway ? I had the opposite idea but maybe it is stupid. I was just thinking that if you build the indices in the background and you don't know if they are already rebuild or not, what would happen ? Does it give some error or possible problem ? So if it's in the foreground, you are sure about when it is done. But you are right, for large dbs, it may be long. What do you think ? We can keep it as such anyway if you think is better this way.

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: Indeed, I forgot about jobflow already making this check. I then don't see such a case unless with a random clash or explicit uuid setting. Then it's probably not needed to do the transaction or the manual check. I would add a comment there to explain why we don't need this check (maybe some test should be added to make sure jobflow continues to behave this way ? not sure if it's needed, probably not..).

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: For custom indexes, their presence (or absence) should not add or prevent any kind of error. 
For rebuilding all the indexes, I do not really expect anybody doing this except now to fix the issue.
Normally during the the initial reset, the DB is empty and the creation of the indexes should be very fast. I would keep it this way, but I don't really have a strong opinion.

davidwaroquiers: Ok for me. Resolving

src/jobflow_remote/cli/types.py
davidwaroquiers: Resolving based on your reply in the job controller.

src/jobflow_remote/cli/admin.py
davidwaroquiers: Resovling based on your reply in the job controller comment

PR #183, opened 2024-09-18
Option to set Job priority

gpetretto:
While the `JobDoc` always had a `priority` attribute that is taken into account to sort the order of execution of jobs, there was actually no direct way of setting it. This PR introduces options to choose the priority at submission and update it in the DB.


src/jobflow_remote/jobs/submit.py
```diff
@@ -37,12 +38,14 @@ def submit_flow(
     project
         the name of the project to which the Flow should be submitted. If None the
         current project will be used.
-    exec_config: str or ExecutionConfig
+    exec_config
         the options to set before the execution of the job in the submission script.
         In addition to those defined in the Worker.
-    resources: Dict or QResources
+    resources
         information passed to qtoolkit to require the resources for the submission
```
davidwaroquiers:
```suggestion
        Information passed to qtoolkit to require the resources for the submission
```

tests/db/cli/test_job.py
```diff
@@ -210,6 +210,17 @@ def test_set_resources(job_controller, one_job) -> None:
     assert job_controller.get_job_doc(db_id="1").resources == {"ntasks": 1}
 
 
+def test_set_priority(job_controller, one_job) -> None:
+    from jobflow_remote.testing.cli import run_check_cli
+
```
davidwaroquiers, line 215:
```suggestion
    assert job_controller.get_job_doc(db_id="1").priority == 0
```

tests/db/jobs/test_jobcontroller.py
```diff
@@ -526,6 +526,9 @@ def test_set_job_run_properties(job_controller, one_job) -> None:
     assert job_controller.set_job_run_properties(resources=qr)
     assert job_controller.get_job_doc(job_id=one_job[0].uuid).resources == qr
 
+    assert job_controller.set_job_run_properties(priority=10)
+    assert job_controller.get_job_doc(job_id=one_job[0].uuid).priority == 10
+
```
davidwaroquiers, line 533: Maybe check it's not set back to 0 if we don't explicitly set priority in the set_job_run_properties?

```suggestion
    assert job_controller.get_job_doc(job_id=one_job[0].uuid).priority == 10
    assert job_controller.set_job_run_properties()
    assert job_controller.get_job_doc(job_id=one_job[0].uuid).priority == 10
```

tests/integration/test_slurm.py
```diff
@@ -305,3 +305,37 @@ def test_undefined_additional_stores(worker, job_controller) -> None:
...
+
+    runner = Runner()
+    runner.run_all_jobs(max_seconds=120)
+
+    assert job_controller.count_jobs(states=[JobState.COMPLETED]) == 3
+    # check that the jobs were executed according to the priority
+    jobs_info = job_controller.get_jobs_info()
+    jobs_info = sorted(jobs_info, key=lambda x: x.priority, reverse=True)
+    for i in range(len(jobs_info) - 1):
+        assert jobs_info[i].end_time < jobs_info[i + 1].start_time
```
davidwaroquiers: With just three jobs, maybe the probability that jobs are executed according to the priority by chance is high ? Currently, it wouldn't be the case because mongo fetches the next job from the latest modified one (I think ?) but I don't think it guarantees that (or does it ?). Maybe consider making e.g. 10 jobs with priorities from 1 to 10 in a predefined random order and check they are executed in the right order. What do you think ?

davidwaroquiers (approved):
Great thx! Just one or two comments and suggestions.
One thing I mention here (but does not need to be done in this PR) is that maybe we could/should document all the things that are in set_run_config in subsubtitles.

tests/integration/test_slurm.py
gpetretto: switched to 5 to reduce the likelyhood of random correct order, but to avoid increasing too much the number of job to be executed.

PR #187, opened 2024-09-19
Allow list of jobs from scheduler using username

gpetretto:
Some schedulers (e.g. SGE) do not allow fetching job details providing a list of job ids, which is how jobflow-remote usually gets information about the running jobs. See https://github.com/Matgenix/qtoolkit/pull/43. To avoid this issue an option to use the username as filtering option has been added. If it is present, the runner will use that instead of the list of ids. 

@QuantumChemist, do you think this will be fine for your use case?

### DIFF
--- a/src/jobflow_remote/config/base.py
+++ b/src/jobflow_remote/config/base.py
@@ -177,6 +177,12 @@ class WorkerBase(BaseModel):
         None,
         description="Options for batch execution. If define the worker will be considered a batch worker",
     )
+    scheduler_username: Optional[str] = Field(
+        None,
+        description="If defined, the list of jobs running on the worker will be fetched based on the"
+        "username instead that from the list of job ids. May be necessary for some "
+        "scheduler_type (e.g. SGE)",
+    )
     model_config = ConfigDict(extra="forbid")
 
     @field_validator("scheduler_type")
--- a/doc/source/user/projectconf.rst
+++ b/doc/source/user/projectconf.rst
@@ -103,6 +103,14 @@ configurations. Those can better be set with the :ref:`projectconf execconfig`.
     If a single worker is defined it will be used as default in the submission
     of new Flows.
 
+.. warning::
+
+    By default, jobflow-remote fetches the status of the jobs from the scheduler by
+    passing the list of ids. If the selected scheduler does not support this option
+    (e.g. SGE), it is also necessary to specify the username on the worker machine
+    through the ``scheduler_username`` option. Jobflow-remote will use that as a filter,
+    instead of the list of ids.
+
 .. _projectconf jobstore:
 
 JobStore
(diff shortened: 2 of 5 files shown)


davidwaroquiers:
Thanks @gpetretto 


This option could also enable a "global" max number of jobs per user on a worker right ? (we already talked about such an option as it is sometimes a constraint in some supercomputers).

QuantumChemist:
This sounds very good to me. I guess we simply have to test it how (in)convenient it will be in the end and take adjustments accordingly 😃

QuantumChemist:
> Thanks @gpetretto
> 
> This option could also enable a "global" max number of jobs per user on a worker right ? (we already talked about such an option as it is sometimes a constraint in some supercomputers).

that would be very useful for our clusters (that run with slurm) as well

gpetretto:
> This option could also enable a "global" max number of jobs per user on a worker right ? (we already talked about such an option as it is sometimes a constraint in some supercomputers).

Indeed it could be a step in that direction. However, the logic of the Runner is still to check only the processes corresponding to that worker, while the count of processes running on a specific runner comes from the DB. So more changes and possibly more options in the configuration will be needed to provide that functionality.

davidwaroquiers:
Sure, I wasn't suggesting to do it now but basically, we need this option before being able to possibly think about implementing/adding this feature.

src/jobflow_remote/config/base.py
```diff
@@ -177,6 +177,12 @@ class WorkerBase(BaseModel):
         None,
         description="Options for batch execution. If define the worker will be considered a batch worker",
     )
+    scheduler_username: Optional[str] = Field(
+        None,
+        description="If defined, the list of jobs running on the worker will be fetched based on the"
+        "username instead that from the list of job ids. May be necessary for some "
+        "scheduler_type (e.g. SGE)",
+    )
```
davidwaroquiers, line 185: Just one question here. For sge, it is mandatory to use this option, otherwise it won't work, should there be something to prevent *not* having it defined (or maybe jf project check can fail if it's not set ?) ?

davidwaroquiers (approved):
Thanks, Good to me, just one comment/question

QuantumChemist:
My colleague could with a lot of experimenting now figure out how you can actually do query via job list

`qstat -j 203614,203613,203612`

Screenshot 2024-09-20 at 11 48 07

Sorry for not finding this out earlier. But I still find querying via username very useful (also for my own use cases with slurm)

I will also adjust this in the qtoolkit PR!

src/jobflow_remote/config/base.py
gpetretto: It would be relatively easy to add this check. But apparently it is then possible to query by id. I am not sure if this should be a strict limitation.

PR #189, opened 2024-09-23
Require date to reset DB

### LINKED ISSUE #102: preventing reset issues
To prevent unintentional resets of production database, I suggest the user needs to type the name of the project (not just "y") in order to reset if the project has more than a handful of Flows in it.

### LINKED ISSUE #188: Set `projects_folder` via env var
Small QoL suggestion: would be nice if the `projects_folder` which currently can only be configured in `~/.jfremote.yaml` could also be set via env var `JF_REMOTE_PROJECTS_FOLDER`. one motivation for this is that some HPC setups don't provide access to the home folder from compute nodes

https://github.com/Matgenix/jobflow-remote/blob/8df197923b77862035f893d58d1cf1e022869cf1/src/jobflow_remote/config/settings.py#L19-L21

gpetretto:
After some internal discussion, we have decided to address  #102 by requiring the addition of today's date from the CLI to reduce the probability of an accidental DB reset.

This also includes an update of the documentation concerning configuration with environment variables. @janosh, can you check if this fits your expectations?

### DIFF
--- a/doc/source/user/projectconf.rst
+++ b/doc/source/user/projectconf.rst
@@ -239,28 +239,29 @@ Project specs
 
 .. _projectconf general:
 
-General Settings
-================
+General Settings - Environment variables
+========================================
 
 Aside from the project specific configuration, a few options can also be
 defined in general. There are two ways to set these options:
 
 * set the value in the ``~/.jfremote.yaml`` configuration file.
-* export the variable name prepended by the ``jfremote`` prefix::
+* set an environment variable composed by the name of the variable and
+  prepended by the ``JFREMOTE_`` prefix::
 
-    export jfremote_project=project_name
+    export JFREMOTE_PROJECT=project_name
 
 .. note::
 
-    The name of the exported variables is case-insensitive (i.e. JFREMOTE_PROJECT
+    The name of the exported variables is case-insensitive (i.e. jfremote_project
     is equally valid).
 
 The most useful variable to set is the ``project`` one, allowing to select the
 default project to be used in a multi-project environment.
 
 Other generic options are the location of the projects folder, instead of
-``~/.jfremote`` (``projects_folder``) and the path to the ``~/.jfremote.yaml``
-file itself (``config_file``).
+``~/.jfremote`` (``JFREMOTE_PROJECT_FOLDER``) and the path to the ``~/.jfremote.yaml``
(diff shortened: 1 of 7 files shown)


doc/source/user/projectconf.rst
```diff
@@ -239,28 +239,29 @@ Project specs
 
 .. _projectconf general:
 
-General Settings
-================
+General Settings - Environment variables
```
janosh, line 242: thanks, that should help with discoverability! 👍

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -2540,10 +2550,12 @@ def reset(self, reset_output: bool = False, max_limit: int = 25) -> bool:
         # what if the outputs are in other stores? Should take those as well
         if max_limit:
             n_flows = self.flows.count_documents({})
-            if n_flows >= max_limit:
+            today = datetime.now().strftime("%Y-%m-%d")
```
janosh, line 2553: not sure if that was considered but could also require typing in the number of flows to be deleted as confirmation step.

> The database contains {n_flows} flows. Type {n_flows} and press enter to confirm their deletion

janosh (approved):

davidwaroquiers (approved):
All good to me!

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: Thanks for the suggestion, but since I was already leaning towards just leaving things as they were, I hope that checking the date would be enough to prevent problematic DB resets. Do you think it would be important to have this check as well?

janosh: i meant prompting for the number of flows _instead_ of prompting for the date (to avoid any timezone ambiguities, not that i expect those to be a problem). i think an additional guard rail against accidental deletions can make sense but could also pose a hurdle for automating/testing resets where passing CLI input may be hard

gpetretto: I see. I think this is already a pretty decent improvement in the safety, so let's proceed with this version and we can reevaluate based also on usage feedbacks. Changing this again should not be a big issue. Thanks.

PR #190, opened 2024-09-23
Add backup functionality

### LINKED ISSUE #165: Add a backup feature
Would be good to have a backup command in the jf command line to dump all collections, e.g. before updating/upgrading the database.

gpetretto:
Initial implementation of a backup feature to create and restore a dump of the queue store. Closes #165.
Two options either based on the presence of the mongodump/mongorestore executable in the system, or a simple python dump to BSON files.
This may be important to have once we put in place the automatic update procedure (see  #150). It might be convenient to suggest the users to perform a backup before proceeding with the update of the DB.

### DIFF
--- a/doc/source/user/backup.rst
+++ b/doc/source/user/backup.rst
@@ -22,7 +22,7 @@ There are two options to create and restore a backup. The default relies on the 
 MongoDB tools: ``mongodump`` and ``mongorestore``. For this to work the
 `MongoDB database tools <https://www.mongodb.com/docs/database-tools>`_  need to be
 installed. The connection details provided in the project configuration will be used
-to executed the commands. This is the preferred option, since it is faster and also
+to execute the commands. This is the preferred option, since it is faster and also
 dumps and restores all the metadata of the collections. However, not all the connection
 options defined in the ``queue`` Store may be supported or it may be not possible
 to install the tools. For this reason a second option, based on a pure python implementation
--- a/doc/source/user/backup.rst
+++ b/doc/source/user/backup.rst
@@ -1,0 +1,82 @@
+.. _backup:
+
+******
+Backup
+******
+
+As explained in the :ref:`projectconf` section, jobflow-remote uses a MongoDB
+database to store the information about the state of Jobs and Flows. This is
+defined in the ``queue`` section of the project configuration.
+In several circumstances it may be required to perform a backup of this
+database. For this reason jobflow-remote offers an option to create a dump
+of the relevant collections for a project and restore it if needed.
+
+.. warning::
+    This functionality does **not create a backup of the** ``JobStore`` **containing
+    the output of the workflows**. Since the output store can be any kind of ``Store``
+    and the result may be split in the ``additional_stores``, if a backup is needed
+    it will be required to do that through the ``JobStore`` or directly with
+    the storage system.
+
+There are two options to create and restore a backup. The default relies on the official
+MongoDB tools: ``mongodump`` and ``mongorestore``. For this to work the
+`MongoDB database tools <https://www.mongodb.com/docs/database-tools>`_  need to be
+installed. The connection details provided in the project configuration will be used
+to executed the commands. This is the preferred option, since it is faster and also
+dumps and restores all the metadata of the collections. However, not all the connection
+options defined in the ``queue`` Store may be supported or it may be not possible
+to install the tools. For this reason a second option, based on a pure python implementation
+is also available. This can be activated by selecting the ``--python`` option from
+the CLI.
+
+.. warning::
+    The python version of the backup and restore will not preserve the metadata of the
+    collection. After restoring a backup with this option it would be better to
+    regenerate the standard indexes using the ``jf admin index rebuild`` command.
+
+.. note::
+    It is of course possible to manually create a backup using the MongoDB tools.
+    This jobflow-remote feature is meant to ease the procedure by automatically
+    selecting the appropriate collections to backup.
+
+Create a backup
+===============
+
+A backup can be created with the command::
+
+    jf backup create
+
+As already mentioned, this will use the ``mongodump`` executable, unless the ``--python``
+option is specified. It is possible to specify the destination path of the backup and the
+output folder contains the ``jobs.bson``, ``flows.bson`` and ``jf_auxiliary.bson``
+files. If the ``mongodump`` command is used, the folder will also contain the metadata
+files for each collection. It is also possible to request that the backup files will
+be gzipped, by adding the ``--compress`` option.
+
+.. note::
+    The folder creation follows the convention of the ``mongodump`` executable, so
+    inside the folder specified in the ``create`` command there will be a subfolder
+    with the name of the database.
+
+.. note::
+    The name of the files will be the standard ones, even if the names of the collections
+    defined in the project configuration file are different.
+
+Restore a backup
+================
+
+To restore a backup the following command can be used::
+
+    jf backup restore /path/to/backup/folder
+
+The path should point to the folder containing the bson files generated during the creation.
+The code will automatically determine if the files are zipped, based on their extension.
+
+.. note::
+    The name of the target collection are determined by the values defined in the project
+    settings, not by the names of the files, nor by the names of the collections from
+    which the backup was created.
+
+.. note::
+    The backup can be restored only in an empty database. The code will raise an error
+    if the target database already contains jobs and flows.
--- a/src/jobflow_remote/cli/backup.py
+++ b/src/jobflow_remote/cli/backup.py
@@ -1,0 +1,134 @@
+from typing import Annotated, Optional
+
+import typer
+
+from jobflow_remote.cli.jf import app
+from jobflow_remote.cli.jfr_typer import JFRTyper
+from jobflow_remote.cli.utils import (
+    check_stopped_runner,
+    get_job_controller,
+    loading_spinner,
+    out_console,
+)
+
+app_backup = JFRTyper(
+    name="backup",
+    help="Commands for handling backup of the database",
+    no_args_is_help=True,
+)
+app.add_typer(app_backup)
+
+
+@app_backup.command()
+def create(
+    out_path: Annotated[
+        str,
+        typer.Argument(
+            help="The path of the folder where the output files will be saved",
+            metavar="PATH",
+        ),
+    ] = ".",
+    compress: Annotated[
+        bool,
+        typer.Option(
+            "--compress",
+            "-c",
+            help="Compress the output files",
+        ),
+    ] = False,
+    mongo_dir: Annotated[
+        Optional[str],
+        typer.Option(
+            "--mongo-dir",
+            "-m",
+            help=(
+                "The path to a folder containing the mongodump executable, if not present in the PATH"
+            ),
+        ),
+    ] = None,
+    python: Annotated[
+        bool,
+        typer.Option(
+            "--python",
+            "-p",
+            help="If selected a python implementation will be used to create a backup. "
+            "WARNING: In this case metadata of the collections will not be saved",
+        ),
+    ] = False,
+) -> None:
+    """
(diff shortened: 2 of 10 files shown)


src/jobflow_remote/utils/db.py
```diff
@@ -398,3 +406,213 @@ def from_flow_doc(cls, doc: dict, additional_msg: str | None = None):
         if additional_msg:
             msg += " " + additional_msg
         return cls(msg)
+
+
+def mongo_operation(
+    store: MongoStore | MongoURIStore,
+    collection: str,
```
davidwaroquiers: Maybe have a default None that will take the collection directly from the store ? Not sure if it's "essential". I let you decide.

src/jobflow_remote/utils/db.py
```diff
@@ -398,3 +406,213 @@ def from_flow_doc(cls, doc: dict, additional_msg: str | None = None):
...
+    -------
+    tuple[str, str]
+        The stdout and stderr of the executed command.
+    """
+    if operation not in ("mongodump", "mongorestore"):
+        raise ValueError(f"Operation {operation} not supported")
+
+    if mongo_bin_path:
+        operation = os.path.join(mongo_bin_path, operation)
+
```
davidwaroquiers, line 450: Maybe here check that the operation "exists" (is in the path) using shutil.which so that it does not go through all the rest ?

src/jobflow_remote/utils/db.py
```diff
@@ -398,3 +406,213 @@ def from_flow_doc(cls, doc: dict, additional_msg: str | None = None):
...
+        ]
+    elif type(store) is MongoStore:
+        cmd = [
+            operation,
+            "--host",
+            store.host,
+            "--port",
+            str(store.port),
+        ]
+        if store.username and store.password:
```
davidwaroquiers, line 472: Is it possible to pass only username (no password in mongodb for the user ?) or only password (default user taken from ... where ? no idea) ?
If not, maybe the user should be warned that he has to give both ?

src/jobflow_remote/utils/db.py
```diff
@@ -398,3 +406,213 @@ def from_flow_doc(cls, doc: dict, additional_msg: str | None = None):
...
+    if compress:
+        cmd.append("--gzip")
+
+    if operation.endswith("mongodump"):
+        cmd.extend(["--out", file_path])
+    elif operation.endswith("mongorestore"):
+        cmd.append(file_path)
+
+    str_cmd = " ".join(str(s) for s in cmd)
+    result = subprocess.run(
```
davidwaroquiers, line 503: Just wondering why you do it with a string (i.e. text=True) instead of with the list (and then with shell=False I think, but maybe you need shell=True for some reason ?) ? Or maybe is it because then you have the logger that uses the str_cmd ?

src/jobflow_remote/utils/db.py
```diff
@@ -398,3 +406,213 @@ def from_flow_doc(cls, doc: dict, additional_msg: str | None = None):
...
+        str_cmd, check=True, capture_output=True, text=True, shell=True
+    )
+    logger.debug(
+        f"output during execution of {str_cmd}. Stdout: {result.stdout}. Stderr: {result.stderr}"
```
davidwaroquiers: I'd probably put the command in between quotes (single or double, whichever one). What do you think ?

src/jobflow_remote/utils/db.py
```diff
@@ -398,3 +406,213 @@ def from_flow_doc(cls, doc: dict, additional_msg: str | None = None):
...
+        The store where the data should be restored.
+    collection
+        The name of the collection to be restored.
+    input_file
+        The path of the BSON file containing the data to be restored.
+    mongo_bin_path
+        The path to the folder containing the mongo executable. If None, the executable is
+        searched in the PATH.
+    compress
+        If True, the input file is expected to be compressed.
```
davidwaroquiers: Could it possibly figure out itself that it is compressed or not ? Or do you think it's better not and keep it explicit ?

src/jobflow_remote/utils/db.py
davidwaroquiers: Or maybe this could/should be done in the JobController methods ? But then the logic of having the operation has to be moved or replicated or something ?

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -4105,6 +4113,163 @@ def delete_jobs(
             max_limit=max_limit,
         )
 
+    def backup_dump(
```
davidwaroquiers, line 4129: Just wondering if there should be an option to also backup the results ? Let's suppose someone "thinks" it would be done and uses that functionality to transfer everything from one server to another (maybe it's not the best way to do this but anyway), then dumps. Copies the files to the new server, and then cannot access its outputs with the new.

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -4105,6 +4113,163 @@ def delete_jobs(
...
+        dir_path = Path(dir_path)
+        doc_count = {}
+        collection_names = ["jobs", "flows", "jf_auxiliary"]
+        if python:
+            for name, collection in zip(
+                collection_names, [self.jobs, self.flows, self.auxiliary]
+            ):
+                doc_count[name] = pymongo_dump(
+                    collection=collection, output_path=dir_path, compress=compress
+                )
```
davidwaroquiers: I'd probably rename a few of the variables. Somehow the "collection_names" is not collection names, only the type of thing or type of collection that is there, while the actual collection names are the self.jobs_collection, self.flows_collection and self.auxiliary_collection.

src/jobflow_remote/cli/backup.py
```diff
@@ -0,0 +1,134 @@
...
+    ] = ".",
+    compress: Annotated[
+        bool,
+        typer.Option(
+            "--compress",
+            "-c",
+            help="Compress the output files",
+        ),
+    ] = False,
+    mongo_dir: Annotated[
```
davidwaroquiers: Maybe call it mongo-path here ? I would probably call out_dir and mongo_path the first one is the directory in which output files will be saved, the second one is the path in which mongo commands are looked for. I'm fine with anything anyway but maybe it's a bit more clear ?

src/jobflow_remote/cli/backup.py
```diff
@@ -0,0 +1,134 @@
...
+@app_backup.command()
+def restore(
+    out_path: Annotated[
+        str,
+        typer.Argument(
+            help="The path of the folder where the backup files are located",
+            metavar="PATH",
+        ),
+    ] = ".",
+    compress: Annotated[
```
davidwaroquiers: I made the comment in the job controller, but could this be figured out automatically ?

src/jobflow_remote/cli/backup.py
```diff
@@ -0,0 +1,134 @@
...
+        Optional[str],
+        typer.Option(
+            "--mongo-dir",
+            "-m",
+            help=(
+                "The path to a folder containing the mongorestore executable, if not present in the PATH"
+            ),
+        ),
+    ] = None,
+    python: Annotated[
```
davidwaroquiers, line 101: I did not make the comment in the job controller, but almso here, could this be figured out automatically ?

davidwaroquiers (approved):
Great addition! A few comments/questions/suggestions here and there

src/jobflow_remote/utils/db.py
gpetretto: I added this option

src/jobflow_remote/utils/db.py
gpetretto: added the check

src/jobflow_remote/utils/db.py
gpetretto: it seems that having only of of them may cause issues. Added a check that either both or none of them should be present

src/jobflow_remote/utils/db.py
gpetretto: acutally if I use shell=False I get a complain from ruff (this rule: https://docs.astral.sh/ruff/rules/subprocess-without-shell-equals-true/). In this way it works and does ruff not raise issues.

src/jobflow_remote/utils/db.py
gpetretto: added

src/jobflow_remote/utils/db.py
gpetretto: now it figures it out based on the file extension. Can still be set explicitly

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: This is problematic, since making the backup of all the additional stores of the JobStore might not be trivial. Since the queue part is the only one important for the update feature I would not implement this here.

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: done

src/jobflow_remote/cli/backup.py
gpetretto: done

src/jobflow_remote/cli/backup.py
gpetretto: done. argument removed here

src/jobflow_remote/cli/backup.py
gpetretto: This seems a bit more complicated. I am not sure about how to figure this out. Also, in principle it would not be impossible to mix things up.

gpetretto:
I addressed the comments above. I still need to prepare the documentation of the feature

src/jobflow_remote/cli/backup.py
davidwaroquiers: Which one did you decide in the end? Seems nothing was changed. Just resolve the comment if you think it is fine to say as it is 😀

src/jobflow_remote/cli/backup.py
davidwaroquiers: Ok

src/jobflow_remote/utils/db.py
davidwaroquiers: 👍

src/jobflow_remote/utils/db.py
davidwaroquiers: Ok

src/jobflow_remote/utils/db.py
davidwaroquiers: Ok

src/jobflow_remote/cli/backup.py
gpetretto: Sorry, I had changed all the variable names except this one. It should be fine now.

doc/source/user/backup.rst
```diff
@@ -0,0 +1,82 @@
...
+    the output of the workflows**. Since the output store can be any kind of ``Store``
+    and the result may be split in the ``additional_stores``, if a backup is needed
+    it will be required to do that through the ``JobStore`` or directly with
+    the storage system.
+
+There are two options to create and restore a backup. The default relies on the official
+MongoDB tools: ``mongodump`` and ``mongorestore``. For this to work the
+`MongoDB database tools <https://www.mongodb.com/docs/database-tools>`_  need to be
+installed. The connection details provided in the project configuration will be used
+to executed the commands. This is the preferred option, since it is faster and also
```
davidwaroquiers:
```suggestion
to execute the commands. This is the preferred option, since it is faster and also
```

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: I agree, let's keep it like that.

davidwaroquiers (approved):
All good to me! Great doc.

PR #191, opened 2024-09-25
Add sanitization option to host outputs

gpetretto:
An issue was raised concerning the execution of commands through the Host objects. If some text is added to the output of a remote host when executing a command (e.g. the disk quota) the qtk parsers fail to extract the output of the commands.
This PR adds an option to sanitize the output by prepending and appending and `echo` of a string to the command execution, both in the stdout and stderr. The outputs are then filtered to only considere the text enclosed inside the occurrences of string.

At the moment there is only an integration test checking that with the sanitization option enabled the standard execution still works. @ml-evs is there any easy way of adding a test that will actually check that if there are additional strings returned the command will still work fine? Otherwise I will just either mock the return of the function or test the sanitize method of `BaseHost` individually.


ml-evs:
> At the moment there is only an integration test checking that with the sanitization option enabled the standard execution still works. @ml-evs is there any easy way of adding a test that will actually check that if there are additional strings returned the command will still work fine? Otherwise I will just either mock the return of the function or test the sanitize method of `BaseHost` individually.

Probably both approaches would make sense? It would be straightforward to add some additional quota-like output on the integration tests (i.e. properly replicating the setup on clusters where it's done at the global bashrc level) but for debugging purposes a mocked unit test would be nice too -- the integration test with bashrc will have to wait until I've finished the other PR (which is close).

tests/integration/test_slurm.py
```diff
@@ -362,3 +362,26 @@ def patched_get_jobs_list(self, jobs=None, user=None, timeout=None):
...
+
+    assert job_controller.project.workers[worker].get_host().sanitize is True
+
+    flow = Flow([add(1, 2)])
+    submit_flow(flow, worker=worker)
+
+    runner = Runner()
+    runner.run_one_job()
+
+    assert job_controller.count_jobs(states=JobState.COMPLETED) == 1
```
davidwaroquiers, line 434: The worker here does not have spurious echos here. Would we need something like that ? Then that's another worker that's spawned (or maybe the worker could be modified temporarily ? like modifying the .bashrc or bash profile i don't know and then putting it back to "normal" after the test). Not sure if it is worth anyway, just wondering.

davidwaroquiers (approved):
All good to me!
Just one thing about what we discussed, to probably add a specific check test in jf project check that tells the user what to try (i.e. setting sanitize_command to True) in case there is such spurious echos.

gpetretto:
I changed the way the prepend and append strings are generated, in order to be more brodly compatible with different shells. For future reference, we may also consider executing the commands in the local worker through the `bash -c` command. 

I will wait for confirmation that this still works fine in a real case before mergin.

Concerning the test, if in the meanwhile the PR with the update of the integration tests is merged we can consider adding specific tests for this feature.

davidwaroquiers:
> I changed the way the prepend and append strings are generated, in order to be more brodly compatible with different shells. For future reference, we may also consider executing the commands in the local worker through the `bash -c` command.
> 
> I will wait for confirmation that this still works fine in a real case before mergin.
> 
> Concerning the test, if in the meanwhile the PR with the update of the integration tests is merged we can consider adding specific tests for this feature.

Ok great, I resolved the comment on a possible specific integration test for this (I'm not even sure it is really needed, but let's see anyway after it has been confirmed and the other PR merged).

PR #201, opened 2024-10-25
Delete job files when rerunning; customizable execution command

### LINKED ISSUE #196: call jf from container
Hi, all

Due to the requirement of the HPC I am using, I had to install it in a container. I found that in the interactive shell mode, I could run jf command in this way:
`singularity exec ~/jobflowenv.sif jf` and it worked as expected.

However, it didn't work if I tried execute it in the slurm job script.
Here's my job script submit.sh
```
#!/bin/bash

#SBATCH --partition=debug
#SBATCH --job-name=relax_job
#SBATCH --nodes=1
#SBATCH --ntasks-per-node=4
#SBATCH --mem=30gb
#SBATCH --time=00:10:00
#SBATCH --output=/scratch/qejobflow/87/50/55/875055c1-7fa4-4a40-bee5-6ddb63be599a_1/queue.out
#SBATCH --error=/scratch/qejobflow/87/50/55/875055c1-7fa4-4a40-bee5-6ddb63be599a_1/queue.err
cd /scratch/qejobflow/87/50/55/875055c1-7fa4-4a40-bee5-6ddb63be599a_1/
module purge
module load QuantumESPRESSO
alias jf="singularity exec ~/jobflowenv.sif jf"
alias python="singularity exec ~/jobflowenv.sif python"
singularity exec ~/jobflowenv.sif python --version
singularity exec ~/jobflowenv.sif jf --help

jf -fe execution run /scratch/qejobflow/87/50/55/875055c1-7fa4-4a40-bee5-6ddb63be599a_1
```
By examining queue.out, I found the lines beginning with 'singularity' in the submit.sh worked and printed the information about python version and help information of jobflow-remote. However, the last line `jf -fe execution run /scratch/qejobflow/87/50/55/875055c1-7fa4-4a40-bee5-6ddb63be599a_1` didn't work and it raised an error "/var/spool/slurmd/job8184623/slurm_script: line 24: jf: command not found
" in the queue.err file.

I am wondering if anyone has similar experience. I'll greatly appreciate if anyone could make some comments!

### LINKED ISSUE #197: Delete VASP files when rerunnning jobs?
Hello everyone 😃 ,

I wanted to know if there is a possibility to delete all (VASP) files automatically when rerunning a job? 

For now, when I want to rerun a VASP job, I have to manually delete the old VASP output files or the job will fail because there is a problem with overwriting the old files. With other job types, there is no such problem. 

Thank you for your help in advance!

gpetretto:
Closing #197 and #196.

### Delete files on rerun
As discussed in #197, this will remove the files from the worker when a Job is rerun. 
1) The removal is applied by default, but can be deactivated
2) If an error happens while removing the files, the Job will not be rerun
3) If failing to remove files due to the fact that the `jfremote_in.json` is missing a warning will be printed, but Job will be rerun

Note that this functionality might have a direct impact for @JaGeo: now if a worker has a MFA when rerunning a job it will ask to insert the OTP. 

Also I am not 100% convinced about point 2, if it instead should print a warning and still rerun the Job. 

Also pinging @QuantumChemist @utf. 

### Customizable execution command

As discussed in #196 added a new option to customize the `jf -fe execution run`: `execution_cmd`

@yw-fang would this be fine?

### DIFF
--- a/src/jobflow_remote/config/base.py
+++ b/src/jobflow_remote/config/base.py
@@ -215,6 +215,16 @@ class WorkerBase(BaseModel):
             raise ValueError("`work_dir` must be an absolute path")
         return v
 
+    @field_validator("execution_cmd")
+    @classmethod
+    def check_execution_cmd(cls, v) -> Optional[str]:
+        if v is not None and "{}" not in v:
+            raise ValueError(
+                "`execution_cmd` must contain a '{}' part to allow setting "
+                "the execution folder at runtime"
+            )
+        return v
+
     def get_scheduler_io(self) -> BaseSchedulerIO:
         """
         Get the BaseSchedulerIO from QToolKit depending on scheduler_type.
--- a/src/jobflow_remote/config/base.py
+++ b/src/jobflow_remote/config/base.py
@@ -166,6 +166,13 @@ class WorkerBase(BaseModel):
         None,
         description="String with commands that will be executed after the execution of the Job",
     )
+    execution_cmd: Optional[str] = Field(
+        None,
+        description="String with commands to execute the Job on the worker. By default will be "
+        "set to `jf -fe execution run {}`. The `{}` part will be used to insert the path to "
+        "the execution directory and it is mandatory. Change only for specific needs (e.g. the"
+        "jf command needs to be executed in a container).",
+    )
     timeout_execute: int = Field(
         60,
         description="Timeout for the execution of the commands in the worker "
@@ -637,6 +644,16 @@ class Project(BaseModel):
                     f"error while converting jobstore to JobStore. Error: {traceback.format_exc()}"
                 ) from e
         return jobstore
+
+    @property
+    def has_interactive_workers(self) -> bool:
+        """
+        True if any of the workers have interactive_login set to True.
+        """
+        for worker in self.workers.values():
+            if isinstance(worker, RemoteWorker) and worker.interactive_login:
+                return True
+        return False
 
     model_config = ConfigDict(extra="forbid")
 
--- a/src/jobflow_remote/jobs/data.py
+++ b/src/jobflow_remote/jobs/data.py
@@ -119,6 +119,7 @@ class RemoteInfo(BaseModel):
     process_id: Optional[str] = None
     retry_time_limit: Optional[datetime] = None
     error: Optional[str] = None
+    cleanup: bool = False
 
 
 class JobInfo(BaseModel):
--- a/src/jobflow_remote/jobs/data.py
+++ b/src/jobflow_remote/jobs/data.py
@@ -119,7 +119,7 @@ class RemoteInfo(BaseModel):
     process_id: Optional[str] = None
     retry_time_limit: Optional[datetime] = None
     error: Optional[str] = None
-    cleanup: bool = False
+    prerun_cleanup: bool = False
 
 
 class JobInfo(BaseModel):
--- a/src/jobflow_remote/jobs/jobcontroller.py
+++ b/src/jobflow_remote/jobs/jobcontroller.py
@@ -2467,7 +2467,7 @@ class JobController:
         return result.modified_count
 
     def _safe_delete_files(
-        self, jobs_info: list[JobInfo | dict]
+        self, jobs_info: Sequence[JobInfo | dict]
     ) -> list[JobInfo | dict]:
         """
         Delete the files associated to the selected Jobs.
--- a/src/jobflow_remote/jobs/jobcontroller.py
+++ b/src/jobflow_remote/jobs/jobcontroller.py
@@ -68,7 +68,7 @@ from jobflow_remote.utils.db import (
     pymongo_dump,
     pymongo_restore,
 )
-from jobflow_remote.utils.remote import SharedHosts
+from jobflow_remote.utils.remote import SharedHosts, safe_remove_job_files
 
 if TYPE_CHECKING:
     from collections.abc import Generator, Sequence
@@ -839,27 +839,25 @@ class JobController:
         list
             List of db_ids of the updated Jobs.
         """
-        # Open the SharedHosts so that hosts will be shared for all the Jobs
-        with SharedHosts(self.project):
-            return self._many_jobs_action(
-                method=self.rerun_job,
-                action_description="rerunning",
-                job_ids=job_ids,
-                db_ids=db_ids,
-                flow_ids=flow_ids,
-                states=states,
-                start_date=start_date,
-                end_date=end_date,
-                name=name,
-                metadata=metadata,
-                workers=workers,
-                custom_query=custom_query,
-                raise_on_error=raise_on_error,
-                force=force,
-                wait=wait,
-                break_lock=break_lock,
-                delete_files=delete_files,
-            )
+        return self._many_jobs_action(
+            method=self.rerun_job,
+            action_description="rerunning",
+            job_ids=job_ids,
+            db_ids=db_ids,
+            flow_ids=flow_ids,
+            states=states,
+            start_date=start_date,
+            end_date=end_date,
+            name=name,
+            metadata=metadata,
+            workers=workers,
+            custom_query=custom_query,
+            raise_on_error=raise_on_error,
+            force=force,
+            wait=wait,
+            break_lock=break_lock,
+            delete_files=delete_files,
+        )
 
     def rerun_job(
         self,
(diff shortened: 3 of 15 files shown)


yw-fang:
> fe

I like this added feature! Thanks very much!

yw-fang:
I like this added feature! Thanks very much @gpetretto

JaGeo:
@gpetretto oh, i see. This might be a lot of work.

gpetretto:
> @gpetretto oh, i see. This might be a lot of work.

There is a `--no-delete` option to skip the deletion. Not sure if that would be convenient enough for you. Otherwise we may consider adding a CLI setting to determine the default behaviour. See for example: https://github.com/Matgenix/jobflow-remote/blob/82457776ed3ffa5fabca74daa4ee7af24cbaab4e/src/jobflow_remote/config/settings.py#L27

davidwaroquiers:
To comment on the problem with the OTP, one thing we discussed with @gpetretto is to also implement the deletion of the files when a job starts. This could be activated through, e.g. a "rerun_delete_strategy" option or something like that. Default would be e.g. "UPON_RERUN" with other two options being "NONE" (don't delete files) and "WHEN_STARTING". In your case @JaGeo, you'd be using this second option. I don't remember what were the downsides of this.

src/jobflow_remote/config/base.py
```diff
@@ -166,6 +166,13 @@ class WorkerBase(BaseModel):
         None,
         description="String with commands that will be executed after the execution of the Job",
     )
+    execution_cmd: Optional[str] = Field(
+        None,
+        description="String with commands to execute the Job on the worker. By default will be "
+        "set to `jf -fe execution run {}`. The `{}` part will be used to insert the path to "
```
davidwaroquiers, line 172: Why not having the default "jf -fe execution run {}" set here ?

src/jobflow_remote/jobs/runner.py
```diff
@@ -637,7 +637,8 @@ def submit(self, lock: MongoLock) -> None:
 
         remote_path = Path(doc["run_dir"])
 
-        script_commands = [f"jf -fe execution run {remote_path}"]
+        execution_cmd = worker.execution_cmd or "jf -fe execution run {}"
```
davidwaroquiers, line 670: If the default is set in the config base, no need here. (but see comment in config base, is there a reason to put the default here and not in config base ?)

src/jobflow_remote/config/base.py
davidwaroquiers: Could/should we have a validation of this somewhere ? In jf project check ? Or everytime it is used ?

src/jobflow_remote/config/base.py
```diff
@@ -166,6 +166,13 @@ class WorkerBase(BaseModel):
...
+        "the execution directory and it is mandatory. Change only for specific needs (e.g. the"
+        "jf command needs to be executed in a container).",
```
davidwaroquiers, line 174:
```suggestion
        "the execution directory and it is mandatory. Change only for specific needs (e.g. in "
        "case the jf command needs to be executed in a container).",
```

JaGeo:
@davidwaroquiers i see. Thanks! Might be good to have a detailed documentation on this then.

src/jobflow_remote/utils/remote.py
```diff
@@ -0,0 +1,104 @@
...
+    to share the hosts across different sections of the code, if needed.
+    Hosts connections are all closed only when leaving the last context
+    manager.
+
+    Examples
+    --------
+
+    >>> with SharedHosts(project) as shared_hosts:
+    ...     host = shared_hosts.get_host("worker_name")
+    ...     # Use host as required
```
davidwaroquiers, line 32: Maybe it would be good to add a unit test ?

davidwaroquiers (approved):
Great! Just a few things here and there. + comment regarding OTP.

gpetretto:
After some discussion with @davidwaroquiers, we have thought that having multiple options might be confusing and it is not entirely clear if there is a benefit. So, instead, we could entirely remove the deletion process at the time the `jf job rerun` command is executed and the folder will instead be cleaned by the runner during the UPLOAD phase, before uploading the data. In that case we would leave this as the only option, so there is no ambiguity on how things are performed.
This will avoid entirely the MFA issue. Oher advantages of this option are that it will speed up the `rerun` command and it will remove a potential point of failure in a tricky process as the rerun of multiple jobs. The `--no-delete` option will still be available. If the deletion fails it will go in a REMOTE_ERROR state and the user might need to manually clean up the folder.
Of course the whole procedure will be described in the documentation.

@JaGeo, @utf do you see any reason to keep the file deletion in the `rerun` command as an alternative option?

JaGeo:
@gpetretto yes, this sounds like the best solution!

src/jobflow_remote/config/base.py
gpetretto: If you set the default to a value different from `None` and you generate a project with something like `jf project generate`, or create it from the python interface and dump it, the value will be explicitly set in the yaml file. I see two downsides there: 1) making the config a bit more cramped 2) this is a setting that it is extremely unlikely to be changed. If for any reason we need to make a change to the default in the future, this is a potential source for backward incompatibility issues. I know we prepared the upgrade procedure, but if we can reduce the amount of times this needs to be used it would be better. So I decided to set the default to `None` and describe the default in the doc. 

For the validation, I had also considered and not added it for the usual fear of overloading the process of the Project validation process. Again, since this is a very unlikely option to be set I imagined that this will almost always be a wasted check. It is also true that if the default value is None there is also nothing to check, so the overhead should be negligible. So it can still be added. In that case I would just check that if it is not `None` there should be a `{}` inside the string. Or should it contain the full string `jf -fe execution run {}`? I wonder if there might be cases where one needs to write the pieces of the command with some kind of separator.

davidwaroquiers: I see, good point to keep it outside then. You can add the validation if it's not None as you mention, I think it may help (and it should be easy to implement right ?) : there should be one and only one {} inside the string. (I would indeed not go the the fact that it shoud contain the full string).

src/jobflow_remote/jobs/runner.py
davidwaroquiers: closing as you answered in the other comment

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -3881,8 +3881,14 @@ def lock_job_for_update(
                 no_retry = True
             except RemoteError as e:
                 error = f"Remote error: {e.msg}"
-                if e.__cause__:
-                    trace = traceback.format_exception(e.__cause__)
+                cause = e.__cause__
+                if cause:
+                    # this is required for support of python 3.9. In 3.10 the API
+                    # changed and format_exception(e) could be used instead.
+                    # Do that if/when support for 3.9 is dropped.
```
davidwaroquiers, line 3889: I would not drop it now but just to mention, pymatgen is python >= 3.10. Maybe we should btw add python 3.12 now in the testing ?

src/jobflow_remote/jobs/data.py
```diff
@@ -119,6 +119,7 @@ class RemoteInfo(BaseModel):
     process_id: Optional[str] = None
     retry_time_limit: Optional[datetime] = None
     error: Optional[str] = None
+    cleanup: bool = False
```
davidwaroquiers: Should this be "cleanup_on_rerun" or something ? (if we were to add something to cleanup files after a job has run)

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -904,6 +908,8 @@ def rerun_job(
             Forcibly break the lock on locked documents. Use with care and
             verify that the lock has been set by a process that is not running
             anymore. Doing otherwise will likely lead to inconsistencies in the DB.
+        delete_files
+            Delete all the files in the worker folder of the rerun Job.
```
davidwaroquiers, line 912:
```suggestion
            Delete all the files in the worker folder of the rerun Job.
            Note that the deletion will not be performed directly but only when the job effectively restarts.
```

davidwaroquiers: I leave it to you to include this additional comment or not.

davidwaroquiers (approved):
Great! One or two suggestions for comments (up to you). Also one comment about python 3.12 (maybe not needed for this specific PR ?)

src/jobflow_remote/jobs/data.py
gpetretto: renamed to `prerun_cleanup`

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: Indeed that is a good point. Jobflow is still declared 3.9+, so I would keep aligned with that, as long as there are no strong limitations for jobflow-remote. Maybe adding 3.12 could be discussed in #160?

src/jobflow_remote/utils/remote.py
gpetretto: done

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: Good point, I'll add a comment there.

PR #228, opened 2024-12-10
Option for stored data in `jf job list`

### LINKED ISSUE #227: Check status for simulation code in `jf job list`
This is a basically a feature request, which I guess is not so easy to implement. But maybe someone knows an easier way to achieve what I want.

Here is the story: I run mainly a simulation code with jobflow-remote, so something like VASP or Abinit. Usually a single, lets say VASP, calculation corresponds to one jobflow job. Now looking at the result with `jf job info` or `jf job list`, I see that the job(s) is COMPLETED. This is good, of course. But it does only mean that there was no python error. Basically I catch any error in the simulation code via python to ensure appropriate error handling and a proper entry in the output database. What I would like to see also is directly a hint whether the actual calculation from the simulation code was successful or not.

So one question would be on how to pass this information to jobflow(-remote) since it is a very specific information. I could imagine creating directly during runtime a special file `success.out` which contains a status string or whatever. Another option might be to somehow return a `Response` object with a new field `simulation_success`. But this might be already to specific. Then jobflow-remote would need to parse this and include in the JobDoc which is quite some overhead I guess.

Another option would be to generate this info only locally while running the `jf job list` command. But this again requires a very specific format to be kept. In principle I could provide a check function for my simulation code which takes the run_dir and the stdout/stderr and determines the run success. But as it needs to read different files depending on rather complex logic, one would need to download all those files which is also not a very good option I guess.

Any opinion?

FabiPi3:
Add the option to display data from the `stored_data` field in the `Response` object. A User can select a key or multiple keys to display those values.

On a sidenote, another option was added to not display the uuid of the jobs. I typically don't gain much from this information since the db_id already gives me the unique information I need to find the job in the database.

Things to discuss:

- Name of the cli options (including short-cuts)
- Name of the column in the output table: For now only the key. Should be sufficient since the user does know what fields he/she requested, right?
- Name of the default_value which is displayed once no data is available or the specified key is not present. For now a light grey italic `none`

Example:

See the attached screenshot.

Bildschirmfoto 2024-12-10 um 11 32 32

Maybe closes #227

### DIFF
--- a/src/jobflow_remote/cli/formatting.py
+++ b/src/jobflow_remote/cli/formatting.py
@@ -26,99 +26,100 @@ if TYPE_CHECKING:
     from jobflow_remote.jobs.upgrade import UpgradeAction
 
 
+def format_state(ji: JobInfo) -> Text:
+    state = ji.state.name
+    if ji.state in (JobState.REMOTE_ERROR, JobState.FAILED):
+        state = f"[bold red]{state}[/]"
+    elif ji.remote.retry_time_limit is not None:
+        state = f"[bold orange3]{state}[/]"
+    return Text.from_markup(state)
+
+
+def format_run_time(ji: JobInfo) -> str:
+    prefix = ""
+    if ji.state == JobState.RUNNING:
+        run_time = ji.estimated_run_time
+        prefix = "~"
+    else:
+        run_time = ji.run_time
+    if not run_time:
+        return ""
+    m, s = divmod(run_time, 60)
+    h, m = divmod(m, 60)
+    return prefix + f"{h:g}:{m:02g}"
+
+
+time_zone_str = f" [{time.tzname[0]}]"
+header_name_data_getter_map = {
+    "db_id": ("DB id", lambda ji: str(ji.db_id)),
+    "name": ("Name", lambda ji: ji.name),
+    "state": ("State", format_state),
+    "job_id": ("Job id (Index)", lambda ji: f"{ji.uuid}  ({ji.index})"),
+    "worker": ("Worker", lambda ji: ji.worker),
+    "last_updated": (
+        "Last updated" + time_zone_str,
+        lambda ji: convert_utc_time(ji.updated_on).strftime(fmt_datetime),
+    ),
+    "queue_id": ("Queue id", lambda ji: ji.remote.process_id),
+    "run_time": ("Run time", format_run_time),
+    "retry_time": (
+        "Retry time" + time_zone_str,
+        lambda ji: convert_utc_time(ji.remote.retry_time_limit).strftime(fmt_datetime)
+        if ji.remote.retry_time_limit
+        else None,
+    ),
+    "prev_state": (
+        "Prev state",
+        lambda ji: ji.previous_state.name if ji.previous_state else None,
+    ),
+    "locked": ("Locked", lambda ji: "*" if ji.lock_id is not None else None),
+    "lock_id": ("Lock id", lambda ji: str(ji.lock_id)),
+    "lock_time": (
+        "Lock time" + time_zone_str,
+        lambda ji: convert_utc_time(ji.lock_time).strftime(fmt_datetime)
+        if ji.lock_time
+        else None,
+    ),
+}
+
+
 def get_job_info_table(
     jobs_info: list[JobInfo],
     verbosity: int,
+    header_keys: list[str] | None = None,
     stored_data_keys: list[str] | None = None,
-    skip_job_id: bool = False,
 ) -> Table:
-    time_zone_str = f" [{time.tzname[0]}]"
+    stored_data_keys = stored_data_keys or []
+    all_header_keys = list(header_name_data_getter_map)
+    if not header_keys:
+        header_keys = all_header_keys[:6]
+        if verbosity >= 1:
+            header_keys += all_header_keys[6:10]
+        if verbosity == 1:
+            header_keys.append(all_header_keys[10])
+        if verbosity >= 2:
+            header_keys += all_header_keys[11:13]
 
     table = Table(title="Jobs info")
-    table.add_column("DB id")
-    table.add_column("Name")
-    table.add_column("State")
-    if not skip_job_id:
-        table.add_column("Job id  (Index)")
+    for key in header_keys:
+        table.add_column(header_name_data_getter_map[key][0])
+    for key in stored_data_keys:
+        table.add_column(key)
 
-    table.add_column("Worker")
-    table.add_column("Last updated" + time_zone_str)
-
-    if verbosity >= 1:
-        table.add_column("Queue id")
-        table.add_column("Run time")
-        table.add_column("Retry time" + time_zone_str)
-        table.add_column("Prev state")
-        if verbosity < 2:
-            table.add_column("Locked")
-
-    if verbosity >= 2:
-        table.add_column("Lock id")
-        table.add_column("Lock time" + time_zone_str)
-
-    if stored_data_keys:
-        for key in stored_data_keys:
-            table.add_column(f"{key}")
+    default_stored_data_value = Text.from_markup("[italic grey58]none[/]")
 
     for ji in jobs_info:
-        state = ji.state.name
-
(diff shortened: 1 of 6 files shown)


gpetretto:
Hi @FabiPi3,

sorry for the delay, but I did not have much time to dedicate to jobflow-remote in the last days. I think the option for the `stored_data` is a good update.
Concerning the job id, I think it would be good to introduce a better customization of the displayed result, rather than just an option to skip one value. Otherwise we could end up with a very large number of options to control each displayed fields. I think it would be interesting to introduce a `--format`/`--output` option, to control the columns printed. With the additional option to directly customize this in the `~/.jfremote.yaml` configuration file. What do you think about this solution? And would you be interested/available to try implementing it? In that case I can give some more tips about the code that needs to be modified. No pressure anyway in case you would rather not work more on this point.

In any case, for all the features that will be added I will also ask you to implement a test in https://github.com/Matgenix/jobflow-remote/blob/develop/tests/db/cli/test_job.py before merging.

FabiPi3:
I think you should upgrade the minimal typer version. I have troubles running the tests with typer 0.9.0, but newer versions does work. In the CI for python 3.12 typer version 0.12.something is used.

FabiPi3:
I have added an option to explicitly specify all columns that will be displayed. This can also be set in the config file. I had to rewrite the formatting table function to make it more dynamic. Let me know what you think.

I have also added a small test, would that be sufficient?

I was just wondering about to how to specify the columns in the terminal. Right now I use the list option from typer, which can be quite lengthy to type: `jf -fe job list -m 3 -hk db_id -hk name -hk state -hk worker -hk locked -hk last_updated -hk run_time`

Maybe is would be better to just pass a string: `jf job list -hk db_id&name&state&worker` and than split this string afterward. What do you think?

src/jobflow_remote/cli/formatting.py
```diff
@@ -26,79 +26,100 @@
...
-        table.add_column("Run time")
-        table.add_column("Retry time" + time_zone_str)
-        table.add_column("Prev state")
-        if verbosity < 2:
-            table.add_column("Locked")
-
-    if verbosity >= 2:
-        table.add_column("Lock id")
-        table.add_column("Lock time" + time_zone_str)
+    default_stored_data_value = Text.from_markup("[italic grey58]none[/]")
```
gpetretto: Why not leaving it just empty, as for the other fields that may not be filled?

src/jobflow_remote/cli/job.py
```diff
@@ -105,9 +106,25 @@ def jobs_list(
...
+            "--stored-data-key",
+            "-sdk",
+            help="Key to be shown from the stored_data field.",
+        ),
+    ] = None,
+    header_keys: Annotated[
+        Optional[list[str]],
+        typer.Option(
+            "--header-keys",
+            "-hk",
```
gpetretto: I would use `--output` and `-o` as options, I think it is more intuitive and matches other commands like `df`, `lsblk`. What do you think?

src/jobflow_remote/cli/job.py
```diff
@@ -105,9 +106,25 @@ def jobs_list(
...
+            help="Table columns to be shown. Overrides verbosity option. Can also be set in the config file",
+        ),
```
gpetretto, line 125: Maybe this should also contain the list of available options? Otherwise one would need to either look at the source code or proceed by trial and error. Would it be too long?

src/jobflow_remote/config/settings.py
```diff
@@ -30,6 +30,10 @@ class JobflowRemoteSettings(BaseSettings):
     cli_log_level: LogLevel = Field(
         LogLevel.WARN, description="The level set for logging in the CLI"
     )
+    cli_job_list_columns: list[str] = Field(
+        default_factory=list,
```
gpetretto: I think it would be less ambiguous to have this as `Optional[list[str]]` and have `None` as a default, instead of an empty list.

src/jobflow_remote/config/settings.py
```diff
@@ -30,6 +30,10 @@ class JobflowRemoteSettings(BaseSettings):
...
+        description="The list of columns to show in the `jf job list` command.",
```
gpetretto: Also here maybe it would be useful to have the full list?

src/jobflow_remote/cli/job.py
```diff
@@ -128,6 +145,11 @@ def jobs_list(
             worker_name,
         ],
     )
+    header_keys = header_keys or SETTINGS.cli_job_list_columns
```
gpetretto: There are two improvements that can be done here:
1) it would be good if the user is directly warned that passing header keys and verbose is incompatible. I think you can use the `check_incompatible_opt` function and just pass the two.
2) as things are now, if you set the defaults in the config file and then pass `-v`, at the moment it will be just ignored. If you want to see more columns you will be forced to pass it in CLI or modify the config file. So if there is a `-v` and `cli_job_list_columns` is set, the `-v` should have the priority.

gpetretto (commented):
Hi @FabiPi3,

thanks a lot for implementing the additional options! I think the implementation is fine. I left some minor comments on the code. 
I agree with you that repeating the option multiple times is really inconvenient for this kind of field. I think it would be good to switch to a single string, as you suggested. But I would use a `,` as a separator, rather than a `&`. It seems more standard and usually easier to type on the keyboard.

The test is fine, but can you also add one where it fails by passing an option that does not exist? You can pass `error=True` to `run_check_cli` and check that (part of) the error message is present in the output. 
Also, if you add the error for incompatible options and that -v overrides the configuration, it would be good to have tests for those use cases as well.

src/jobflow_remote/cli/formatting.py
FabiPi3: Ok

src/jobflow_remote/cli/job.py
FabiPi3: I am not familiar with those other commands but will stick to your suggestion here

src/jobflow_remote/cli/job.py
FabiPi3: Added it. I think its fine.

src/jobflow_remote/config/settings.py
FabiPi3: Ok.

src/jobflow_remote/config/settings.py
FabiPi3: Actually I would like to avoid hard-coding those values. But importing them will give a circular import error. So either I need to move the map definition or it could point to the help page of the list command. What do you think?

src/jobflow_remote/cli/job.py
FabiPi3: Done.

FabiPi3:
Hi Guido,

thanks for your feedback. I tried implementing all of it. I have also changed my code a bit.

I added a unit to the run time field as it felt necessary to me. What do you think?

Regarding the tests, I added two failing tests. I am not sure how to modify the SETTINGS object for the test to test this. Do you have a standard procedure?

src/jobflow_remote/config/settings.py
gpetretto: I think that referring to the help of `jf job list` for the list of available options should be enough.

gpetretto:
Thanks for the updates. 
For the test with changing the `SETTINGS` you can do it as in this test: https://github.com/Matgenix/jobflow-remote/blob/41bf0b287e7fd60615222b2ef38d01ebdd1bfc19/tests/db/cli/test_project.py#L15

Aside from this and adjusting the description of `cli_job_list_columns`, I think it will be ready to be merged.
I see that the tests are failing, but it seems unrelated to your changes and it is probably due to an inaccurate tests based on the dates of the job report.

src/jobflow_remote/config/settings.py
FabiPi3: Great.

FabiPi3:
Added your comments. I agree, the failing tests seemed unrelated to my changes.

gpetretto:
Thanks!

gpetretto (approved):

PR #232, opened 2024-12-26
Testing adding flags for codecov, separating unit and integration tests.

davidwaroquiers:


davidwaroquiers:
I still have a doubt on this PR. I am not entirely sure that any PR actually overwrites the "general" coverage reported in codecov... I will try to make a few tests about this.

davidwaroquiers:
Ok I've checked and it should be fine except that I think the coverage is uploaded for each python version. I'll add a condition to the workflow and only use the 3.12. I've also created a specific action to simplify the workflow itself.

gpetretto (approved):
All good to me

Issue #219, opened 2024-11-27
Integration tests are very slow and test bench build is not cached properly locally


ml-evs:
PR #160 comes with a fairly significant performance cost on the integration tests. These are not run by default in the local test suite (can be forced with env var CI=1), but this should probably be revisited soon. There is discussion in the comments of #160, but it seems the main slowdown is the fact that docker-py does not seem to be make good use of the existing cache, perhaps due to using some features not fully supported (docker-py cannot use buildkit). 

We can switch over to something that maps more closely to the Docker CLI, which should improve performance. More generally, the integration tests can be a bit slow as the containers are simulating real queues that have their own polling times. At the very least, we can parallelize the integration tests over Slurm/SGE, though this will be a bit fiddly with the current setup. I've added the pytest-durations plugin to track these slow steps.

PR #233, opened 2024-12-30
[WIP] test direct execution of docker commands and addition of PBS container

gpetretto:
Given the issues mentioned in #219 I am trying to replace the standard docker python package with [python-on-whales](https://gabrieldemarmiesse.github.io/python-on-whales/). This will test if there is any difference in timing when it comes to build the container in the CI. 
As for the caching, in my case this solves the issues of the container always being recreated and leaving dangling images.
It may be worth considering if the python-on-whales project is stable enough to fully rely on it.

I have also added the PBS container. it seems to work, but on my laptop building it is even slower than the SGE one. Keeping it disabled for the moment.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,7 @@ dependencies = [
 
 [project.optional-dependencies]
 dev = ["pre-commit>=3.0.0"]
-tests = ["docker ~= 7.0", "pytest ~= 8.0", "pytest-cov >= 4,< 6", "pytest-durations ~= 1.3", "pytest-mock ~= 3.14", "python-on-whales"]
+tests = ["docker ~= 7.0", "pytest ~= 8.0", "pytest-cov >= 4,< 6", "pytest-durations ~= 1.3", "pytest-mock ~= 3.14", "python-on-whales ~=0.75"]
 docs = [
     "autodoc_pydantic>=2.0.0",
     "pydata-sphinx-theme",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,7 @@ dependencies = [
 
 [project.optional-dependencies]
 dev = ["pre-commit>=3.0.0"]
-tests = ["docker ~= 7.0", "pytest ~= 8.0", "pytest-cov >= 4,< 6", "pytest-durations ~= 1.3", "pytest-mock ~= 3.14"]
+tests = ["docker ~= 7.0", "pytest ~= 8.0", "pytest-cov >= 4,< 6", "pytest-durations ~= 1.3", "pytest-mock ~= 3.14", "python-on-whales"]
 docs = [
     "autodoc_pydantic>=2.0.0",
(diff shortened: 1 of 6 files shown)


gpetretto:
For reference, testing with just slurm and SGE the integration tests took ~6 minutes to run here, while in other PRs with the standard docker implementation they required ~10 minutes. 
I will try enabling PBS tests as well.

gpetretto:
PBS tests are passing, but building the container seems to take ~5 additional minutes. I am not sure if it is really worth to have all the queueing systems tested in jobflow-remote, or if it would be better to use all of them only in qtoolkit. Using GHCR as mentioned in the other PR could be an option.

I should mention that I still have some issues running these containers on Mac. Slurm and SGE run fine as long as I run them in amd64 emulation with Rosetta. However for PBS the Jobs remain hanging in that case and the only way I managed to make it work is to run in amd64 emulation but using [Docker VMM](https://docs.docker.com/desktop/features/vmm/#docker-vmm-beta) to emulate the amd64 architecture. On the other hand Slurm and SGE container did not work when I tried to run them with Docker VMM.

pyproject.toml
```diff
@@ -28,7 +28,7 @@ dependencies = [
     "psutil >= 5.9,< 7.0",
     "pydantic ~= 2.4",
     "python-dateutil>=2.8.2",
-    "qtoolkit @ git+https://github.com/Matgenix/qtoolkit.git@develop",
```
ml-evs: I guess this is still needed for PBS?

pyproject.toml
```diff
@@ -39,7 +39,7 @@ dependencies = [
 
 [project.optional-dependencies]
 dev = ["pre-commit>=3.0.0"]
-tests = ["docker ~= 7.0", "pytest ~= 8.0", "pytest-cov >= 4,< 6", "pytest-durations ~= 1.3", "pytest-mock ~= 3.14"]
+tests = ["docker ~= 7.0", "pytest ~= 8.0", "pytest-cov >= 4,< 6", "pytest-durations ~= 1.3", "pytest-mock ~= 3.14", "python-on-whales"]
```
ml-evs:
```suggestion
tests = ["docker ~= 7.0", "pytest ~= 8.0", "pytest-cov >= 4,< 6", "pytest-durations ~= 1.3", "pytest-mock ~= 3.14", "python-on-whales ~= 0.74"]
```

This is the version that was installed in the last test run, though 0.75 is now out too

ml-evs (approved):
Looks great @gpetretto! I only have a couple of minor comments but these shouldn't hold up merging.

pyproject.toml
gpetretto: indeed it was needed for PBS. switched to qtk 0.1.16

pyproject.toml
gpetretto: set to `~=0.75`

Issue #220, opened 2024-11-27
Better error handling for `pre_run`


ml-evs:
In #160 I repeatedly ran into a hard-to-debug issue where SGE was defaulting to the wrong shell, and thus `pre_run` was failing. There was not any good user feedback for this issue, instead jobs would fail without any outputs. This should be checked explicitly in the code with a better error message.

gpetretto:
Thanks for reporting this issue. The problem here is that the `pre_run` is executed inside the script. At the moment jobflow-remote determines that the job did not complete successfully. I am not sure how it could determine if the error happened during the pre_run or during the execution of the Job. I suppose in your case you there were errors in both parts: an error while trying to activate the environement and then an error for the unknown `jf` command. So it would still be hard to tell the cause. 
In a normal execution, the job would have gone in the `REMOTE_ERROR` state and the user can probably figure out relatively easily the origin of the problem by inspecting the `queue.out` and `queue.err` files (for which there is also the `jf job queue-out` command). Maybe there should be something specific for the tests, where additional information is extracted from the container, before it is siwtched off?

ml-evs:
> Thanks for reporting this issue. The problem here is that the `pre_run` is executed inside the script. At the moment jobflow-remote determines that the job did not complete successfully. I am not sure how it could determine if the error happened during the pre_run or during the execution of the Job. I suppose in your case you there were errors in both parts: an error while trying to activate the environement and then an error for the unknown `jf` command. So it would still be hard to tell the cause. In a normal execution, the job would have gone in the `REMOTE_ERROR` state and the user can probably figure out relatively easily the origin of the problem by inspecting the `queue.out` and `queue.err` files (for which there is also the `jf job queue-out` command). Maybe there should be something specific for the tests, where additional information is extracted from the container, before it is siwtched off?

Yeah, I was imagining some kind of preflight check, but that won't work well with different compute/login node environments. Perhaps just automatically doing the equivalent of `jf job queue-out` when storing the the error in the local database?

Issue #211, opened 2024-11-19
An Error occurred during the command execution: ValueError No daemon runner document.


chiang-yuan:
Thanks for developing such a great tool!

When I tried to run `jf runner start`, I got the following error:

```
An Error occurred during the command execution: ValueError No daemon runner document.
The auxiliary collection does not contain information about running daemon. Your database was likely set up or reset using an old version of Jobflow Remote. You can upgrade the database using the command "jf admin upgrade".
```

Running `jf admin upgrade` gives another error that has little message to debug:

```
An Error occurred during the command execution: TypeError expected string or bytes-like object, got 'NoneType'
``` 

`jf project check --errors` has passed without error. Not sure how to to debug this. Perhaps it is related to batch submission #172 
 ?

```
┏━━━━━━━┳━━━━━━━━━━━━━━━┳━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━┓
┃ DB id ┃ Name          ┃ State ┃ Job id  (Index)                           ┃ Worker       ┃ Last updated [PST] ┃
┡━━━━━━━╇━━━━━━━━━━━━━━━╇━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━┩
│ 1     │ MP GGA static │ READY │ aa0dfc19-6cdd-4296-9454-5c54319d0c9f  (1) │ batch_worker │ 2024-11-18 18:58   │
└───────┴───────────────┴───────┴───────────────────────────────────────────┴──────────────┴────────────────────┘
```

gpetretto:
Hi @chiang-yuan,

thanks for following up and testing the batch functionality of jobflow-remote. 
[...]

Issue #210, opened 2024-11-15
Optionally delay download after job end


gpetretto:
From a discussion in qtoolkit https://github.com/Matgenix/qtoolkit/pull/43 it emerged that it may be convenient to delay the dowload step after job completion in case of slow NFS on the worker. 
This should be possible by adding a `delay_download` option for the worker and in that case set the `retry_time_limit` when a job goes in the `TERMINATED` state.

ml-evs:
If I can't see anything obvious causing the issues at #160, I'll try to hack this feature in as a first draft -- though I'd be surprised if that is the cause of my current woes.

In general this feature would be useful given the peculiarities of parallel file systems. It's probably better to err on the side of waiting for a very long time for downloading files that we expect to appear even if the job errored, i.e., the jfremote outputs.

gpetretto:
Implemeted in #234

PR #234, opened 2025-01-09
Various updates

gpetretto:
Several updates addressing previously discussed issues.
* Allow to specify a delay between the end of a job on the worker and the start of a download. Closing #210
* If the `remote_job_data.json` file is missing from the run folder this usually generates an unclear error message (as discussed in #157. If this is missing, allow to complete the download step and raise an error during the completion. This should allow to often provide better error messages if they are included in the jfremote_out.json.
* To better show the error queue.out and queue.err are also downloaded and stored in the DB. Cut if they are too long and only showed in `jf job info` with the `-v` option. This partially addresses #220. More updates for the CI in an upcoming PR.
* Improve the upgrade procedure to handle development version. Also making the option to set the target version visibile. Closing #211
* Option `--locked` for the list of flows
* Actually handle the newly introduced `Flow.metadata`. Closing #124
* fix the handling of weeks interval in reports. There were inconsitent results depending on the day of the week and year the report was generated. Also unexpected causing test failures. 
* Don't check the MAC address when checking the running runner in the DB. The code could give different values for the same machine (if multiple network devices are present), or different result if a virtual machine with the same hostname is switched on and off.


src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -3074,18 +3099,49 @@ def get_trends(
...
+                result_dict[r["_id"]] = {
+                    state_cls(s.upper()): n for s, n in r.items() if s != "_id"
+                }
+
+            for i in range(1, num_intervals + 1):
+                expected_date_str = get_past_time_rounded(
+                    interval=interval, num_intervals=i, reference=tznow
+                ).strftime(date_format)
+                if expected_date_str not in result_dict:
+                    result_dict[expected_date_str] = {state: 0 for state in states}
```
davidwaroquiers, line 3144: Maybe this could go into a separate utils package somewhere ? Not sure it needs to be done now, if at all but the jobcontroller.py file is getting very big. Maybe also it's not worth for this part of the code as it is only used here. I just mentioned here not to forget :)

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -4046,6 +4154,57 @@ def lock_auxiliary(self, **lock_kwargs) -> Generator[MongoLock, None, None]:
...
+            run_dir=job_doc["run_dir"],
+        )
+        if not local_path_str:
+            return None, None
+        local_path = Path(local_path_str)
+        queue_out_path = local_path / "queue.out"
+        queue_err_path = local_path / "queue.err"
+        queue_out = None
+        queue_err = None
+        length_limit = 3000
```
davidwaroquiers, line 4174: Capitalized as it is a constant ? Should it be tunable somehow ? Let's keep as it is for now anyway.

davidwaroquiers (approved):
Great work. All good to me!

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: Indeed we might need to discuss about how to handle the JobController code. Let's postpone the discussion for another PR.

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: Let's see if someone asks for changing this into a customizable option.

PR #237, opened 2025-01-10
[WIP] Save db dumps for test failures

gpetretto:
To ease the debugging of failing tests, with this PR a dump of the DB is done for each test that fails. In case the tests are running in CI the dump is uploaded to github as an artifact for inspection. This, with the content of queue.out and queue.err stored in the DB (#234), should hopefully solve #220.

A test contains a fake failure to check if the upload of the artifact is working correctly.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -155,3 +155,87 @@ def upgrade_test_dir(test_dir):
     """
 
     return test_dir / "upgrade"
+
+
+@pytest.hookimpl(tryfirst=True, hookwrapper=True)
+def pytest_runtest_makereport(item, call):
+    """
+    Hook to set the report of the test. Needed to check if a test failed.
+    """
+    outcome = yield
+    report = outcome.get_result()
+    if report.when == "call":
+        item.rep_call = report
+
+
+@pytest.fixture(scope="session")
+def shared_test_out_dir(tmp_path_factory):
+    """
+    Fixture to lazily create a shared temporary directory to store the dump
+    of MongoDB for failed tests and move them to store as an artifact is
+    running on github.
+    """
+    import os
+
+    _tmp_dir = None
+
+    def get_or_create():
+        nonlocal _tmp_dir
+        if _tmp_dir is None:
+            _tmp_dir = Path(tmp_path_factory.mktemp("shared_test_dir"))
+        return _tmp_dir
+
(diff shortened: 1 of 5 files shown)


gpetretto:
The procedure seems to work properly. Removed the fake test failure.

tests/conftest.py
```diff
@@ -155,3 +155,87 @@ def upgrade_test_dir(test_dir):
...
+
+    yield get_or_create
+
+    # If on github and the folder was created (i.e., some test failed), move
+    # it so that it can be uploaded as an artifact.
+    if os.getenv("GITHUB_WORKSPACE") and _tmp_dir:
+        artifact_path = Path(os.getenv("GITHUB_WORKSPACE")) / "test_folder"
+        os.rename(_tmp_dir, artifact_path)
+
+        print(f"Database dump saved as artifact: {artifact_path}")
```
davidwaroquiers, line 196: This will only be printed if pytest is run with -s no ? And as it is on github, we can't really change that "easily". Or maybe I missed some config somewhere that actually makes this print shown.

tests/conftest.py
```diff
@@ -155,3 +155,87 @@ def upgrade_test_dir(test_dir):
...
+        test_dump_dir = target_dir / sanitized_test_name
+        test_dump_dir.mkdir(parents=True, exist_ok=True)
+
+        # don't use the backup to create indented json files for easier access
+
+        for coll in [jc.jobs, jc.flows, jc.auxiliary]:
+            dump_file_path = test_dump_dir / f"{coll.name}.json"
+            dumpfn(list(coll.find()), dump_file_path, indent=2)
+
+        print(f"Data dumped to {test_dump_dir}")
```
davidwaroquiers, line 228: Same as above

davidwaroquiers (approved):
Sounds good to me except for the printing of the path. But maybe I am misled (and you actually did some tests so I guess it actually worked ? Not sure why the print is shown in that case though). Feel free to merge if my comments on these prints were wrong.

tests/conftest.py
gpetretto: Actually locally it is printed when a test fails and I think I checked that this is printed also on github.

davidwaroquiers: Ok

PR #241, opened 2025-01-13
Web GUI for jobflow remote

fraricci:
I'm proposing here a simple GUI for jobflow remote.
The GUI is a web application implemented with the recent 
[fasthtml](https://fastht.ml/) python library by Jeremy Howard.

At the moment, it is a preliminary draft, with very simple looking, but it is usable and can do the following:
- Handle more projects
- Show a general reports about the jobs and flows status
- Query jobs and flows according different criteria and show them in a table with pagination
- Some actions (Rerun, Play, Pause, Stop, Retry for jobs; Delete for flows) can be performed on selected jobs/flows
- Details of each job/flow are available through a popup dialog
- Show the mermaid graph for flows in the same dialog
- Show the current status of the Runner and can Start/Stop it

Hope you like it!
Please, if you're interested, give it a try and provide any feedback.
Hopefully, we can start from here to build something more complete, robust, and better looking.


gpetretto (approved):
Great work. Thanks a lot!

PR #246, opened 2025-01-17
Test for save db artifact

davidwaroquiers:


davidwaroquiers:
This should be ready to review @gpetretto

PR #247, opened 2025-01-20
Docker optimisations for CI

ml-evs:
Building on #233 by using the ghcr.io for caching. Will likely need permissions updates to get working properly, plus a few rounds of debugging in the CI

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -90,7 +90,7 @@ jobs:
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
-      
+
       - name: List images
         run: docker images
 
@@ -99,6 +99,7 @@ jobs:
         with:
           load: true
           files: tests/integration/dockerfiles/docker-bake.hcl
+          set: "*.cache-from=type=gha,scope=global"
 
       - name: List images
         run: docker images
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -47,10 +47,14 @@ jobs:
         uses: docker/bake-action@v6
         with:
           load: true
+          push: false
           files: tests/integration/dockerfiles/docker-bake.hcl
           set: |
             *.cache-to=type=gha,scope=global,mode=max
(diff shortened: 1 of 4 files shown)


ml-evs:
Hi @gpetretto, this PR seems to be working ok and cuts out a chunk of a build time. I pushed my local build and cache to the registry, so they should be usable by everyone locally too, and the CI now makes much better use of the GHA cache. I'm still not entirely sure whether its actually caching between the two different build jobs, or if they now just have their own cache, so we might also be able to remove the build-test-bench job at some point. I'll wait until the other PR is merged to tidy this one up!

gpetretto:
Hi @ml-evs, this looks great! Thanks!

I have tested this locally and it works nicely, except for the issue that I have mentioned in the other PR https://github.com/Matgenix/jobflow-remote/pull/233#issuecomment-2571663002: Slurm and SGE work only if I emulate am64 with Rosetta and PBS works only if I emulate with Docker VMM. Anyway, it seems this is just an issue with my setup. If I can't find a better solution, I will probably introduce an option for pytest to set the `WORKERS` variable in the integration tests to make my life easier.

ml-evs:
Hi @gpetretto / @davidwaroquiers, I've just rebased this and think it should provide a simple speed up to future CI runs, please merge whenever you are happy!

gpetretto:
Hi @ml-evs, thanks a lot for this!

Just one question: what would be the procedure if there would be a need to upgrade the images for any reason? Do you think it would be worth to document it somewhere?

ml-evs:
> Hi @ml-evs, thanks a lot for this!
> 
> Just one question: what would be the procedure if there would be a need to upgrade the images for any reason? Do you think it would be worth to document it somewhere?

You could just build locally and push, or even just run the workflow once with push: true to refresh the cache, but probably most changes won't even affect most layers so the speedup will work fine. We could also just add a release workflow that rebuilds the images from scratch.

gpetretto:
Thanks for the clarification. I think we can start to merge this and see later if a workflow is needed.

gpetretto (approved):

PR #253, opened 2025-02-05
Bug fix for Remote Host

gpetretto:
Two issues in  `RemoteHost`:
* `connect` and `close` could have been trapped in an infinite loop if a gateway was present, but not an instance of `Connection`
* The `_execute_remote_func` approach replaced the `Connection` with a new one, but did not use it when retrying the command immediately. This is because it was always called so that `remote_cmd` contained a referece to `self.connection`, that then was still pointing to the old `Connnection`.

### DIFF
--- a/src/jobflow_remote/remote/host/remote.py
+++ b/src/jobflow_remote/remote/host/remote.py
@@ -307,6 +307,11 @@ class RemoteHost(BaseHost):
             Keyword arguments to pass to the remote command
         """
 
+        # The function needs to be called in this way because if retry_on_closed_connection
+        # is True and the first execution of the command fails a new instance self._connection
+        # will be recreated. Since remote_cmd are typically methods of the Connection object,
+        # this ensures that the method is called on the actual instance of the connection, and
+        # not on a previous one.
         def execute_cmd():
             # create the function associated to the current connection
             with ExitStack() as stack:
--- a/src/jobflow_remote/remote/host/remote.py
+++ b/src/jobflow_remote/remote/host/remote.py
@@ -320,7 +320,7 @@ class RemoteHost(BaseHost):
                 return execute_cmd()
             except OSError as e:
                 msg = getattr(e, "message", str(e))
-                error: Exception = e
+                error: BaseException = e
                 if "Socket is closed" not in msg:
                     raise
             except SSHException as e:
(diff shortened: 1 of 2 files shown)


davidwaroquiers (approved):
Good to me, thx!

src/jobflow_remote/remote/host/remote.py
```diff
@@ -265,24 +271,56 @@ def is_connected(self) -> bool:
...
     def copy(self, src, dst) -> None:
         cmd = ["cp", str(src), str(dst)]
         self.execute(cmd)
 
-    def _execute_remote_func(self, remote_cmd, *args, **kwargs):
+    def _execute_remote_func(
+        self,
+        remote_cmd: Callable,
+        *args,
+        _connection_workdir: str | Path | None = None,
```
davidwaroquiers, line 289: Just wondering here why it's with an underscore in front ? it usually means something private right ? I think I had never seen it in an argument of a function or method.

src/jobflow_remote/remote/host/remote.py
```diff
@@ -265,24 +271,56 @@ def is_connected(self) -> bool:
...
+            Keyword arguments to pass to the remote command
+        """
+
+        def execute_cmd():
+            # create the function associated to the current connection
+            with ExitStack() as stack:
+                if _connection_workdir:
+                    stack.enter_context(self.connection.cd(_connection_workdir))
+                func = remote_cmd(self)
+                return func(*args, **kwargs)
```
davidwaroquiers, line 321: A little bit of why it is done this way and maybe why the initial way of calling this method (with self.connection.run for example instead of a lambda function that is then used here to get the command from self) would help :)

davidwaroquiers (approved):
Ok I think it's fine. Just maybe add a note/comment/explanation as to why it is using a lambda (maybe even mentioning why the other option was not working?).

src/jobflow_remote/remote/host/remote.py
gpetretto: To be fair I did not see any such usage either. The point is that args and kwargs are passed to the function called, so there should be no conflict in the potential arguments of the function that should be called and this "workdir" argument I have added. The possibility of a clash is basically null, but I though that using the underscore this would make a clash impossible. However, I don't think this is a standard. I can remove the underscore.

davidwaroquiers: I see. I think its fine to keep it like this. Up to you.

gpetretto: let's leave it like this then

src/jobflow_remote/remote/host/remote.py
gpetretto: added some explanation

PR #255, opened 2025-02-06
Include coverage from code executed remotely

davidwaroquiers:
In integration tests, part of the jobflow-remote code is executed in docker containers (jf execution command line within slurm/sge/pbs submission scripts). This PR activates coverage in the docker containers and copies back the coverage files locally. These are then combined with the local coverage file to give a proper coverage.


tests/integration/conftest.py
```diff
@@ -194,6 +194,22 @@ def compose_containers(
...
+                        continue
+                    flist = c.execute(
+                        ["ls", "-a", "/home/jobflow/coverage/"]
+                    ).splitlines()
+                    for file in flist:
+                        if file.startswith(".coverage"):
+                            c.copy_from(
+                                f"/home/jobflow/coverage/{file}",
+                                f"{coverage_dir}/{file}",
+                            )
```
davidwaroquiers, line 243: Even though it does not guarantee that everything works, maybe I should add a check that at least one file was copied ?

tests/integration/dockerfiles/Dockerfile
davidwaroquiers, line 1: Wondering here about one thing. We have a matrix of python versions from 3.9 to 3.12, but in the docker containers, it is always the 3.10. (at some point I thought my code for the coverage from the containers would only work for 3.10 because the coverage.path.source in pyproject.toml is hard coded to 3.10 but it works as it's always 3.10)
Should we do something here ?

ml-evs: I thought it would potentially be overkill to make the containers with different Python versions too (as probably it should also be tested that the Python 3.11 daemon can interact with Python 3.10 runner etc). We could add it though, probably the most sensible thing would be to install a virtual environment for each Python version inside one "fat" container, though I don't know how this would break your hardcoded coverage path at the moment (perhaps just needs a list of all the Python virtualenv locations).

pyproject.toml
```diff
@@ -169,7 +170,10 @@ parallel = true
 branch = true
 
 [tool.coverage.paths]
-source = ["src/"]
+source = [
+    "src/",
+    "/home/jobflow/.venv/lib/python3.10/site-packages/",  # for coverage from docker containers
```
ml-evs: Does this break anything locally if this folder doesn't exist? Perhaps we could only enable coverage when testing in the CI

ml-evs (approved):
Looks nice, thanks @davidwaroquiers! 

My only general comment is that perhaps the "remote" coverage could be flagged as such -- i.e., take the coverage file from docker as a separate xml and upload it to codecov with a separate flag before combining it with the rest. Might help some future debugging, though can definitely be done in a different PR! Otherwise happy to approve pending the conflict fix and discussion of Python versions.

tests/integration/dockerfiles/Dockerfile
davidwaroquiers: I agree that one container per python version is overkill. As you say, maybe we can have one environment per python version in each container. In any case, I think we can put that to another PR for now.

davidwaroquiers: Closing and opened #256

pyproject.toml
davidwaroquiers: No it does not. What this does is: when it opens a .coverage file to combine it with others, if the path starts with any of the source, it replaces it with the first one (here, src/). Locally, covered files are e.g. src/jobflow_remote/core/... In the containers, covered files are e.g. /home/jobflow/.venv/lib/python3.10/site-packages/jobflow_remote/core/... If source was not defined, then the coverage would mention twice each file.
I think we can keep coverage enabled at all times.

davidwaroquiers:
> Looks nice, thanks @davidwaroquiers!
> 
> My only general comment is that perhaps the "remote" coverage could be flagged as such -- i.e., take the coverage file from docker as a separate xml and upload it to codecov with a separate flag before combining it with the rest. Might help some future debugging, though can definitely be done in a different PR! Otherwise happy to approve pending the conflict fix and discussion of Python versions.

Thanks for your review!

This can be done and we can discuss it. My only "concern" is that in any case, this will only ever cover a "small" part of the code as the "remote" coverage is only about the "jf execution" command (hidden command from the user btw, it doesn't appear in the help as it is only meant to be used "internally" within the submitted script). Where would you see this to help debugging ?

tests/integration/conftest.py
davidwaroquiers: I will not do that for now as I am not completely sure what do in case pytest is explicitly executed on a specific test file in which containers are not used. In that case it wouldn't copy a coverage file obviously ...

davidwaroquiers:
This is ready for review/test @gpetretto (@ml-evs, if you want to have another look ?).

- Fixed problem that made integration tests in conflict with other tests
  - Can now do "CI=1 pytest"
  - Was due to a missing SETTINGS import
- Fixed problem that prevented from running tests with "-s"
- All tests are now marked as either unit, db or integration (automatically, based on the directory)
  - You can do e.g. "pytest -m unit" or "pytest -m integration" => this is now used in the testing.yml github workflow
  - Tests outside the unit/db/integration dirs HAVE to be marked as either unit, db or integration (this could be modified if needed ?)
- Coverage from remote execution (in the dockers for integration tests) is now integrated
- Added new flags. All flags are: unit, db, integration_remote, integration_local, integration, all_local, all
- Added --coverage-per-flag option to pytest to get the coverage per flag when running tests locally
- Flags (and the corresponding coverage files to be combined) are defined in a yaml file that is used both by the coverage-per-flag option, and by the github CI
- Added two pytest fixtures to
  - Copy the run files from the containers at the end of the integration tests (--copy-files-from-containers)
  - Keep the containers alive at the end of the integration tests (--keep-containers-alive)
- Fixed the clean_slurm_queue fixture to first wait until batch slurm jobs are finished before cancelling them. This is needed to make sure coverage is properly accounted for.

Once you tell me it's fine, I will merge. I have also updated the documentation.

ml-evs:
Looks good, if a little more complicated than we initially thought :upside_down_face: 

From the codecov side, the logs seem to indicate that you might need to push an extra empty commit to get their diffs working correctly? e.g., from https://app.codecov.io/github/Matgenix/jobflow-remote/commit/ed9b70375b580b33e008b79e9a03734a724cf326 I see the following:

2025-08-19-103050_384x431_scrot

2025-08-19-103319_1005x122_scrot

tests/conftest.py
```diff
@@ -259,6 +309,112 @@ def job_controller_drop(random_project_name, job_controller):
...
+    if session.config.getoption("coverage_per_flag"):
+        modified_args = [
+            arg for arg in sys.argv[1:] if arg not in ("--coverage-per-flag",)
+        ]
+        if "-m" not in modified_args:
+            index_last_minus_m = None
+        else:
+            index_last_minus_m = next(
+                i for i, v in reversed(list(enumerate(modified_args))) if v == "-m"
+            )
```
gpetretto, line 397: wouldn't it be possible to add this option in `pytest_addoption` instead of manually parsing it like this?
Or does this serve some other purpose?

tests/db/cli/test_admin.py
```diff
@@ -21,11 +21,12 @@ def test_reset(job_controller, one_job) -> None:
         f = Flow(add(1, 2))
         submit_flow(f, worker="test_local_worker")
 
-    run_check_cli(
-        ["admin", "reset"],
-        required_out="The database was NOT reset",
-        cli_input="y",
-    )
+    with capsys.disabled():
```
gpetretto: why this capsys.disabled? and why only here?

tests/conftest.py
```diff
@@ -259,6 +309,112 @@ def job_controller_drop(random_project_name, job_controller):
...
+            session.config.option.cov = "jobflow_remote"
+            os.environ["COVERAGE_FILE"] = f".coverage-{marker}"
+            this_marker_args.extend(
+                ["--cov=jobflow_remote", "--cov-config=pyproject.toml"]
+            )
+            pytest.main(this_marker_args)
+
+        import yaml
+
+        module_dir = Path(__file__).resolve().parent
```
gpetretto, line 423: is it really necessary to do here all this logic? Cannot be done as an additional action to merge the files?
If not, better adding some comments about why that's the case.

tests/integration/conftest.py
```diff
@@ -442,6 +518,12 @@ def clean_slurm_queue(write_tmp_settings):
     project = write_tmp_settings
     worker = project.workers["test_remote_slurm_worker"]
     queue_manager = QueueManager(worker.get_scheduler_io(), worker.get_host())
+    # If tests are run with coverage, first try to wait until the slurm job finishes smoothly
+    if coverage_file:
+        for _ in range(60):
```
gpetretto: wouldn't a shorter time be enough here? Tests are already quite slow, without risking to wait for minutes to wait for queues to be cleaned.

tests/db/conftest.py
```diff
@@ -173,7 +169,9 @@ def write_tmp_settings(
     jobflow_remote.SETTINGS = JobflowRemoteSettings()
 
     yield
-    shutil.rmtree(tmp_dir)
+
+    if tmp_proj_dir.exists():
+        shutil.rmtree(tmp_proj_dir)
```
gpetretto: Since the directory is now created in the tmp_proj_work_dirs fixture, wouldn't it make more sense to delete it there as well?

tests/integration/conftest.py
```diff
@@ -193,27 +201,90 @@ def compose_containers(
...
+                    ).splitlines()
+                    for file in flist:
+                        if file.startswith(".coverage"):
+                            c.copy_from(
+                                f"/home/jobflow/coverage/{file}",
+                                coverage_container_dir / file,
+                            )
+                    with cd(coverage_container_dir):
+                        # Combining coverage from the different jf execution runs for all the tests on this container
+                        cov = Coverage()
```
gpetretto, line 246: What about importing the `Coverage` in this part of the code? I know that probably it will be installed if someone is running the integration tests or if it was installed with "pip install .[tests]", but maybe often one does not do that...

tests/integration/conftest.py
```diff
@@ -416,23 +487,28 @@ def write_tmp_settings(
...
+    with open(tmp_proj_dir / f"{random_project_name}.json", "w") as f:
         f.write(project_json)
 
+    # In some cases it seems that the SETTINGS have already been imported
+    # and thus not taking the new configurations into account.
+    # Regenerate the JobflowRemoteSettings after setting paths and project
+    import jobflow_remote
+    from jobflow_remote.config.settings import JobflowRemoteSettings
+
+    jobflow_remote.SETTINGS = JobflowRemoteSettings()
```
gpetretto, line 525: I wonder if it wouldn't be safer to use the monkeypatch here? Maybe whatever set the SETTINGS earlier may want what was set before?

tests/integration/conftest.py
```diff
@@ -416,23 +487,28 @@ def write_tmp_settings(
...
-        os.environ["JFREMOTE_PROJECTS_FOLDER"] = original_jf_remote_projects_folder
-    if original_jf_remote_project is not None:
-        os.environ["JFREMOTE_PROJECT"] = original_jf_remote_project
-    if original_config_file is not None:
-        os.environ["JFREMOTE_CONFIG_FILE"] = original_config_file
+    if pytestconfig.getoption("keep_containers_alive"):
+        print("\n * Containers are kept alive ... also keeping project configuration:")
+        print(f"\n  - Directory for project configuration file: {tmp_proj_dir}")
+    elif tmp_proj_dir.exists():
+        shutil.rmtree(tmp_proj_dir)
```
gpetretto: here as well, wouldn't it make sense to delete this in the `tmp_proj_work_dirs` fixture?

tests/integration/dockerfiles/Dockerfile
```diff
@@ -31,7 +32,11 @@ RUN sed -i '/\[tool.versioningit.vcs\]/,+3d' /home/${USERNAME}/jobflow-remote/py
 WORKDIR /home/${USERNAME}/jobflow-remote
 RUN uv venv /home/${USERNAME}/.venv && \
     uv pip install --python /home/${USERNAME}/.venv/bin/python -r /home/${USERNAME}/jobflow-remote/requirements/requirements.txt && \
-    uv pip install --python /home/${USERNAME}/.venv/bin/python --no-deps .
+    uv pip install --python /home/${USERNAME}/.venv/bin/python --no-deps . && \
+    uv pip install --python /home/${USERNAME}/.venv/bin/python -r /home/${USERNAME}/jobflow-remote/requirements/requirements_tests.txt
+
+# Create a .pth file to auto-start coverage
+RUN printf "import coverage; coverage.process_startup()\n" > /home/${USERNAME}/.venv/lib/python3.10/site-packages/coverage_autostart.pth
```
gpetretto: This looks a bit shady. Isn't there any cleaner way of starting the coverage? Are pth files used for these kinds of things? How does it start normally without manually forcing it?
If strictly necessary, this is also hardcoding a path that depends on the structure in the container. Isn't there a way to avoid hardcoding the path?

gpetretto (commented):
Thanks! I left some points to be discussed

tests/integration/dockerfiles/Dockerfile
davidwaroquiers: It is the way provided by coverage: https://coverage.readthedocs.io/en/latest/subprocess.html#manual-sub-process-coordination

I think one possibility could be to run coverage by changing the jf command to something like "coverage run jf" or maybe "coverage run -m jf"

tests/conftest.py
davidwaroquiers: you mean adding --unit-tests, --db-tests etc ? The goal here is that if you pass both a marker (or multiple ones) with -m and --coverage-per-flag it is properly accounted for.

tests/conftest.py
davidwaroquiers: Here, all this part is only used if you run with the --coverage-per-flag locally. When running on github, it is not using this coverage-per-flag.

tests/db/cli/test_admin.py
davidwaroquiers: I'm not completely sure why exactly, but I had to add it in a few places so that pytest with "-s" would work. Why here and not in other places, honestly I don't remember.

tests/db/conftest.py
davidwaroquiers: good point

tests/integration/conftest.py
davidwaroquiers: Ok can do that

tests/integration/conftest.py
davidwaroquiers: good point

tests/integration/conftest.py
davidwaroquiers: I can try indeed, I guess 30 seconds should be fine (maybe even less). I'll make a few tests

tests/conftest.py
davidwaroquiers: I will add some comments about what is done here

tests/conftest.py
davidwaroquiers: Also add some comment here

tests/db/cli/test_admin.py
davidwaroquiers: I can try to see again.

tests/integration/conftest.py
```diff
@@ -193,27 +201,90 @@ def compose_containers(
...
+
+                    shutil.move(
+                        ".coverage", coverage_dir / ".coverage-integration-remote"
+                    )
         finally:
-            try:
-                print("\n * Stopping containers...")
+            if pytestconfig.getoption("keep_containers_alive"):
+                print("\n * Keeping containers alive...")
+                print(f"\n  - Docker compose yaml file: {f.name}")
```
davidwaroquiers, line 267: Add how to connect

tests/integration/conftest.py
davidwaroquiers: I will check if is still needed

tests/integration/dockerfiles/Dockerfile
davidwaroquiers: I will see if I can have the site-packages directory dynamically

tests/conftest.py
davidwaroquiers: Added comment at the beginning of the method.

tests/conftest.py
davidwaroquiers: Added comment.

tests/db/cli/test_admin.py
davidwaroquiers: I don't know why exactly but it seems to work without this capsys.disabled now ... I removed them all

tests/db/conftest.py
davidwaroquiers: Moved that part to tmp_proj_work_dirs (actually I moved the other part, that included the check on the "keep containers alive" there)

tests/integration/conftest.py
davidwaroquiers: Moved to tmp_proj_work_dirs

tests/integration/conftest.py
davidwaroquiers: Moved. (also moved another coverage import that was in tests/conftest.py as well)

tests/integration/conftest.py
davidwaroquiers: Done

tests/integration/dockerfiles/Dockerfile
davidwaroquiers: Done. note that there are a few other things where we explicitly install 3.10. We can try to deal with that in another PR. At least here whichever environment is used, this coverage part should use the right environment

tests/integration/conftest.py
davidwaroquiers: Ok just removing that made all the integration tests fail. The exact same thing is also done in db/conftest.py. I guess it is needed, not exactly sure why and if we can remove it. I'd propose to keep it that way for now ?

tests/integration/conftest.py
davidwaroquiers: I reduced the delay batch and the max_wait and now it seems to work with 30 seconds.

davidwaroquiers (commented):
Should be ready for re-review @gpetretto

tests/integration/dockerfiles/Dockerfile
gpetretto: thanks. Indeed, we can try to address the other points as well in another PR

tests/integration/conftest.py
gpetretto: ok, I did not remember. let's leave it like this

PR #260, opened 2025-03-02
Reset stored data field on rerun

FabiPi3:
I've added a line of code which should ensure that the stored_data field is also reset when calling `jf job rerun`. To be honest I've not tested it so far.

I am not sure how this field is intended to be used, but in the way I am it makes sense to reset it also on rerun.

What is your opinion?

### DIFF
--- a/src/jobflow_remote/jobs/data.py
+++ b/src/jobflow_remote/jobs/data.py
@@ -517,4 +517,5 @@ def get_reset_job_base_dict() -> dict:
         "updated_on": datetime.utcnow(),
         "start_time": None,
         "end_time": None,
+        "stored_data": None,
     }


gpetretto:
Hi @FabiPi3, indeed I think this is reasonable. 
If I remember correctly, you are using it to get more information on the final results of your job. Is that correct? I believe that this should be the general use case and if the job is going to be executed again the previous `stored_data` should not apply anymore.

FabiPi3:
Hi @gpetretto, yes I use it to get detailed information on how the job ended. In any case this data is specific to the run and will not accumulate in case of rerun but will be regenerated.

gpetretto (approved):

gpetretto:
Thanks!

Issue #155, opened 2024-07-25
`Runner` doesn't pick up config and code changes until restarted


janosh:
when i modify `some-project.yaml` config file, the changes are not picked up by current or new calculations that start after those changes were made until the `Runner` is killed and restarted. the same is true for direct changes to Python code (e.g. in `jf-remote`, `qtoolkit`, `jobflow`, `atomate2`). while i can see this being a feature in that it ensures homogeneous settings until you deliberately restart the `Runner`, it wasn't clear to me when i started using `jf-remote` and took at least an hour to figure out. ideally, this should be clearly documented somewhere (apologies if it is already and i missed it). another option to consider would be to make this configurable in `~/.jfremote.yaml`

```yaml
runner_mode: static | reload-config | reload-code | reload
```

`static` mode would be the current behavior while `reload-config` would reload the `some-project.yaml` once a minute or so. `reload-code` i'm less sure is worth implementing as it's more of an edge case and might be a can of worms to get working reliably. in case `reload-config` also goes against the `jobflow-remote` design philosophy, feel free to consider this issue as only a suggestion for more documention.

davidwaroquiers:
Thanks @janosh for this issue. Indeed, this may not be described in the documentation. I understand it could indeed be a use case and that the runner could regularly check if the config has been updated. We already though about this and discussed with @gpetretto but I remember there were a few tricky things. What are the things in the project config you'd like to update dynamically ? I guess it's the exec_config ? or is it something else ? maybe the workers ? Thing is what if someone modifies the stores (maybe by mistake). That would be quite problematic and the database could end up in an inconsistent state. If this dynamical reload is done, there should be some checks on what can be reloaded from the project and maybe the runner should stop itself if something "not allowed" has changed

In any case, we should at least document this more clearly

janosh:
> What are the things in the project config you'd like to update dynamically ? I guess it's the exec_config ? or is it something else ? maybe the workers?

in my case both the `workers` (as i often have to iterate on `pre_run` and `post_run` for things like `module load` before things start running smoothly on a new cluster)

```yml
    pre_run: &pre_run |
      module purge
      module load slurm/stage openmpi/stage vasp/stage
      export ATOMATE2_VASP_CMD="srun vasp_std"
      echo "slurm job with id $SLURM_JOB_ID started at $(date +"%Y-%m-%d %H:%M:%S %Z") on $(hostname) running '$ATOMATE2_VASP_CMD'"
[...]

PR #269, opened 2025-03-20
CLI and other small updates

### LINKED ISSUE #263: How to solve the question: TypeError: 'dict' object is not callable
Dear Developers，

Recently, I tried to submit a task locally to a remote server, but the task failed. However, queue stderr file has no error message, so I suspect there might be an issue with the format of my configuration file. Could you please help me? 

My configuration file is as follows:
```yaml
name: BGI
workers:
  jftest:
    type: remote
    scheduler_type: slurm
    work_dir: /public23/home/a21000450/project/C_SiO2_alpha
    pre_run: xxx
    timeout_execute: 60
    host: xxx
    user: xxx
    password: xxx
    max_jobs: 5
    resources:
      partition: xxx
      ntasks: 64
      nodes: 1
      time: "48:00:00"
queue:
  store:
    type: MongoStore
    host: xxx
    database: bgi
    collection_name: jobs
exec_config: {}
jobstore:
  docs_store:
    type: MongoStore
    database: qe
    host: xxx
    collection_name: outputs
  additional_stores:
    data:
      type: GridFSStore
      database: qe
      host: xxx
      collection_name: outputs_blobs
```

The error message is as follows:
```
╭─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│      db_id = '14'                                                                                                                                   │
│       uuid = 'c0536078-004c-48d3-90cc-791e3ee82d4e'                                                                                                 │
│      index = 1                                                                                                                                      │
│       name = 'download_cif'                                                                                                                         │
│      state = 'FAILED'                                                                                                                               │
│      error = Traceback (most recent call last):                                                                                                     │
│                File "/public23/home/a21000450/.conda/envs/py11/lib/python3.11/site-packages/jobflow_remote/jobs/run.py", line 69, in run_remote_job │
│                  response = job.run(store=store)                                                                                                    │
│                             ^^^^^^^^^^^^^^^^^^^^                                                                                                    │
│                File "/public23/home/a21000450/.conda/envs/py11/lib/python3.11/site-packages/jobflow/core/job.py", line 604, in run                  │
│                  response = function(*self.function_args, **self.function_kwargs)                                                                   │
│                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                   │
│              TypeError: 'dict' object is not callable                                                                                               │
│     remote = {'step_attempts': 0, 'process_id': '4588298', 'prerun_cleanup': False}                                                                 │
│ created_on = '2025-03-06 15:44'                                                                                                                     │
│ updated_on = '2025-03-06 16:21'                                                                                                                     │
│ start_time = '2025-03-06 16:20'                                                                                                                     │
│   end_time = '2025-03-06 16:20'                                                                                                                     │
│   metadata = {}                                                                                                                                     │
│    run_dir = '/public23/home/a21000450/project/C_SiO2_alpha/c0/53/60/c0536078-004c-48d3-90cc-791e3ee82d4e_1'                                        │
│    parents = []                                                                                                                                     │
│   priority = 0                                                                                                                                      │
│     worker = 'jftest'                                                                                                                               │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

### LINKED ISSUE #264: Warn user to not use the same database for queue and output store
Add this to the documentation clearly and maybe also add a check in jf project check ?

### LINKED ISSUE #266: The `run_dir` text is often cutoff in the `jf job info ID` print-out
When doing `job info ID`, one of the most valuable pieces of information is the `run_dir`. However, the `run_dir` is cutoff on the screen unless the screen is wide enough, and there is no clear way to access it otherwise.

I wonder if there is a way with `typer` to have the text be wrapped?

### LINKED ISSUE #273: Put CECAM slides in the docs?
I just checked out the Jobflow Remote slides from the CECAM workshop and they are excellent!! In fact, they are so good that I would highly encourage you to consider linking to them from the docs!!

gpetretto:
Some small fixes and new functionalities:

* Avoid truncation of text in `jf job info` #266
  After some tests it seems that while it is possible to pass `overflow="fold"` to the `print` method of `Console` in rich, the value gets reset to `ellipsis` somewhere during the visualization. Unfortunately I don't have the time to dig more into rich, so for the moment I created our custom version of `render_scope` that prevents the issue.
* Better error message if the Job cannot be deserialized by the worker: #263
* Test implementation of a `jf job todir` option in the CLI, that opens a shell directly on the worker and moves to the job directory. Since this runs directly an ssh command I have implemented a converter from `Connection` to the string to be executed, but this of course cannot cover all possible use cases. At least it works for the workers that I have defined in my project.
* "rerun" for certain states switched the state to `CHECKED_OUT` supposedly to save some time to the runner, but this could cause problems since some actions cannot be performed if the job is not `READY`. Now it brings back the Job to `READY`. 
* `jf runner restart`, equivalent to stop+start. To smooth the restart when updating the project configuration. Partially addressing #155
* If possible create indexes in the JobStore collection as resolving references may be slow for large databases otherwise. See https://github.com/materialsproject/jobflow/issues/627

### DIFF
--- a/doc/source/user/projectconf.rst
+++ b/doc/source/user/projectconf.rst
@@ -56,7 +56,16 @@ For all these folders the paths are set with defaults, but can be customised set
 
 .. warning::
   The project name does not take into consideration the configuration file name.
-  For coherence it would be better to give use the project name as file name.
+  For coherence it would be better to use the project name as file name.
+
+.. dropdown:: Example
+
+    Standard usage will not require filling in all directories path and usually
+    just the name can be provided in the configuration:
+
+    .. code-block:: yaml
+
+        name: my_project
 
 .. _projectconf worker:
 
@@ -111,6 +120,68 @@ configurations. Those can better be set with the :ref:`projectconf execconfig`.
     through the ``scheduler_username`` option. Jobflow-remote will use that as a filter,
     instead of the list of ids.
 
+.. dropdown:: Example
+
+    Several workers of different kinds can be defined:
+
+    .. code-block:: yaml
+
+        workers:
+          my_cluster_front_end: # A worker on the front end of the cluster
+            scheduler_type: shell
+            work_dir: /path/to/run/dir
+            # Activate the conda environment on the worker
+            pre_run: "\neval \"$(conda shell.bash hook)\"\nconda activate jfr\n"
+            type: remote
+            # No connection details if they are defined in the ~/.ssh/config file
+            host: my_cluster
+          my_cluster: # same cluter, a worker using the SLURM queue
+            scheduler_type: slurm
+            work_dir: /path/to/run/dir
+            resources:
+            pre_run: "\neval \"$(conda shell.bash hook)\"\nconda activate jfr\n"
+            type: remote
+            host: my_cluster
+          another_cluster: # A remote worker on another cluster
+            scheduler_type: slurm
+            work_dir: /path/to/another/run/dir
+            pre_run: source /data/venv/jobflow/bin/activate
+            type: remote
+            # also possible to define connection details here
+            host: another.cluster.host.net
+            user: username
+            key_filename: /path/to/ssh/private/key
+          another_cluster_batch: # A batch worker on the second cluster
+            scheduler_type: slurm
+            work_dir: /path/to/another/run/dir
+            # Each batch job will run on two nodes with 24 cores each
+            resources:
+              nodes: 2
+              ntasks_per_node: 24
+              mem: 95000
+              partition: xxx
+            pre_run: source /data/venv/jobflow/bin/activate
+            type: remote
+            host: another.cluster.host.net
+            user: username
+            key_filename: /path/to/ssh/private/key
+            # maximum 5 batch Slurm jobs in the queue at the same time
+            max_jobs: 5
+            # This will determine that the worker is a "batch" worker
+            batch:
+              jobs_handle_dir: /some/remote/path/run/batch_handle_slurm
+              work_dir: /some/remote/path/run/batch_work_slurm
+              # two jobflow Job executed in parallel in a single SLURM submission
+              parallel_jobs: 2
+          local_shell: # A local worker running in the shell
+            scheduler_type: shell
+            work_dir: /local/path/to/run/jobs
+            pre_run: "\neval \"$(conda shell.bash hook)\"\nconda activate atomate2_pyd2\n"
+            type: local
+            # Limit the number of local jobs. Since there is no queue they can all
+            # start simultaneously otherwise
+            max_jobs: 3
+
 .. _projectconf jobstore:
 
 JobStore
@@ -128,6 +199,28 @@ in this project.
 
     The ``JobStore`` should be defined in jobflow-remote's configuration file.
     The content of the standard jobflow configuration file will be ignored.
+
+.. warning::
+
+    If you have been using jobflow without jobflow-remote and you have a
+    ``JobStore`` defined in a ``jobflow.yaml`` it will be ignored. Only the
+    definition in the jobflow-remote configuration file will be considered
+
+.. dropdown:: Example
+
+    Define a ``JobStore`` similarly to standard jobflow
+
+    .. code-block:: yaml
+
+        jobstore:
+          docs_store:
+            type: MongoStore
+            host: <host name>
+            port: 27017
+            username: <username>
+            password: <password>
+            database: <database name>
+            collection_name: outputs
 
 .. _projectconf queuestore:
 
@@ -151,6 +244,34 @@ of these two collections can also be customized.
     Some key operations required by jobflow-remote on the collections are not
     supported by any file based MongoDB implementation at the moment.
 
+.. warning::
+
+    If the ``JobStore`` is also based on a MongoDB, it is often convenient to have
+    its main ``docs_store`` in the same database as the ``queue`` store, in that
+    case it is important that the two do **not point to the same collection**.
+    Unexpected errors may happen otherwise.
+
+.. dropdown:: Example
+
+    Define a queue store as maggma store. It is possible to use the same syntax
+    as for the ``JobStore``. Customizing the names of the additional collections
+    is also possible but not necessary
+
+    .. code-block:: yaml
+
+        queue:
+          store:
+            type: MongoStore
+            host: <host name>
+            port: 27017
+            username: <username>
+            password: <password>
+            database: <database name>
+            collection_name: jobs
+          flows_collection: flows
+          auxiliary_collection: jf_auxiliary
+
+
 .. _projectconf execconfig:
 
 Execution configurations
@@ -160,6 +281,35 @@ It is possible to define a set of ``ExecutionConfig`` objects to quickly set up
 configurations for different kind of Jobs and Flow. The ``exec_config`` key
 contains a dictionary where the keys are the names associated to the configurations
 and for each a set of instruction to be set before and after the execution of the Job.
+See the :ref:`tuning exec_config` section for more details and a usage examples.
+
+.. dropdown:: Example
+
+    Multiple configurations can be defined, for example one for each version of an
+    external code or for different software requirements. If multiple workers
+    are present, different ``exec_config`` will need to be defined for each of them.
+
+    .. code-block:: yaml
+
+        exec_config:
+          xxx_v1_1_my_cluster:
+            modules:
+            - releases/2021b
+            - intel/2021b
+            export:
+              PATH: /path/to/executable/v1.1:$PATH
+            pre_run:
+            post_run:
+          xxx_v3_2_my_cluster:
+            modules:
+            - releases/2023b
+            - intel/2023b
+            export:
+              PATH: /path/to/executable/v3.2:$PATH
+          yyy_local:
+            export:
+              PATH: /path/to/local/executable:$PATH
+            pre_run: "echo 'test'\necho 'test2'"
 
 Runner options
 --------------
@@ -172,6 +322,26 @@ reasonable values are set for the delay between each check of the database for
 different kind of actions performed by the ``Runner``. These intervals can be
 changed to better fit your needs. Remind that reducing these intervals too much
 may put unnecessary strain on the database.
+
+.. dropdown:: Example
+
+    Most of the times default values should be fine. Here is how to customize
+    the Runner execution
+
+    .. code-block:: yaml
+
+        runner:
+          delay_checkout: 10
+          delay_check_run_status: 10
+          delay_advance_status: 10
+          delay_update_batch: 10
+          lock_timeout: 86400
+          delete_tmp_folder: true
+          max_step_attempts: 3
+          delta_retry:
+          - 30
+          - 300
+          - 1200
 
 Metadata
 --------
--- a/src/jobflow_remote/cli/job.py
+++ b/src/jobflow_remote/cli/job.py
@@ -1,6 +1,4 @@
 import io
-import os
-import shlex
 from datetime import datetime
 from pathlib import Path
 from typing import Annotated, Optional
@@ -105,7 +103,7 @@ def jobs_list(
         typer.Option(
             "--error",
             "-e",
-            help="Select the jobs in FAILED and REMOTE_ERROR state",
+            help="Select the jobs in FAILED and REMOTE_ERROR state. Compatible with other state filtering options",
         ),
     ] = False,
     running: Annotated[
@@ -113,7 +111,8 @@ def jobs_list(
         typer.Option(
             "--running",
             "-run",
-            help="Select the jobs in one of the running states (from CHECKED_OUT to DOWNLOADED)",
+            help="Select the jobs in one of the running states (from CHECKED_OUT to DOWNLOADED). "
+            "Compatible with other state filtering options",
         ),
     ] = False,
     stored_data_keys: Annotated[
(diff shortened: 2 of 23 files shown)


src/jobflow_remote/jobs/daemon.py
```diff
@@ -856,6 +856,47 @@ def shut_down(self, raise_on_error: bool = False) -> bool:
...
+        with self.lock_runner_doc(allow_missing=True) as lock:
+            doc = lock.locked_document
+            doc_error = self._check_running_runner(doc, raise_on_error=raise_on_error)
+            if doc_error:
+                logger.error(doc_error)
+                return False
+            status = self.check_status()
+            if status not in (
+                DaemonStatus.RUNNING,
+                DaemonStatus.PARTIALLY_RUNNING,
```
davidwaroquiers, line 882: Just wondering if we should allow SHUT_DOWN and STOPPED here. If it is already stopped and we do restart, maybe it's fine to just start it ?

tests/db/cli/test_runner.py
```diff
@@ -53,6 +53,10 @@ def test_std_operations(
         required_out=[*info_required, "hostname", "last_pinged", "processes_info"],
     )
 
+    run_check_cli(
+        ["runner", "restart"],
+    )
```
davidwaroquiers, line 59: Should this check something ? I don't remember but the runner has a log of what he does right ? Does it have some history about something like "I stopped" then "I started" ? Not sure if it is needed anyway but just wondering.

src/jobflow_remote/cli/job.py
```diff
@@ -171,7 +182,16 @@ def jobs_list(
     db_sort: list[tuple[str, int]] = [(sort.value, 1 if reverse_sort else -1)]
 
     if error:
-        state = [JobState.REMOTE_ERROR, JobState.FAILED]
+        if state:
+            state.extend(ERROR_STATES)
+        else:
+            state = ERROR_STATES
```
davidwaroquiers, line 187: So this means if you pass both a -s and a --error (and/or below a --running), then it queries all the states ? E.g. --running -s COMPLETED will get all jobs that are either in one of the running states or in COMPLETED state ? If that's the case, I think it should maybe be documented explicitly. One can even do --running --error right ? I think it's fine but maybe a note somewhere in the jf documentation would be good.

src/jobflow_remote/cli/job.py
```diff
@@ -771,6 +793,74 @@ def report(
...
+        typer.Option(
+            "--command",
+            "-c",
+            help="Only print the ssh command that will be used to connect",
+        ),
+    ] = False,
+):
+    """
+    Connect to the worker and go to the job run_dir.
+    For remote workers and explicit ssh command is generated, so only a subset
```
davidwaroquiers:
```suggestion
    For remote workers an explicit ssh command is generated, so only a subset
```

src/jobflow_remote/remote/host/base.py
```diff
@@ -147,6 +147,25 @@ def rmtree(self, path: str | Path, raise_on_error: bool = False) -> bool:
...
+        ----------
+        dir_path
+            The directory to reach
+        target_shell
+            Shell command to be used to start the shell on the worker to access the
+            target directory
+
+        Returns
+        -------
+            The string to be used to reach the chose directory.
```
davidwaroquiers:
```suggestion
            The string to be used to reach the chosen directory.
```

src/jobflow_remote/remote/host/local.py
```diff
@@ -156,3 +156,22 @@ def onerror(func, dir_path, err):
...
+            The directory to reach
+        target_shell
+            Shell command to be used to start the shell on the worker to access the
+            target directory
+
+        Returns
+        -------
+            The string to be used to reach the chose directory.
+        """
+        return f"cd {dir_path}; {target_shell}"
```
davidwaroquiers: What if a user passes (no idea why he would do that ...) "bash; rm -rf *" for example for the target shell. Is that a security issue ?

src/jobflow_remote/remote/host/remote.py
```diff
@@ -430,6 +430,27 @@ def _check_connected(self) -> bool:
...
+        target_shell
+            Shell command to be used to start the shell on the worker to access the
+            target directory
+
+        Returns
+        -------
+            The string to be used to reach the chose directory.
+        """
+        ssh_cmd = build_ssh_command(self.connection)
+        return f'{ssh_cmd} "cd {dir_path}; {target_shell}"'
```
davidwaroquiers: Same as for the local. Is it secure ?

src/jobflow_remote/remote/host/remote.py
```diff
@@ -492,3 +513,107 @@ def get_sources(self):
         #     )
 
         yield Interactive(username=self.username)
+
+
+def build_ssh_command(connection: fabric.Connection) -> str:
+    """
+    Given a fabric Connection generate the ssh command replicating
+    that same connection.
```
davidwaroquiers: I guess you tested but isn't it possible to just do connection.run("bash") or something similar ? (that still keeps the security open if it is a indeed an issue)

src/jobflow_remote/cli/utils.py
```diff
@@ -714,3 +717,70 @@ def confirm_project_name(style: str | None = "red", exit_on_error: bool = True)
...
+        )
+        items_table.add_row(
+            key_text,
+            Pretty(
+                value,
+                highlighter=highlighter,
+                indent_guides=indent_guides,
+                max_length=max_length,
+                max_string=max_string,
+                overflow=overflow,
```
davidwaroquiers, line 778: I guess it's the case for the other options of Pretty, but should/could/is there a need for/... having overflow depend on the column/item ?

tests/integration/test_workers.py
```diff
@@ -457,3 +457,40 @@ def test_sanitize(worker, job_controller):
...
+    run_check_cli(
+        ["job", "todir", j.uuid],
+        required_out="The selected Job does not have a run_dir defined yet",
+        error=True,
+    )
+
+    runner = Runner()
+    runner.run_all_jobs(max_seconds=MAX_TRY_SECONDS)
+
+    # outputs from the connected shell are not either not executed properly
```
davidwaroquiers:
```suggestion
    # outputs from the connected shell are either not executed properly
```

davidwaroquiers (commented):
Very nice additions and fixes. My main comment is probably about the todir option which is very nice but just wondering if 1/ there is no security issue and/or what is the "garde-fou" and 2/ isn't there something directly from fabric ? or using fabric itself ?

src/jobflow_remote/remote/host/remote.py
davidwaroquiers: I saw this in Fabric: https://docs.fabfile.org/en/latest/api/connection.html#fabric.connection.Connection.shell
Would this be an option ?

davidwaroquiers: Also note this in fabric's doc:
image

src/jobflow_remote/jobs/daemon.py
gpetretto: One point is that "restart" does not have all the arguments of "start" (e.g., single/multiple processes). The idea is that the runner is restarted with the same conditions it was in when it was stopped. I mainly did that to avoid bothering with implenting complicated cross checks or duplicating options. So, in general, this definitely excludes SHUT_DOWN and I would avoid the additional effort of supporting the restart from this state. Potentially STOPPED should be fine. If you think it would be beneficial I will make a few tests and I will add it.

tests/db/cli/test_runner.py
gpetretto: no, there is no such log. There is just the python log when a loggin message is generated. I expect that going through that would be difficult.
One could check if the PIDs of the processes are changed, but in principle there is no guarantee they will actually change. So in the end I am not sure if there is any way to guarantee that the daemon has actually stopped and restarted.

src/jobflow_remote/cli/job.py
gpetretto: Yes, this is a change with the previous behaviour. I thought that having `--running --error` is a legit use case and was easy to implement. I will add a comment saying that all the state options are compatible.

src/jobflow_remote/remote/host/local.py
gpetretto: In this point of the code it is just generating a string, not executing the command. So here I don't necessarily see as a security issue. When the command is executed (in the CLI), this is to access the shell which gives exactly the same rights to the users as those in the `bash; rm -rf*`. I would rather consider this a secutiry issue if the `bash` command was executed with higher rights (e.g. within sudo), but in this case, you can just do `bash` and run your `rm -rf *` once you have connected, if you wish.
Another kind of security issue would be if we were to run simulations on an internal cluster and only expose access to the user thorugh, for example, a GUI. But in that case I would directly not expose any of the configuration or this low lever machinery to the user.

Som the main point to me is that here we are not giving any more power to the user than what he would normally have. A user can always add `rm -rf ~/*` in the `pre_run` of the worker or the exec_config. If the users want to delete their own home folder I think they can do 😄

Do you think there is some other case where this could be exploited by some malicious actor?

src/jobflow_remote/remote/host/remote.py
gpetretto: At the very beginning of the jobflow-remote development I had tested the options to connect with fabric directly wihtout success. If I remember correctly I had tried `c.shell()`, `c.run("bash")` and `c.client.invoke_shell()` without maanging to get access to the shell. So this time around I went directly for a different solution.
However, given your suggestion I tried again now. The old solutions still do not seem to work, but after some searches it seems that the trick it to use `c.run("bash", pty=True)`.

I can make a few more tests, but if everything works I will switch to this, as it seems more reliable than the hand made implementation of the ssh command 😅

src/jobflow_remote/cli/utils.py
gpetretto: you mean to customize the `overflow` based on the cell of the table generated by the `jf job info`? I don't think that at this stage would be very useful. It seems that the only other thing that could overflow would be the error messages, and I think it would also be good that those are fully shown as well.

src/jobflow_remote/remote/host/remote.py
gpetretto: If everything works, I will probably still leave the argument for which shell to use (with `bash` as a default) in the CLI, so it is still worth discussing if this should be considered as a vulnerability.

src/jobflow_remote/jobs/daemon.py
davidwaroquiers: Indeed, did not think about that. I think it is fine to just keep it for just RUNNING and PARTIALLY_RUNNING. As you wish. Resolving this already anyway.

tests/db/cli/test_runner.py
davidwaroquiers: Maybe from the running_runner document in the database, which contains the start_time, and checking that before and after restart that the start_time has changed ?
Otherwise indeed there is the python log that logs the stopping and starting, but that's indeed a bit cumbersome ...

src/jobflow_remote/remote/host/local.py
davidwaroquiers: Yes indeed this was just the generation of the command but my comment was more general. I think it's fine.

src/jobflow_remote/cli/utils.py
davidwaroquiers: Sure, sounds reasonable. Resolving this.

tests/db/cli/test_runner.py
gpetretto: good point. I added this check

src/jobflow_remote/cli/job.py
gpetretto: added the comment

src/jobflow_remote/remote/host/remote.py
gpetretto: switched to the fabric/invoke version. Not sure if it better in general, but it is cleaner and supports more use cases

gpetretto:
I have switched to fabric to handle the connection to the worker. 
Changed the flag to enable tab completion for the CLI (#272).
Added links to the slides and videos of the CECAM school (#273)

gpetretto:
Given last day's discussions, I have tried to improve the documentation of the project set up #264 #274 #277
@Andrew-S-Rosen do you think this would be enough?

doc/source/user/projectconf.rst
```diff
@@ -151,6 +244,34 @@ of these two collections can also be customized.
     Some key operations required by jobflow-remote on the collections are not
     supported by any file based MongoDB implementation at the moment.
 
+.. warning::
+
+    If the ``JobStore`` is also based on a MongoDB, it is often convenient to have
+    its main ``docs_store`` in the same database as the ``queue`` store, in that
+    case it is important that the two do **not point to the same collection**.
+    Unexpected errors may happen otherwise.
```
davidwaroquiers, line 252: Maybe the "jf project check" can give an error if it's the case ?

davidwaroquiers: Ah it's done I see :)

davidwaroquiers (approved):
All good to me (I had a comment about jf project check and then I saw it was there so forget about the comment :p)

Andrew-S-Rosen:
I don't have time to look in super great depth, but from a quick glance this looks great! Thank you!!

gpetretto:
> I don't have time to look in super great depth, but from a quick glance this looks great! Thank you!!

Thanks for having a look.
I will start merging this. I will try to gradually improve the documentation but other suggestions are welcome about which portions could benefit some updates.

PR #288, opened 2025-04-18
fix ignoring resolve_references in JobConfig

gpetretto:
jobflow-remote, at variance with other jobflow mangers, does not take into account the `resolve_references` option in `JobConfig`. Currently references are always resolved by the runner without checking the job configuration.
This PR fixes the issue.
Thanks @VicTrqt and @emarazzi for pointing this out.

### DIFF
--- a/src/jobflow_remote/remote/data.py
+++ b/src/jobflow_remote/remote/data.py
@@ -239,6 +239,8 @@ def resolve_job_dict_args(job_dict: dict, store: JobStore) -> dict:
     Similar to Job.resolve_args, but without the need to deserialize the Job.
     The references are resolved inplace.
 
+    If resolve_references in job.config is False references are not resolved.
+
     Parameters
     ----------
     job_dict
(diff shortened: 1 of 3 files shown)


davidwaroquiers (commented):
Thanks for this!
Two questions here. Maybe it is irrelevant, pardon me :)
What happens with external references ?
Does that solve the issue you were mentioning that output references are always resolved ?

gpetretto:
> What happens with external references ? 

You mean references that do not belong to the same Flow? These will not be resolved as well. It is the Job itself that controls if the references that are passed to it are resolved or not (e.g. https://github.com/materialsproject/jobflow/blob/5c8546bf4bdd4e4c39e1877986d38aa9a57c83fb/src/jobflow/core/job.py#L1319). And in principle the Job in jobflow does not even know if an external reference belongs to the same Flow or not

> Does that solve the issue you were mentioning that output references are always resolved ?

Yes, this is to address that issue that I was mentioning

davidwaroquiers (approved):
All good to me

PR #302, opened 2025-05-26
Fix testing, upgrade test requirements

gpetretto:
Latest changes in click and typer require small adjustments of the CLI tests.

### DIFF
--- a/src/jobflow_remote/testing/cli.py
+++ b/src/jobflow_remote/testing/cli.py
@@ -37,25 +37,32 @@ def run_check_cli(
         terminal_width=terminal_width,
     )
 
+    # Since typer 0.16.0 and click 8.2 the stderr is necessarily separated from the output.
+    # To keep backward compatibility with previous click versions handle both defaults cases.
+    try:
+        result_out = result.stdout + "\n" + result.stderr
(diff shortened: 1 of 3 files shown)


src/jobflow_remote/testing/cli.py
```diff
@@ -37,25 +37,32 @@ def run_check_cli(
         terminal_width=terminal_width,
     )
 
+    # Since typer 0.16.0 and click 8.2 the stderr is necessarily separated from the output.
+    # To keep backward compatibility with previous click versions handle both defaults cases.
```
davidwaroquiers, line 41: Do you think we'd need to test the different versions ? (up to some time of course) This is probably for another PR. I think pymatgen is testing the latest versions and the earliest versions it can install or something like that. At least I remember something like that somewhere.

davidwaroquiers (approved):
All good to me. I just did one comment for which versions of the dependencies to test but It does not need to be tackled here (or even maybe at all, it's a decision we could make).

src/jobflow_remote/testing/cli.py
gpetretto: thanks. I will check this.

PR #307, opened 2025-06-08
Add --count option in CLI

gpetretto:
Adding a `--count` option for `jf job list` and `jf flow list`, printing just the number of jobs/flows matching the query.

### DIFF
--- a/src/jobflow_remote/cli/flow.py
+++ b/src/jobflow_remote/cli/flow.py
@@ -17,6 +17,7 @@ from jobflow_remote.cli.formatting import (
 from jobflow_remote.cli.jf import app
 from jobflow_remote.cli.jfr_typer import JFRTyper
 from jobflow_remote.cli.types import (
+    count_opt,
     days_opt,
     db_ids_opt,
     delete_all_opt,
@@ -74,6 +75,7 @@ def flows_list(
     metadata: metadata_opt = None,
     locked: locked_flow_opt = False,
     verbosity: verbosity_opt = 0,
+    count: count_opt = False,
     max_results: max_results_opt = 100,
     sort: sort_opt = SortOption.UPDATED_ON,
     reverse_sort: reverse_sort_flag_opt = False,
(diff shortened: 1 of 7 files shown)


src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -2950,6 +2950,8 @@ def count_flows(
         start_date: datetime | None = None,
         end_date: datetime | None = None,
         name: str | None = None,
+        metadata: dict | None = None,
+        locked: bool = False,
```
davidwaroquiers, line 2956: Good that you added. Makes we wonder whether the _build_query_flows and _build_query_jobs should be without any default arguments. They are private methods so the caller should itself have some default value that are passed down. I guess metadata and locked were added after the rest somehow and maybe we forgot to change/add it in the caller classes/methods. If there is no default arg in the _build_... methods, then if we add something in there, we will directly know if we missed one somewhere. What do you think ? (maybe something like that also applies to other private methods). I don't know if this is a standard practice.

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -2975,6 +2977,11 @@ def count_flows(
         name
             Pattern matching the name of Flow. Default is an exact match, but all
             conventions from python fnmatch can be used (e.g. *test*)
+        metadata
+            A dictionary of the values of the metadata to match. Should be an
+            exact match for all the values provided.
+        locked
+            If True only locked Flows will be selected.
```
davidwaroquiers:
```suggestion
            If True only locked Flows will be counted.
```

src/jobflow_remote/cli/flow.py
```diff
@@ -88,31 +90,46 @@ def flows_list(
...
+                db_ids=db_id,
+                flow_ids=flow_id,
+                states=state,
+                start_date=start_date,
+                end_date=end_date,
+                name=name,
+                metadata=metadata,
+                locked=locked,
+            )
+        out_console.print(f"Selected Flows: {n_flows}")
```
davidwaroquiers:
```suggestion
        out_console.print(f"Number of Flows matching query: {n_flows}")
```
Not sure if it is more clear or less clear. What do you think ? I let you decide :)

src/jobflow_remote/cli/job.py
```diff
@@ -211,40 +208,63 @@ def jobs_list(
                 name=name,
                 metadata=metadata,
                 workers=worker_name,
-                limit=max_results,
-                sort=db_sort,
             )
+        out_console.print(f"Selected jobs: {n_jobs}")
```
davidwaroquiers: Same as above for the Flows, I don't know if it is more clear or not and I let you decide :)
```suggestion
        out_console.print(f"Number of jobs matching query: {n_jobs}")
```

davidwaroquiers (approved):
Very nice, as always :)
Just one comment on the private methods for building the queries, as we discussed on Teams, and a few small suggestions. Otherwise all good to me.

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: It seems a good point. I removed the defaults for those private methods

src/jobflow_remote/cli/flow.py
gpetretto: I used "Number of" I hope it is fine. We can always change at a later stage if it is reported that it is not clear

PR #310, opened 2025-06-16
Update release.yml

davidwaroquiers:
Testing to replace RELEASE_PAT_BOT by GITHUB_TOKEN


ml-evs (commented):
I think this likely needs something like:

```
    permissions:
      contents: write
```

so that it can commit changes to the changelog. More at https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token

gpetretto (approved):

PR #313, opened 2025-06-17
bugfix and test release workflow

gpetretto:
Two bugfixes: 
* In case of limited workers, when starting the runner the order of operations was such that the currently active jobs were not taken into account.
* The serialization of the JobDoc was using `enum_values=True` to also serialize the `Job` object. While this is fine for the rest of the document, where pydantic converts the value back to an Enum during deserialization, this is not the case for example for attributes of a Maker. While this may still be working in some cases, the Enum object should likely be preserved. This could be a potentially breaking change, if anything relied on the old serialization or if for some specific job the serialization becomes unrelieable, but I consider the previous implementation incorrect.

Test changing the permissions in the release workflow.

### DIFF
--- a/src/jobflow_remote/jobs/runner.py
+++ b/src/jobflow_remote/jobs/runner.py
@@ -275,15 +275,9 @@ class Runner:
                 self.checkout
             )
 
-        if transfer or queue or complete:
-            try:
-                self.advance_state(states)
-            except Exception:
-                logger.exception("Error during initial advance_state")
-            scheduler.every(self.runner_options.delay_advance_status).seconds.do(
-                self.advance_state, states=states
-            )
-
+        # first perform this step with check run status and update the status of
+        # the limited workers (if any). Otherwise, upon start of the runner the initial
+        # count of the currently submitted jobs is 0.
         if queue:
             try:
                 self.check_run_status()
@@ -315,6 +309,15 @@ class Runner:
                 scheduler.every(self.runner_options.delay_update_batch).seconds.do(
                     self.update_batch_jobs
                 )
+
+        if transfer or queue or complete:
+            try:
+                self.advance_state(states)
+            except Exception:
+                logger.exception("Error during initial advance_state")
+            scheduler.every(self.runner_options.delay_advance_status).seconds.do(
+                self.advance_state, states=states
+            )
 
         # all the processes will ping the running_runner document to signal
         # that at least one is still active.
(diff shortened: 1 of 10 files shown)


src/jobflow_remote/testing/__init__.py
```diff
@@ -111,3 +113,18 @@ def no_resolve(ref):
...
+
+
+@dataclass
+class EnumMaker(Maker):
+    e: TestEnum = TestEnum.A
+    name: str = "enum maker"
+
+    @job
+    def make(self):
+        assert isinstance(self.e, TestEnum)
```
davidwaroquiers, line 130: Shouldn't that be actually used to make the test ? The assert isinstance(self.e, TestEnum) is never done in this case if I'm not wrong.
I would have thought we need something that uses the EnumMaker, makes a job and test that (maybe in another python test file).

src/jobflow_remote/jobs/runner.py
```diff
@@ -275,15 +275,9 @@ def run(
...
-                self.advance_state(states)
-            except Exception:
-                logger.exception("Error during initial advance_state")
-            scheduler.every(self.runner_options.delay_advance_status).seconds.do(
-                self.advance_state, states=states
-            )
-
+        # first perform this step with check run status and update the status of
+        # the limited workers (if any). Otherwise, upon start of the runner the initial
+        # count of the currently submitted jobs is 0.
```
davidwaroquiers, line 280: Any way to test that somehow ? How did you find out about that bug ?

davidwaroquiers (approved):
All good to me. Just two comments regarding the tests. I don't know if it's easy to add a test for the number of jobs in limited workers. For the Enum problem, unless I'm wrong, there might be something missing for the test no ?

src/jobflow_remote/testing/__init__.py
gpetretto: For some reason I did not add the test file to the commit... I just pushed it. Thanks for spotting that.

src/jobflow_remote/jobs/runner.py
gpetretto: I stopped and restarted the runner and more jobs were submitted, even if the maximum was already reached. There is already a test for the workers with maximum number of jobs: https://github.com/Matgenix/jobflow-remote/blob/a68ff9d2cae0d91478b85c440ddaf4f44742af86/tests/integration/test_advanced_options.py#L81
The problem is that it is already a bit tricky and I would need to add to this the fact of stopping and restarting the runner. Plus I would need to make the jobs sleep for at least 30 seconds to make sure that they don't finish during the restart of the runner. Still, it should be feasible, if you think it is better I can try to add it.

davidwaroquiers: I don't know, it's just that if for some reason, we forget about this, and/or someone else modifies the code (e.g. moving the snippets of code for some reason) and we approve (because we forgot), then we get it back. I would say it would probably be a good thing to have.

davidwaroquiers (approved):
All perfect! Thanks

Issue #289, opened 2025-04-24
Feature request: Pause flow to wait for user interaction


FabiPi3:
Hi @gpetretto and other developers,

I was thinking about some way to programatically pause a flow. The reason for that would be some sort of human interaction is required to judge a certain warning. I could imagine that at the end of a job a Response is returned, which contains some manager_kwargs that specify the pausing. Then all children, but also directly replaced and added jobs/flows will go in a flow_user_paused state. 

Now the user has to check the warning and can use some `jf flow resume` command to play the flow again.

What do you think?

davidwaroquiers:
Hi @FabiPi3 

I think it's a nice idea. I would suppose this should also be something that involves jobflow itself in a way. Pinging @utf on ideas/thoughts about this.

gpetretto:
Hi @FabiPi3, 
I think this is a reasonable use case. However, as @davidwaroquiers suggested, this would necessarily require a modification in the `Response` object directly in jobflow, since it is the only thing a Job can return.

Approaching this from a different point of view though, it seems that what are you asking is very similar to the current `stop_jobflow` functionality. Am I wrong? The key difference at the moment is that jobflow-remote does not allow further modification to a `STOPPED` Job. But this is just a choice we made in the beginning of the development of jobflow-remote and I don't think it would be a problem introducing a `jf flow resume` command that reverts the effect of `stop_jobflow`. What do you think?

FabiPi3:
Hi @gpetretto,
yes, `stop_jobflow`does that in principle. If I could resume than, this would be fine for me.

However, I run some test jobs and noticed the following behavior:

Assume I have a job like: 

```python
@job
def replace_with_job_and_stop_jobflow() -> Response[None]:
    return Response(replace=add(1, 1), stop_jobflow=True)
```

and run this via:

```python
my_job = replace_with_job_and_stop_jobflow()
job_2 = add(1, my_job.output)
print(submit_flow([my_job, job_2], "my_worker", exec_config="my_config", resources={}))
```

of course, `my_job` will run first. Than, `job_2` will go in a stopped state, and the replaced `add` job will be created. But this replaced new job will actually RUN as well. I think this is not the intended behavior. At least as described in my previous post. Should I open a new issue for that?

gpetretto:
Thanks for spotting this. The problem is due to this point of the code: 
[...]

PR #316, opened 2025-07-21
Minor updates: color job list, allow resuming STOPPED jobs

### LINKED ISSUE #291: Freshly created replaced jobs should be stopped as well
As discussed in #289, open a new issue for that.

Assume I have a job like: 

```python
@job
def replace_with_job_and_stop_jobflow() -> Response[None]:
    return Response(replace=add(1, 1), stop_jobflow=True)
```

and run this via:

```python
my_job = replace_with_job_and_stop_jobflow()
job_2 = add(1, my_job.output)
print(submit_flow([my_job, job_2], "my_worker", exec_config="my_config", resources={}))
```

of course, `my_job` will run first. Than, `job_2` will go in a stopped state, and the replaced `add` job will be created. But this replaced new job will actually RUN as well. I think this is not the intended behavior.

gpetretto:
Introducing few minor updates

### color job names
Introducing an option to optionally color the Job names in `jf job list`, so that jobs belonging to the same Flow have the same color. Example:
Screenshot 2025-07-21 alle 16 22 00
Two potential issues here:
* I had to add the `hosts` attribute to the `JobInfo` model, so that additional information is always fetched from the DB when `JobInfo` is retrieved.
* I manually created a list of colors. I definitely do not want to add something like matplotlib as a dependency to get arbitrary colormaps. Alternatively there would be this nice package: https://cmap-docs.readthedocs.io/en/stable/, but it still depends on numpy. I suppose that numpy and matplotlib will likely be installed in any environment using jobflow-remote, but I would rather keep the dependencies to the minimum. It also avoids importing external libraries at runtime, which might be better when using the CLI.

### Resume STOPPED jobs

Up to now `STOPPED` and `USER_STOPPED` jobs were considered as terminal states, which could not be restarted. Following the discussions in #289 it emerged that it would be a convenient use case to allow resuming a `STOPPED` Job. 
This PR also fixes #291. 
Two issues to consider:
* As discussed in #289, if a Job triggers a `stop_jobflow` and another Job is currently running this will not be stopped (this is fine), but if this latter Job dynamically creates new jobs these will be created as `STOPPED`. Currently they are created normally and will be executed. Fixing this will need adding more information to the DB.
* With the current change the `USER_STOPPED` and `PAUSED` state are almost equivalent. Originally they were mostly differentiated in the (im)possibility of restarting, but now they basically differ just for the states of the Jobs to which they can be applied (`jf job stop` can be applied to any running state, `jf job pause` just to `READY` and `WAITING`). We might consider merging these states in a single one. 

@FabiPi3 let me know if you have any comment or if you want to test this before it is merged.

### DIFF
--- a/tests/db/cli/test_job.py
+++ b/tests/db/cli/test_job.py
@@ -16,6 +16,9 @@ def test_jobs_list(job_controller, two_flows_four_jobs) -> None:
     run_check_cli(["job", "list", "-v"])
     run_check_cli(["job", "list", "-vvv"])
 
+    # not checking that the output is actually colored. Just check that runs correctly
+    run_check_cli(["job", "list", "--color"], required_out=outputs)
+
     outputs = ["add1", "READY"]
     excluded = [f"add{i}" for i in range(2, 5)]
     run_check_cli(
@@ -285,7 +288,7 @@ def test_retry(job_controller, two_flows_four_jobs) -> None:
     run_check_cli(["job", "retry", "-did", "2"], required_out="Error while retrying")
 
 
-def test_play_pause(job_controller, two_flows_four_jobs) -> None:
+def test_pause_resume(job_controller, two_flows_four_jobs) -> None:
     from jobflow_remote.jobs.state import JobState
     from jobflow_remote.testing.cli import run_check_cli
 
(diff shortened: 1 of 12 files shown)


FabiPi3:
So just for my understanding, if my inital job spawns three child jobs and stops the flow, all three children will go in the STOPPED state, right? So with `jf job resume` I can resume them one by one? Feels like there would be also a need for a `jf flow resume` to resume the whole flow, so all three jobs at once? Or would I set flow_id to my flow id and the whole flow resumes?

gpetretto:
> So just for my understanding, if my inital job spawns three child jobs and stops the flow, all three children will go in the STOPPED state, right? 

Correct

> So with `jf job resume` I can resume them one by one? Feels like there would be also a need for a `jf flow resume` to resume the whole flow, so all three jobs at once? Or would I set flow_id to my flow id and the whole flow resumes?

You can indeed run
```
jf job resume -fid $FLOW_ID -s STOPPED
```
and it would basically be equivalent to `jf flow resume -fid $FLOW_ID`. But I can also add it, if you think it might be convenient.

FabiPi3:
If you could provide a `jf flow resume $FLOW_ID` command, I think this would be quite beneficial.

tests/db/cli/test_job.py
```diff
@@ -16,6 +16,9 @@ def test_jobs_list(job_controller, two_flows_four_jobs) -> None:
     run_check_cli(["job", "list", "-v"])
     run_check_cli(["job", "list", "-vvv"])
 
+    # not checking that the output is actually colored. Just check that runs correctly
```
davidwaroquiers, line 19: I think it would still be good to test that the colors are indeed applied. It seems possible by capturing the output with rich's Console and checking for the correct ANSI codes.

davidwaroquiers (approved):
All good to me. Just one small comment if it's possible to test the colors ? Otherwise it may be that it works now and then fails later on. Also, is there anything to be modified in the documentation ? (besides the jf tree CLI that will be done automatically).
Thanks a lot!

tests/db/cli/test_job.py
gpetretto: Did not find an easy way to test this with the standard CLI testing. Added a unit test.

FabiPi3:
Sorry I saw it too late but the description of `jf job stop -h` still says:

`The operation is irreversible.`

This is not true anymore.

Issue #320, opened 2025-07-28
Feature Request: Flow_DB_ID option for jf job list command


FabiPi3:
Hi @gpetretto,

a hopefully not too complex request: I would like to pass a DB_ID to the `jf job list` command a see only the jobs belonging to this flow (which is unique through the job which has this DB_ID). This is quite similar to the flow_id option, but there I have to pass the flow UUID which I don't know. I could get it from a `jf flow info` command, but I am too lazy for this 2-step-process to be honest.

Another option would be to change the `jf flow info` command to have also customizable columns like we did for the `jf job list` command but I think this is much more overhead.

gpetretto:
I think this should be relatively easy to do. Right now `jf flow info ID` accepts as ID either the flow uuid or any of the db_ids of the jobs belonging to the flow (the flow does not have its own db_id value). So it should be enough to mimic that behaviour (if the flow id is not a valid uuid, it is interpreted as a db_id of one of the jobs). This would still require some changes to the logic in the JobController, probably in `_build_query_flow`.

PR #321, opened 2025-07-28
Merge custom and automatic queries

FabiPi3:
Enables the use of a custom_query AND other query-options in the `jf job list` command. This is controlled by a new option `merge_queries`.

Is there a reason why this is not a good idea? It seems to work for me just fine.

For the reason of completeness one might need to transfer this also to other cli commands.

### DIFF
--- a/src/jobflow_remote/cli/types.py
+++ b/src/jobflow_remote/cli/types.py
@@ -62,7 +62,7 @@ flow_ids_opt = Annotated[
     typer.Option(
         "--flow-id",
         "-fid",
-        help="One or more flow ids",
+        help="One or more flow ids. Can the db id (i.e. an integer) or a string (i.e. the uuid)",
     ),
 ]
 
--- a/src/jobflow_remote/cli/types.py
+++ b/src/jobflow_remote/cli/types.py
@@ -419,7 +419,8 @@ query_opt = Annotated[
         "--query",
         "-q",
         help="A JSON string representing a generic query in the form of a dictionary. "
-        "Overrides all other query options. Requires knowledge of the internal structure of the DB. "
+        "Overrides all other query options. Except the merge_queries option is enabled. "
+        "Requires knowledge of the internal structure of the DB. "
         "Can be either a list of comma separated key=value pairs or a string with the JSON"
         " representation of a dictionary containing the mongoDB query that "
         'should be performed (e.g \'{"key1.key2": 1, "key3": "test"}\')',
--- a/src/jobflow_remote/cli/types.py
+++ b/src/jobflow_remote/cli/types.py
@@ -419,7 +419,7 @@ query_opt = Annotated[
         "--query",
         "-q",
         help="A JSON string representing a generic query in the form of a dictionary. "
-        "Overrides all other query options. Except the merge_queries option is enabled. "
+        "Keys must not overlap with those from other specified query options. "
         "Requires knowledge of the internal structure of the DB. "
         "Can be either a list of comma separated key=value pairs or a string with the JSON"
         " representation of a dictionary containing the mongoDB query that "
--- a/src/jobflow_remote/jobs/jobcontroller.py
+++ b/src/jobflow_remote/jobs/jobcontroller.py
@@ -6,7 +6,6 @@ import importlib.metadata
 import logging
 import shutil
 import traceback
-import uuid
 import warnings
 from collections import defaultdict
 from contextlib import ExitStack
@@ -60,6 +59,7 @@ from jobflow_remote.remote.data import (
 )
 from jobflow_remote.remote.queue import QueueManager
 from jobflow_remote.utils.data import (
+    check_valid_uuid,
     deep_merge_dict,
     get_past_time_rounded,
     get_utc_offset,
@@ -84,14 +84,6 @@ if TYPE_CHECKING:
 
 
 logger = logging.getLogger(__name__)
-
-
-def check_valid_uuid(uuid_str) -> bool:
-    with contextlib.suppress(ValueError):
-        uuid_obj = uuid.UUID(uuid_str)
-        if str(uuid_obj) == uuid_str:
-            return True
-    return False
 
 
 class JobController:
@@ -265,21 +257,14 @@ class JobController:
         db_ids = [db_ids] if isinstance(db_ids, str) else db_ids or []
 
         flow_ids = [flow_ids] if isinstance(flow_ids, str) else flow_ids or []
-        flow_uuids, flow_db_ids = [], []
-        for flow_id in flow_ids:
-            if check_valid_uuid(flow_id):
-                flow_uuids.append(flow_id)
-            else:
-                flow_db_ids.append(flow_id)
-
-        for flow_db_id in flow_db_ids:
-            flows_info = self.get_flows_info(
-                db_ids=flow_db_id,
-                limit=1,
-                full=True,  # needed?
-            )
-            if flows_info:
-                db_ids.extend(flows_info[0].db_ids)
+        flow_uuids = [
+            fid
+            if check_valid_uuid(fid)
+            else self.flows.find_one(
+                {"ids": {"$elemMatch": {"0": fid}}}, projection=["uuid"]
+            )["uuid"]
+            for fid in flow_ids
+        ]
 
         if isinstance(states, JobState):
             states = [states]
--- a/src/jobflow_remote/jobs/jobcontroller.py
+++ b/src/jobflow_remote/jobs/jobcontroller.py
@@ -413,6 +413,7 @@ class JobController:
 
     def get_jobs_info(
         self,
+        custom_query: dict | None = None,
         job_ids: tuple[str, int] | list[tuple[str, int]] | None = None,
         db_ids: str | list[str] | None = None,
         flow_ids: str | list[str] | None = None,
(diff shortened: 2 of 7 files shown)


FabiPi3:
Tagging @gpetretto and @davidwaroquiers for their opinion

gpetretto:
Thanks @FabiPi3, I think adding the possibility of merging the queries is fine. Before I go into a detailed review I would ask you to make a few changes. Mainly I dont't think it is really necessary to have the additional `merge_queries` option. I propose one of two options:
1) the merge is always allowed with the values from the generic query overwrite the other options, is there is an overlap (basically what you already do now, except that there is no check for the `merge_queries`)
2) The merge is allowed, but only if there is no overlap. In case of overlap an error is thrown. (The overlap could just be that in the final query dictionary, before doing the merge a check is done to verify that the keys do not overlap)

I do not have a strong preference. Do you have a preferred option, given your use case?
In any case specify the behaviour in the docstrings of the methods and of the CLI.

The other change is to move to `_build_query_job` the merge of the filter dictionaries (and the check of not overlap, if option 2 is chosen) . This should make it slightly easier to handle the same option in other methods of JobController.

FabiPi3:
Thanks for your answer. Since I am on vacation until end of next week I will start working on this afterward

FabiPi3:
I have tried to include your comments and also added a fix for #320 

Not yet super convinced about the solution, but it works.

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -299,7 +326,13 @@ def _build_query_job(
         if workers:
             query["worker"] = {"$in": workers}
 
-        return query
+        custom_query_dict = custom_query or {}
```
gpetretto: maybe avoid defining an additional variable, which is a bit confusing, since custom_query is also a `dict`. You can just do `custom_query = custom_query or {}`

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -250,10 +262,25 @@ def _build_query_job(
...
+        flow_ids = [flow_ids] if isinstance(flow_ids, str) else flow_ids or []
+        flow_uuids, flow_db_ids = [], []
+        for flow_id in flow_ids:
+            if check_valid_uuid(flow_id):
+                flow_uuids.append(flow_id)
+            else:
+                flow_db_ids.append(flow_id)
+
+        for flow_db_id in flow_db_ids:
+            flows_info = self.get_flows_info(
```
gpetretto: For anything that remains internal to JobController I consider fine to use direct MongoDB queries to speed up the operation. You can replace this with something like:

```python
self.flows.find_one({"jobs": flow_db_id}, projection=["uuid"])
```

that will only fetch the strict minimum information and will not need to creat the full FlowInfo object.

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -85,6 +86,14 @@
 logger = logging.getLogger(__name__)
 
 
+def check_valid_uuid(uuid_str) -> bool:
```
gpetretto: Can you move this to utils/data.py?

gpetretto (commented):
Thanks for the updates @FabiPi3!

I should say that I am not very conviced either by the solution of #320, but I also have to admit that I don't have any better solution. We discussed a bit with @davidwaroquiers and we think it may still be fine. I would agree to merge it if it works fine for you. 

I left a few small comments.

src/jobflow_remote/jobs/jobcontroller.py
FabiPi3: Strictly speaking `custom_query` is not a `dict`. It is `dict|None` which is a different type. `custom_query_dict` has the type `dict`. Mypy (configured correctly) is quite strict about that and does not allow to change the type of a variable as this would lead to even more confusion.
What do you prefer?

src/jobflow_remote/jobs/jobcontroller.py
FabiPi3: Sure.

src/jobflow_remote/jobs/jobcontroller.py
FabiPi3: I have reworked the query.

src/jobflow_remote/cli/types.py
```diff
@@ -62,7 +62,7 @@
     typer.Option(
         "--flow-id",
         "-fid",
-        help="One or more flow ids",
+        help="One or more flow ids. Can the db id (i.e. an integer) or a string (i.e. the uuid)",
```
davidwaroquiers, line 65: The db id can also be an integer with a prefix (if you set db_id_prefix in the queue settings). I'd just put that it can be the db id or uuid. If you have another suggestion, go ahead.
```suggestion
        help="One or more flow ids. Each flow id can be the db id or a uuid",
```

src/jobflow_remote/utils/data.py
```diff
@@ -239,3 +241,24 @@ def suuid() -> str:
     from uuid import uuid4
 
     return str(uuid4())
+
+
+def check_valid_uuid(uuid_str: str) -> bool:
```
davidwaroquiers, line 246: Maybe add a unit test for this ? Even if it is used in other more complex test, if it at some point it fails to work, it would be easier to find out the root cause.

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: I think mypy allows this. At least on my side it works. There are some options indeed in mypy to be super strict about it, but mypy e.g. recognizes "if XX is None" checks for example and won't complain if you use it. I would expect the "custom_query = custom_query or {}" to be recognized also by mypy and not complain but maybe I'm wrong.

tests/db/cli/test_job.py
```diff
@@ -44,6 +44,30 @@ def test_jobs_list(job_controller, two_flows_four_jobs, patch_cli_consoles) -> N
...
+            '{"db_id": {"$in": ["1", "2"]}}',
+            "--count",
+            "-s",
+            "READY",
+        ],
+        required_out="Number of jobs: 1",
+    )
+    run_check_cli(
+        ["job", "list", "-fid", "1", "--count"], required_out="Number of jobs: 2"
+    )
```
davidwaroquiers, line 70: Maybe also add a test where you mix a db id and a uuid here ?

davidwaroquiers (commented):
Thanks a lot @FabiPi3 ! Great addition.
Just a few small comments or suggestions. I'll let @gpetretto approve.

src/jobflow_remote/cli/types.py
FabiPi3: I have copied this from other type definitions. Originally the db_id was only an integer, but at some point you allowed a prefix. Probably worth to update all of the documentation.

tests/db/cli/test_job.py
FabiPi3: I am not sure if you already have a test with uuid. I assume they are generated at runtime and vary from testrun to testrun. I am not sure how to best find them

src/jobflow_remote/jobs/jobcontroller.py
FabiPi3: Ah ok, not sure anymore. Now it works also for me. So I will change it.

src/jobflow_remote/utils/data.py
FabiPi3: Done.

tests/db/cli/test_job.py
davidwaroquiers: You can access the uuids from the two_flows_four_jobs fixture. If you do e.g. two_flows_four_jobs[0].uuid, you get the uuid of the first flow. You can get the uuid of all the jobs in this flow with two_flows_four_jobs[0].job_uuids. two_flow_four_jobs[0] and two_flow_four_jobs[1] are Flow objects.

davidwaroquiers: For info/ref, here is the fixture defined in tests/db/conftest.py:

```
@pytest.fixture()
def two_flows_four_jobs(random_project_name):
    """Add two flows with two jobs each to the DB"""
    from jobflow import Flow

    from jobflow_remote import submit_flow
    from jobflow_remote.testing import add

    add_first = add(1, 5)
    add_first.name = "add1"
    add_second = add(add_first.output, 5)
    add_second.name = "add2"

    add_first.update_metadata({"test_meta": 1})

    flow = Flow([add_first, add_second])
    flow.name = "f1"
    submit_flow(flow, worker="test_local_worker")

    add_third = add(1, 5)
    add_third.name = "add3"
    add_fourth = add(add_third.output, 5)
    add_fourth.name = "add4"

    flow2 = Flow([add_third, add_fourth])
    flow2.name = "f2"
    submit_flow(flow2, worker="test_local_worker")

    return [flow, flow2]
```

src/jobflow_remote/cli/types.py
davidwaroquiers: True, we should at some point make a check of the whole doc. I will resolve this now.

tests/db/cli/test_job.py
FabiPi3: Done.

gpetretto (approved):
Thanks!

PR #324, opened 2025-07-30
Dw/test cli colors

davidwaroquiers:
Some fix (to be checked?) of run_check_cli: the replace of spaces was putting a space between every character, not just replacing multiple spaces with only one.
Added possibility to check for colors using markup language on the text (separate argument from the "standard" required_out).
Added tests.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -11,7 +11,7 @@ import pytest
 from rich.console import Console
 
 
-@pytest.fixture(autouse=True)
+@pytest.fixture()
 def patch_cli_consoles(monkeypatch):
     import jobflow_remote.cli
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,3 +1,4 @@
+import inspect
 import logging
 import logging.config
 import random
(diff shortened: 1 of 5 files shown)


davidwaroquiers:
@gpetretto This should be ready to review. I reverted back the force_terminal in the code itself and added a fixture to monkeypatch the consoles with ones using force_terminal=True instead.

tests/conftest.py
```diff
@@ -7,6 +8,22 @@
 from pathlib import Path
 
 import pytest
+from rich.console import Console
+
+
+@pytest.fixture(autouse=True)
+def patch_cli_consoles(monkeypatch):
```
gpetretto, line 15: Why autouse this fixture instead of using it just for the test that needs it?

gpetretto (approved):
Looks good to me. I just left a comment. I let you decide if it is worth changing the way the fixture is use.

tests/conftest.py
davidwaroquiers: Good point. Just changed that. (note I messed up with my git history so I forced push my local branch)

davidwaroquiers:
Should be fine to merge once tests pass. @gpetretto do you want to check again this ?

Issue #326, opened 2025-07-31
Parallel batch job shows fewer `BATCH_RUNNING` jobs than expected after `jf admin reset`


Andrew-S-Rosen:
In trying to debug #323 by switching to atomate2, I came across some weird behavior. Here, I am running Jobflow-Remote such that it makes a parallel batch job requesting three nodes and running three 1-node VASP jobs concurrently. However, Jobflow-Remote reports that <3 jobs are running concurrently. I can confirm that three nodes are requested when I do `squeue`. Note that this simple example doesn't reproduce the error in #323 though.

```
┏━━━━━━━┳━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━┓
┃ DB id ┃ Name              ┃ State           ┃ Job id (Index)                            ┃ Worker     ┃ Last updated [EST] ┃
┡━━━━━━━╇━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━┩
│ 2     │ MatPES GGA static │ BATCH_RUNNING   │ d466692d-ec29-4562-8c7d-f5eca3d04ba6  (1) │ basic_vasp │ 2025-07-31 16:38   │
│ 1     │ MatPES GGA static │ BATCH_RUNNING   │ 4124bbee-b4d2-4845-a41a-c4db3aa5b6f3  (1) │ basic_vasp │ 2025-07-31 16:38   │
│ 9     │ MatPES GGA static │ BATCH_SUBMITTED │ 0c2b11e8-3503-4965-84ab-a896d3c59389  (1) │ basic_vasp │ 2025-07-31 16:37   │
│ 8     │ MatPES GGA static │ BATCH_SUBMITTED │ e467fe42-6a3e-4dad-a3ac-bc30112672b9  (1) │ basic_vasp │ 2025-07-31 16:37   │
[...]

PR #334, opened 2025-08-19
Fix batch worker

gpetretto:
Following the discussion in #326, adding the cleanup of the batch managerment folder for batch workers.

Two small bugs also fixed:
* Wrong counting of the number of required jobs if multiple batch workers present in the same project
* Files for running jobs remained hanging in the folder if the job in the (e.g. slurm) queue is cancelled/killed.

Extended the tests for the batch worker.

### DIFF
--- a/src/jobflow_remote/jobs/batch.py
+++ b/src/jobflow_remote/jobs/batch.py
@@ -145,6 +145,37 @@ class RemoteBatchManager:
         for job_id, index, process_uuid in ids:
             self.host.remove(self.terminated_dir / f"{job_id}_{index}_{process_uuid}")
 
+    def delete_running(self, process_id: str) -> None:
+        """
+        Remove job files from the running folder for a specific process uuid.
+
+        Should be used only for jobs that failed and left dangling running files.
+
+        Parameters
+        ----------
+        process_id
+            The uuid of the process for the running files to be removes
+        """
+        if not self._dir_initialized:
+            self._init_files_dir()
+        running_files = self.host.listdir(self.running_dir)
+        for filename in running_files:
+            if filename.endswith(process_id):
+                self.host.remove(self.running_dir / filename)
+
+    def cleanup(self) -> bool:
+        """
+        Remove the files directory on the host.
+
+        Returns
+        -------
+        bool
+            True if the directory was successfully deleted or was not existing.
+        """
+        if self.host.exists(self.files_dir):
+            return self.host.rmtree(self.files_dir, raise_on_error=False)
+        return True
+
 
 class LocalBatchManager:
     """
--- a/src/jobflow_remote/jobs/batch.py
+++ b/src/jobflow_remote/jobs/batch.py
@@ -154,7 +154,7 @@ class RemoteBatchManager:
         Parameters
         ----------
         process_id
-            The uuid of the process for the running files to be removes
+            The uuid of the process for the running files to be removed.
         """
         if not self._dir_initialized:
             self._init_files_dir()
--- a/src/jobflow_remote/remote/host/base.py
+++ b/src/jobflow_remote/remote/host/base.py
@@ -161,6 +161,21 @@ class BaseHost(MSONable):
         """
         raise NotImplementedError
 
+    @abc.abstractmethod
+    def exists(self, path: str | Path) -> bool:
+        """
+        Test whether a path exists.
+
+        Parameters
+        ----------
+        path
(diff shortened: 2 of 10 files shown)


tests/conftest.py
```diff
@@ -327,3 +327,39 @@ def wait_daemon_shutdown():
...
+    from jobflow_remote.config.base import Project
+
+    d = loadfn(project_file_path)
+
+    if dict_mods:
+        apply_mod(update, d)
+    else:
+        d = d | update
+
+    # validate the Project to get an error, otherwise if the values
```
davidwaroquiers: I did not understand the comment. I guess it is validate the Project to prevent having an error at run time that may be hidden by the runner/daemon starting. Right ? Here initially it seemed: let's validate and get an error.

src/jobflow_remote/remote/host/base.py
```diff
@@ -161,6 +161,21 @@ def shell(self, pre_cmd: str | None = None, shell: str = "bash"):
...
 
+    @abc.abstractmethod
+    def exists(self, path: str | Path) -> bool:
+        """
+        Test whether a path exists.
+
+        Parameters
+        ----------
+        path
+            The path to the directory tree to be removed.
```
davidwaroquiers:
```suggestion
            The path to check whether it exists.
```

src/jobflow_remote/remote/host/local.py
```diff
@@ -175,3 +175,18 @@ def shell(self, pre_cmd: str | None = None, shell: str = "bash"):
...
         Context().run(cmd, pty=True)
+
+    def exists(self, path: str | Path) -> bool:
+        """
+        Test whether a path exists.
+
+        Parameters
+        ----------
+        path
+            The path to the directory tree to be removed.
```
davidwaroquiers:
```suggestion
            The path to check whether it exists.
```

src/jobflow_remote/remote/host/remote.py
```diff
@@ -446,6 +446,30 @@ def shell(self, pre_cmd: str | None = None, shell: str = "bash"):
...
         self.connection.run(cmd, pty=True)
 
+    def exists(self, path: str | Path) -> bool:
+        """
+        Test whether a path exists.
+
+        Parameters
+        ----------
+        path
+            The path to the directory tree to be removed.
```
davidwaroquiers:
```suggestion
            The path to check whether it exists.
```

src/jobflow_remote/jobs/batch.py
```diff
@@ -145,6 +145,37 @@ def delete_terminated(self, ids: list[tuple[str, int, str]]) -> None:
...
+    def delete_running(self, process_id: str) -> None:
+        """
+        Remove job files from the running folder for a specific process uuid.
+
+        Should be used only for jobs that failed and left dangling running files.
+
+        Parameters
+        ----------
+        process_id
+            The uuid of the process for the running files to be removes
```
davidwaroquiers:
```suggestion
            The uuid of the process for the running files to be removed
```

tests/integration/test_advanced_options.py
```diff
@@ -78,6 +79,165 @@ def test_run_batch_multi(job_controller, monkeypatch) -> None:
...
+    from jobflow_remote.remote.queue import QueueManager
+    from jobflow_remote.testing import add_sleep
+
+    def submit_jobs(n: int, sleep: int):
+        job_ids = []
+        for _ in range(n):
+            add_j = add_sleep(2, sleep)
+
+            flow = Flow([add_j])
+            submit_flow(flow, worker=worker_name)
```
davidwaroquiers: submit_flow can take a Job as an argument, any reason why you create the Flow here ?

davidwaroquiers (approved):
Great!
Spent some time to better understand how the batch execution is implemented. A few things may use some more explanations but it's great! Maybe at some point it would be good to add some explanation on the logic in the documentation for dev. I would also maybe change a few of the internal variables, I sometimes had to think or move around the code to know what an id was refering to (batch id, job id, queue/slurm id, ...). Anyway this would be for another PR if we do this.
A few minor comments then you can merge.

davidwaroquiers (approved):
All great!

gpetretto:
Addressed the comments, plus improved one docstring following https://github.com/Matgenix/jobflow-remote/issues/155#issuecomment-3198330749

PR #339, opened 2025-08-27
CLI plugins

gpetretto:
New mechanism for extending CLI from external packages based on plugins.

The idea is that a package can define a section in the `pyproject.toml` like:
```toml
[project.entry-points."jobflow_remote.plugins"]
backup = "path.to.some.local.module"
```

and in the module define a `setup_jf_plugin` function that will be automatically discovered and called by jobflow remote. The function should thus import some of the typer app object from jobflow-remote and add sub-apps or commands. 
For example:
```python
def setup_jf_plugin():
    from jobflow_remote.cli.job import app_job
    @app_job.command()
    def example():
        pass
```
Will add a `jf job example` to the standard CLI.

One note about the `load_plugins` and `list_plugins` functions. The logic is overlapping and they could probably be merged in a single function. For the moment I kept them separated because the `load_plugins` function will be executed at every call of the `jf` command and I would thus like to keep it as minimal as possible, while merging them will likely require a series of additional checks. I doubt this will have a real impact on the perceived time, but I thought of trying to minimize the impact

### TODO

If the approach is fine I will add some documentation.

### DIFF
--- a/src/jobflow_remote/cli/plugin.py
+++ b/src/jobflow_remote/cli/plugin.py
@@ -18,7 +18,7 @@ PLUGIN_LOAD_FUNCTION = "setup_jf_plugin"
 
 app_plugin = JFRTyper(
     name="plugin",
-    help="Commands for handling backup of the database",
+    help="Commands for managing plugins",
     no_args_is_help=True,
 )
 app.add_typer(app_plugin)
--- a/src/jobflow_remote/cli/plugin.py
+++ b/src/jobflow_remote/cli/plugin.py
@@ -1,0 +1,101 @@
+import importlib
+import logging
+import traceback
+from typing import Annotated
+
+import typer
+from rich.padding import Padding
+
+from jobflow_remote.cli.jf import app
+from jobflow_remote.cli.jfr_typer import JFRTyper
+from jobflow_remote.cli.utils import exit_with_warning_msg, out_console
+
+logger = logging.getLogger(__name__)
+
+
+PLUGIN_GROUP = "jobflow_remote.plugins"
+PLUGIN_LOAD_FUNCTION = "setup_jf_plugin"
+
+app_plugin = JFRTyper(
+    name="plugin",
+    help="Commands for handling backup of the database",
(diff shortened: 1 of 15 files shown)


src/jobflow_remote/utils/log.py
```diff
@@ -78,7 +80,16 @@ def initialize_cli_logger(
...
         The log level.
+    full_exc_info
+        If True the full stack trace will be printed in the log message
+        in case of error.
+    loggers
+        Additional loggers to be considered. jobflow_remote is always included
     """
+
+    loggers = loggers or []
+    loggers.insert(0, "jobflow_remote")
```
davidwaroquiers, line 91: Not sure if it's relevant but 1/ what happens if jobflow_remote is also in the list loggers passed ? and 2/ why does it need to be at the beginning of the list ?

src/jobflow_remote/cli/plugin.py
```diff
@@ -0,0 +1,101 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+PLUGIN_GROUP = "jobflow_remote.plugins"
+PLUGIN_LOAD_FUNCTION = "setup_jf_plugin"
+
+app_plugin = JFRTyper(
+    name="plugin",
+    help="Commands for handling backup of the database",
```
davidwaroquiers:
```suggestion
    help="Commands for managing plugins",
```

davidwaroquiers (approved):
Great work! Just two small comments.

src/jobflow_remote/utils/log.py
gpetretto: 1) In principle nothing, because then this list is used as keys of a dictionary, where the values are all the same, so python will already remove one of the two entries.
2) I suppose there is no real good reason. It is just that in the original verions it was in the first place and I kept it like this. I think I can also switch to `append`. Should I do that?

davidwaroquiers: No, was just wondering. I let you decide anyway.

PR #346, opened 2025-09-12
Fixture for run_check_cli that includes patching the consoles

davidwaroquiers:
Following 1 hour of debugging because I forgot I had to use the patch consoles fixture when testing colors, I decided to have that included directly.


gpetretto (approved):
Fine with me. Thanks

PR #347, opened 2025-09-12
Small fixes and tests improvements

gpetretto:
A set of small updates:
* fix `updated_on` upon job completion
* fix docstrings and error messages
* Attempt to cancel running processes when deleting a Flow. Will be done automatically by default
* New option when running integration tests `--worker-types`. Allows to select the types of worker for integration tests. This makes easier to run them only on some containers, since I have problems running on SLURM, PBS and SGE with the same docker options.

### DIFF
--- a/src/jobflow_remote/jobs/jobcontroller.py
+++ b/src/jobflow_remote/jobs/jobcontroller.py
@@ -840,7 +840,7 @@ class JobController:
         if max_limit != 0 and len(queried_dbs_ids) > max_limit:
             raise ValueError(
                 f"Cannot perform {action_description} on {len(queried_dbs_ids)} Jobs "
-                f"as they exceeds the specified maximum limit ({max_limit})."
+                f"as they exceed the specified maximum limit ({max_limit})."
                 f"Increase the limit to complete the action on this many Jobs."
             )
 
@@ -1735,7 +1735,6 @@ class JobController:
         """
         Stop selected Jobs. Only Jobs in the READY and all the running states
         can be stopped.
-        The action is not reversible.
 
         Parameters
         ----------
@@ -1812,7 +1811,6 @@ class JobController:
         can be stopped.
         Selected by db_id or uuid+index. Only one among db_id
         and job_id should be defined.
-        The action is not reversible.
 
         Parameters
         ----------
@@ -2626,6 +2624,7 @@ class JobController:
         max_limit: int = 10,
         delete_output: bool = False,
         delete_files: bool = False,
+        cancel_processes: bool = True,
     ) -> int:
         """
         Delete a list of Flows based on the flow uuids.
@@ -2641,6 +2640,9 @@ class JobController:
             If True also delete the associated output in the JobStore.
         delete_files
             If True also delete the files on the worker.
+        cancel_processes
+            If True will attempt to delete the processes for SUBMITTED and RUNNING jobs.
+            Failure to cancel will not stop the deletion of the Flow.
 
         Returns
         -------
@@ -2655,7 +2657,7 @@ class JobController:
 
         if max_limit != 0 and len(flow_ids) > max_limit:
             raise ValueError(
-                f"Cannot delete {len(flow_ids)} Flows as they exceeds the specified maximum "
+                f"Cannot delete {len(flow_ids)} Flows as they exceed the specified maximum "
                 f"limit ({max_limit}). Increase the limit to delete the Flows."
             )
         deleted = 0
@@ -2664,7 +2666,10 @@ class JobController:
             for fid in flow_ids:
                 # TODO should it catch errors?
                 if self.delete_flow(
-                    fid, delete_output=delete_output, delete_files=delete_files
+                    fid,
+                    delete_output=delete_output,
+                    delete_files=delete_files,
+                    cancel_processes=cancel_processes,
                 ):
                     deleted += 1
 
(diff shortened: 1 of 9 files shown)


src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -2703,9 +2712,22 @@ def delete_flow(
             if jobstore_name := flow.get("jobstore"):
                 jobstore = self.optional_jobstores[jobstore_name]
             jobstore.remove_docs({"uuid": {"$in": job_ids}})
-        if delete_files:
+        if delete_files or cancel_processes:
             jobs_info = self.get_jobs_info(flow_ids=[flow_id])
-            self._safe_delete_files(jobs_info)
+            if cancel_processes:
+                for ji in jobs_info:
+                    if ji.state in [JobState.SUBMITTED, JobState.RUNNING]:
```
davidwaroquiers, line 2719: Would there be any way to cancel batch submitted/running jobs as well?

tests/integration/conftest.py
```diff
@@ -15,6 +15,224 @@
...
+    parser.addoption(
+        "--worker-types",
+        "--wt",
+        dest="worker_types",
+        nargs="+",
+        help="List of workers to be used for the integration tests. All available if not specified.",
+        default=None,
+        action="store",
+        choices=WORKER_TYPES,
+    )
```
davidwaroquiers, line 39: Just wondering if this "should" be in the main conftest.py. It can stay here (and is related to the integration tests of course), but maybe it's easier/best to keep all additional cmd options added trhough pytest_addoption in the same place ? No idea what's the common/best practices. What do you think ?

tests/db/jobs/test_jobcontroller.py
```diff
@@ -77,11 +77,11 @@ def test_queries(job_controller, runner) -> None:
...
     assert job_controller.count_jobs(job_ids=(add_first.uuid, 1)) == 1
     jobs_start_date = job_controller.get_jobs_info(start_date=date_create)
-    assert len(jobs_start_date) == 1
-    assert jobs_start_date[0].uuid == add_first.uuid
+    assert len(jobs_start_date) == 2
+    assert {j.uuid for j in jobs_start_date} == {add_first.uuid, add_second.uuid}
 
     jobs_end_date = job_controller.get_jobs_info(end_date=date_create)
-    assert jobs_end_date[0].uuid == add_second.uuid
+    assert {j.uuid for j in jobs_end_date} == {add_third.uuid, add_fourth.uuid}
```
davidwaroquiers, line 84: Not sure why this changed. There should only be one that started with the run_one_job no ?

davidwaroquiers: Ok I went to the docstring and got it now :) I thought it was "give me the jobs that have started after start_date (i.e. that are in RUNNING or more ? actually it's not even really possible or at least no so easily right ?).

Maybe we should consider renaming ? (not sure ...) like "updated_after" and "updated_before" for example ? There could also be a "created_after" and "created_before".

I propose we keep it like that for now anyway. What do you think ?

davidwaroquiers (approved):
All good to me. Just a few comments. I let you decide or if you want to have a call, we can also.

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: At the moment not. The fact is that it should not delete the queue (e.g. SLURM) process, but rather stop a job inside the batch job. Likely this would need a method like `cancel_process()` in `RemoteBatchManager`. And this should also be added in all the other methods that cancel jobs. Should I implement it?

tests/integration/conftest.py
gpetretto: The reasons I have put it here are:
* to have a single `WORKER_TYPES`
* to keep everything related to integration tests in the same `conftest.py`

None of these are strictly necessary. I don't know if there are best practices for this. At least it works, so pytest explicitly supports the option to add options in different `conftest.py`. 
In the end both options should be fine.

tests/db/jobs/test_jobcontroller.py
gpetretto: I understand the point. Indeed maybe it would have been better a different name for these variables. However I would avoid changing the API if not really necessary.

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: I don't know, it might be useful but at the same time, it might be difficult. Could we encounter an issue similar to the custodian stuff for which jobs are not cancelled ? Maybe we can keep it like this for now ? What do you think ?

tests/integration/conftest.py
davidwaroquiers: ok let's keep it here

tests/db/jobs/test_jobcontroller.py
davidwaroquiers: sure, fine with me, let's keep like this then.

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: Indeed is not obvious that if the subprocess is stopped in the main node it will kill the `srun` in other nodes. It will likely require some further investigations. Also, we may decide to change the how parallel batch jobs are handles. 
I agree leaving it like this and maybe address this in a specific PR for batch jobs

PR #354, opened 2025-09-19
Updates to batch process

davidwaroquiers:
- Some refactoring (splitting up the big update_batch method)
- Renaming of variables
- Added possibility to have a batch collection to keep info about old batches (optional)


davidwaroquiers:
I've added one dependency (bidict) but I think I would remove it and deal with the "inverse" dict directly here

src/jobflow_remote/cli/batch.py
```diff
@@ -30,6 +34,8 @@ def processes_list(
             help="Select the worker.",
         ),
     ] = None,
+    show_all: show_all_batches_opt = False,
+    max_batches: max_batches_per_worker_opt = 10,
```
gpetretto: seems this option is never used. If to be implemented in the future I would suggest to comment it out for the moment.

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -129,6 +131,8 @@ def __init__(
         auxiliary_collection
             The name of the collection used to store other auxiliary data.
             Uses the DB defined in the queue_store.
+        batches_collection
+            The name of the collection used to store the batch processes.
```
gpetretto, line 138: Maybe also here it can be written that it is optional?

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -4755,12 +4786,35 @@ def add_batch_process(
...
         """
+        if self.batches is not None:
+            self.batches.insert_one(
+                {
+                    "batch_uid": batch_uid,
+                    "process_id": process_id,
+                    "batch_state": BatchState.SUBMITTED.value,
+                    "worker": worker,
+                }
+            )
```
gpetretto, line 4807: It is probably better like this, given the simple format of the document, but I was wondering if you considered creating a `BatchDoc` pydantic model to insert/retrieve the data. 

Below the `jobs` value is set in this dictionary. Maybe it can already be added a `"jobs": {}`? Or is it not worth?

Also, maybe it would be useful to add more information?
For example also here an `updated_on`? And also maybe `start_time` and `end_time`? So that one can understand if maybe a batch jobs are being submitted but without doing anything?

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -4755,12 +4786,35 @@ def add_batch_process(
...
+            {"$set": {f"batch_processes.{worker}.{process_id}": batch_uid}},
             upsert=True,
         )
 
+    def add_job_to_batch(self, job_id, job_index, batch_uid, worker, info):
+        if self.batches is not None:
+            self.batches.update_one(
+                {"batch_uid": batch_uid, "worker": worker},
+                {"$set": {f"jobs.{job_id}.{job_index}": info}},
+            )
```
gpetretto, line 4827: Should we consider a safeguard with respect to the maximum 16MB document size in case a lot of jobs are added? (not sure how)

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -4755,12 +4786,35 @@ def add_batch_process(
...
+                }
+            )
         return self.auxiliary.find_one_and_update(
             {"batch_processes": {"$exists": True}},
-            {"$set": {f"batch_processes.{worker}.{process_id}": process_uuid}},
+            {"$set": {f"batch_processes.{worker}.{process_id}": batch_uid}},
             upsert=True,
         )
 
+    def add_job_to_batch(self, job_id, job_index, batch_uid, worker, info):
```
gpetretto: missing typing.

Also the method name is confusing: I have seent that it is called multiple times and it seems it is done to update the current state of the Job, but initially I was confused by the fact that the same job was "added" multiple times.

src/jobflow_remote/cli/batch.py
```diff
@@ -39,29 +45,77 @@ def processes_list(
...
+                    )
+            batch_processes = [
+                {
+                    "batch_uid": batch_uid,
+                    "process_id": process_id,
+                    "worker": worker,
+                }
+                for worker, worker_batches in batch_processes.items()
+                for process_id, batch_uid in worker_batches.items()
+            ]
```
gpetretto, line 73: The fact that here `batch_processes` gets reassigned with a different data structure is a bit confusing. Maybe the original value can be called something like `batch_processes_data` so that it is clear that they are two different structures?

src/jobflow_remote/jobs/runner.py
```diff
@@ -1118,221 +1122,343 @@ def update_batch_jobs(self) -> None:
...
+                                bidict(), batch_uid, worker_name, worker
+                            ),
+                        }
+                    }
+                    self.job_controller.add_job_to_batch(
+                        job_id=job_id,
+                        job_index=job_index,
+                        batch_uid=batch_uid,
+                        worker=worker_name,
+                        info={"state": next_state.value},
```
gpetretto: it seems this is the last time the info is modified. Maybe it can always be TERMINATED? For a generic user it will not make much sense that in some cases it will be DOWNLOADED, in other TERMINATED, but it will never move further from that.

src/jobflow_remote/cli/batch.py
```diff
@@ -39,29 +45,77 @@ def processes_list(
...
+
+        batches_jobs = []
+        if verbosity > 0:
+            for batch in batch_processes:
+                batch_jobs = [
+                    (jid, str(jidx))
+                    for jid, jid_dict in batch.get("jobs", {}).items()
+                    for jidx in jid_dict
+                ]
+                batches_jobs.append(batch_jobs)
```
gpetretto, line 115: It seems that this will not only include the running jobs, but all the jobs that were also terminated. Am I right? But in the table they are still indicated as "Running Job ids".

src/jobflow_remote/cli/formatting.py
```diff
@@ -586,39 +586,37 @@ def format_upgrade_actions(actions: list[UpgradeAction]):
...
+    for ibatch, batch_data in enumerate(batch_processes):
+        worker = workers[batch_data["worker"]]
+        row = [
+            batch_data["process_id"],
+            batch_data["batch_uid"],
+            batch_data["worker"],
+            get_job_path(batch_data["batch_uid"], None, worker.batch.work_dir),
+        ]
+        if status:
+            row.append(batch_data.get("batch_state", "UNKNOWN"))
```
gpetretto: Does "UNKNOWN" get printed for all the jobs if `show_all` is set to False? In that case I would suggest to set `status` to False as well. The "UNKNOWN" gives the feeling that there may be something wrong ("why is it unknown?"). While it is just that there is no way of knowing for the code and they will always be all unknown in that case.

src/jobflow_remote/jobs/batch.py
```diff
@@ -53,6 +55,7 @@ def __init__(
         self.running_dir = self.files_dir / RUNNING_DIR
         self.terminated_dir = self.files_dir / TERMINATED_DIR
         self.lock_dir = self.files_dir / LOCK_DIR
+        self.batches_history_dir = self.files_dir / BATCHES_HISTORY_DIR
```
gpetretto: Is this used somewhere?

src/jobflow_remote/jobs/runner.py
```diff
@@ -1118,221 +1122,343 @@ def update_batch_jobs(self) -> None:
...
+        return running_jobs
+
+    def _cache_batch(self, worker_name, worker, batch_uid, process_id):
+        if worker_name not in self._cached_batches:
+            self._cached_batches[worker_name] = OrderedDict()
+        self._cached_batches[worker_name][batch_uid] = process_id
+        while len(self._cached_batches[worker_name]) > 2 * worker.max_jobs + 5:
+            self._cached_batches[worker_name].popitem(last=False)
+
+    def get_process_id(self, batch_processes, batch_uid, worker_name, worker):
```
gpetretto: maybe it can be renamed `batch_get_process_id` or `get_batch_process_id`?

missing typing

src/jobflow_remote/jobs/runner.py
```diff
@@ -1118,221 +1122,343 @@ def update_batch_jobs(self) -> None:
...
-                    work_dir=remote_path,
-                    create_submit_dir=True,
-                )
+            return doc["process_id"]
+        # What do we do when we don't have it ? (I'm pretty sure it cannot happen if we have the batches collection,
+        # but if we don't, it could happen, although relatively "unlikely")
+        return "UNKNOWN"
+
+    def batch_update_status(
+        self, batch_manager, queue_manager, worker_name, worker, running_jobs
```
gpetretto: missing typing

src/jobflow_remote/jobs/runner.py
```diff
@@ -1118,221 +1122,343 @@ def update_batch_jobs(self) -> None:
...
+                                # The above loop will thus proceed normally
+                                # This is done to automatically pass down the error message
+                                raise RemoteError(err_msg, no_retry=True)
+                # Also remove the corresponding files from the running folder of batch manager
+                batch_manager.delete_running(batch_processes_data[pid])
+
+            return list(running_processes)
+        return []
+
+    def submit_batch_processes(self, queue_manager, worker_name, worker, processes):
```
gpetretto: missing typing

src/jobflow_remote/jobs/runner.py
```diff
@@ -1118,221 +1122,343 @@ def update_batch_jobs(self) -> None:
...
-                        else:
-                            next_state = JobState.DOWNLOADED
-                        set_output = {
-                            "$set": {
-                                "state": next_state.value,
-                                "remote.process_id": process_running_uuid,
-                            }
+
+            for pid in running_processes:
+                self.job_controller.set_running_batch_process(pid, worker_name)
```
gpetretto: This keeps updating the state of the state of the running process for all the jobs, for all their duration in the DB, right? I suppose it is not very intensive action, but they are still a lot of calls to the DB for nothing.

src/jobflow_remote/jobs/runner.py
```diff
@@ -1118,221 +1122,343 @@ def update_batch_jobs(self) -> None:
...
-                    pre_run=worker.pre_run,
-                    post_run=worker.post_run,
-                    options=resources,
-                    work_dir=remote_path,
-                    create_submit_dir=True,
-                )
+            return doc["process_id"]
+        # What do we do when we don't have it ? (I'm pretty sure it cannot happen if we have the batches collection,
+        # but if we don't, it could happen, although relatively "unlikely")
+        return "UNKNOWN"
```
gpetretto: Is there a point in returning "UNKNOWN"? I think dealing with hardcoded string values is often a pain. Isn't there any better solution here? Would it be better to raise, since it should never happen?
Or what about returning `None`?

src/jobflow_remote/jobs/runner.py
```diff
@@ -1118,221 +1122,343 @@ def update_batch_jobs(self) -> None:
...
-                                    err_msg = f"The batch process that was running the job (process_id: {pid}, uuid: {process_running_uuid} was likely killed before terminating the job execution"
-                                    raise RemoteError(err_msg, no_retry=True)
-                    # Also remove the corresponding files from the running folder of batch manager
-                    batch_manager.delete_running(batch_processes_data[pid])
-
-                processes = list(running_processes)
+            # Check the batch processes that should be running on the remote queue
+            # and update the state in the DB if something changed (both for batch processes and
+            # for the states of the jobs that were running in these batch processes)
+            processes = self.batch_update_status(
```
gpetretto: Maybe the `processes` names can be more explicit? Like `running_processes`?

src/jobflow_remote/jobs/runner.py
```diff
@@ -1118,221 +1122,343 @@ def update_batch_jobs(self) -> None:
...
+            )
+
+            if submit_result.status == SubmissionStatus.FAILED:
+                logger.error(f"submission failed. {submit_result!r}")
+            elif submit_result.status == SubmissionStatus.JOB_ID_UNKNOWN:
+                logger.error(
+                    f"submission succeeded but ID not known. Job may be running but status cannot be checked. {submit_result!r}"
+                )
+                # TODO: should there be anything here to check with the uuid somehow on the worker ?
+                #  More specifically, maybe we could just add the batch process to the batches collection with None as an ID ?
```
gpetretto, line 1477: quick comment on this point: if you add the entry with None as process_id, what if there are more than one? How do you tell them apart?

src/jobflow_remote/remote/queue.py
```diff
@@ -224,7 +224,11 @@ def get_jobs_list(
         # list of job ids, if the user is passed ignore the job ids.
         if user is not None:
             jobs = None
-        job_cmd = self.scheduler_io.get_jobs_list_cmd(jobs, user)
+        # Old versions of qtoolkit just returned the command.
+        # New versions of qtoolkit return the command and the list of stringified job_ids.
+        # This ensures backward compatibility.
+        result = self.scheduler_io.get_jobs_list_cmd(jobs, user)
+        job_cmd = result[0] if isinstance(result, tuple) else result
```
gpetretto: Can we first settle if it is really needed to change the API of QTK?

tests/conftest.py
```diff
@@ -538,15 +543,28 @@ def update_project_data(
...
+            updates,
+            dict_mods=True,
+            project_file_path=Path(current_project_data.filepath),
+        )
+        yield
+    else:
+        # case where patch_project is not parametrized but just used inside the test
+        yield partial(
+            update_project_data, project_file_path=Path(current_project_data.filepath)
+        )
```
gpetretto: Maybe it would be worth making two different fixtures for the two different cases? One might be tempted to parametrize **and** also try to call `patch_project` to make some further tests during the test execution. Since in this case it does not yield nothing, this will result in an error.

Or maybe always yield `partial(update_project_data, project_file_path=Path(current_project_data.filepath))`?

tests/integration/test_advanced_options.py
```diff
@@ -29,21 +41,50 @@ def test_run_batch(job_controller, monkeypatch, clean_slurm_queue) -> None:
...
+        insert_batch_process_spy = mocker.spy(
+            runner.job_controller.batches, "insert_one"
+        )
+
     # set this so it will be called
-    monkeypatch.setattr(runner.runner_options, "delay_update_batch", 1)
+    monkeypatch.setattr(runner.runner_options, "delay_update_batch", 0.05)
+    monkeypatch.setattr(runner.runner_options, "delay_advance_status", 0.05)
+    monkeypatch.setattr(runner.runner_options, "delay_check_run_status", 0.05)
+    monkeypatch.setattr(runner.runner_options, "delay_checkout", 0.05)
```
gpetretto, line 54: Should we consider doing this generally during the tests? Does it really speed things up with such small timings?

tests/integration/test_advanced_options.py
```diff
@@ -29,21 +41,50 @@ def test_run_batch(job_controller, monkeypatch, clean_slurm_queue) -> None:
...
+    # make sure the process id is not a uuid (we could check that it is a "valid" slurm id also)
+    assert not check_valid_uuid(slurm_job_id, raise_on_error=False)
+
+    if request.node.callspec.id == "with_batches_collection":
+        assert runner.job_controller.batches is not None
+        assert insert_batch_process_spy.call_count == 1
+    elif request.node.callspec.id == "without_batches_collection":
+        assert runner.job_controller.batches is None
+    else:
+        pytest.fail("Should not be here!")
```
gpetretto: The message here maybe can be a bit informative? Like "Wrong parametrization id: {request.node.callspec.id}"?

tests/integration/test_advanced_options.py
```diff
@@ -29,21 +41,50 @@ def test_run_batch(job_controller, monkeypatch, clean_slurm_queue) -> None:
...
+    assert runner.job_controller.count_jobs(states=JobState.COMPLETED) == 6
 
     # verify that only one job was executed at the time. start_time of a job
     # is after the end_time of the one preceding it.
     # This should test that the batch runner is not running with multiple
     # parallel processes
-    jobs_info = job_controller.get_jobs_info()
+    jobs_info = runner.job_controller.get_jobs_info()
```
gpetretto: I am curious, why is this needed?

tests/integration/test_advanced_options.py
```diff
@@ -186,7 +280,23 @@ def submit_jobs(n: int, sleep: int):
...
         raise RuntimeError("The Jobs were not set to REMOTE_ERROR state")
 
+    run_check_cli(["batch", "list"], required_out="No batch processes running")
+    if request.node.callspec.id == "with_batches_collection":
+        run_check_cli(
+            ["batch", "list", "-a"],
+            required_out=["Batches info", "Status"],
+        )
+
+    assert len(batch_manager.get_running()) == 0
```
gpetretto, line 296: there is `get_running` twice

src/jobflow_remote/jobs/runner.py
```diff
@@ -1118,221 +1122,343 @@ def update_batch_jobs(self) -> None:
...
+        while len(self._cached_batches[worker_name]) > 2 * worker.max_jobs + 5:
+            self._cached_batches[worker_name].popitem(last=False)
+
+    def get_process_id(self, batch_processes, batch_uid, worker_name, worker):
+        # First, try to take from the cached batches info
+        if worker_batches := self._cached_batches.get(worker_name):  # noqa: SIM102
+            if batch_uid in worker_batches:
+                return worker_batches[batch_uid]
+        # Then try to take it from the running batch processes
+        process_id = batch_processes.inverse.get(batch_uid)
```
gpetretto: if the fact of being a bidict is used only here, maybe indeed the dependecy on bidict can be removed and just invert the dictionary on the fly if needed?
In general it seems that this bidict would be useful if the same dictionary needs to be updated and keep track of changes in both direction, but if I am not wrong here it is just instanciated and used as it is.

gpetretto (commented):
Thanks for the updates.
I have left some comments.
A general one is that it seems that the batches collection could in principle replace entirely the auxiliary collection, am I right? Should we consider switching to that, instead of keeping duplicated information in the DB? Also, if someone is interested in using the batch, why would they not be interested in having all the possible information? Assuming that this is not too big.
This would be of course a backward incompatible change, so it might be a bit annoying to handle.

src/jobflow_remote/cli/batch.py
davidwaroquiers: Now it's done

src/jobflow_remote/cli/batch.py
davidwaroquiers: Done

src/jobflow_remote/cli/batch.py
davidwaroquiers: Correct, I changed

src/jobflow_remote/cli/formatting.py
davidwaroquiers: Here it's the batch_state. If show_all is set to False, then it's just the jobflow jobs that are not shown, but the batch state can still be there. But I agree that, as a user, you can wonder why it's unknown. I don't know what to do here. Maybe put nothing if the batch_state is unknown ? row.append(batch_data.get("batch_state", ""))

src/jobflow_remote/jobs/batch.py
davidwaroquiers: No, leftover. I removed everything related

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: Done

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: For the BatchDoc pydantic model, I indeed wondered if it was worth it. What do you think ?
I've set jobs as a dict from the beginning.
For the additional information, I've added updated_on, start_time and end_time. For start and end though I am not sure. This is the time at which the runner "finds out" that the batch process started or finished. It might be delayed (or even not found out in the case of start_time, if it directly goes to finished state).

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: Added typing and changed name to "update_job_in_batch"

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: Maybe just a try except that only does the update of the updated_on in case adding the job makes it go beyond the 16 Mb ? I would guess it would be an edge case no ? (would really need a lot of jobs, except if we start to add a lot of info ..)

src/jobflow_remote/jobs/runner.py
davidwaroquiers: Done

src/jobflow_remote/jobs/runner.py
davidwaroquiers: Done

src/jobflow_remote/jobs/runner.py
davidwaroquiers: I agree. Removed.

src/jobflow_remote/jobs/runner.py
davidwaroquiers: Yes, I just did not know what to do at the time, hence the comment :) I've put a RuntimeError. What do you think ?

src/jobflow_remote/jobs/runner.py
davidwaroquiers: Added

src/jobflow_remote/jobs/runner.py
davidwaroquiers: True, and also the start_time and updated_on time is now updated every time ...
I have updated the set_running_batch_process to set it only when start_time is None.
At least it's correct now.
Still have to think about how to not do the db calls everytime (maybe caching also in the _cached_batches somehow ?)

src/jobflow_remote/jobs/runner.py
davidwaroquiers: Added

src/jobflow_remote/jobs/runner.py
davidwaroquiers: Fully agree ... no clue. Maybe from the name of the job ? Maybe there could be a workaround to get something from the remote worker in the batch handle directory ?

src/jobflow_remote/jobs/runner.py
davidwaroquiers: Should we update the batch documents with the real job states when they go beyond TERMINATED ? That would be an additional call to the database. Not sure if it's worth ... maybe manually if something goes wrong ?

src/jobflow_remote/remote/queue.py
davidwaroquiers: I will go back to qtk to see about that. I agree

tests/conftest.py
davidwaroquiers: I did it with your second proposal (always yield the partial)

tests/integration/test_advanced_options.py
davidwaroquiers: It does speed up the tests indeed. I've also made the changes to the config base so that all timings can be expressed in floats. I guess we can see in another PR to speed up tests that are long

tests/integration/test_advanced_options.py
davidwaroquiers: Done

tests/integration/test_advanced_options.py
davidwaroquiers: This was before I modified the patch_project to be applied directly before the job_controller fixture. Now it can be removed.

tests/integration/test_advanced_options.py
davidwaroquiers: Removed

davidwaroquiers (commented):
I've made some updates. You can have a look and resolve conversations that are now fixed and we can discuss the remaining ones.

src/jobflow_remote/cli/batch.py
gpetretto: if these are not just running jobs anymore maybe the name of the argument of `get_batch_processes_table` can also be changed, instead of `running_jobs`?

src/jobflow_remote/cli/batch.py
```diff
@@ -39,29 +46,94 @@ def processes_list(
...
+            batch_state=[BatchState.SUBMITTED, BatchState.RUNNING]
+        )
+
+        finished_batch_processes = jc.get_all_batches(
+            batch_state=BatchState.FINISHED,
+            max_batches_per_worker=max_batches,
+            sort={"end_time": -1},
+        )
+
+        batch_processes.extend(finished_batch_processes)
```
gpetretto: It seems a bit weird to ask for maximum 10 jobs and still get more than 10. Why this?
Wouldn't it be easier to add the option to filter based on `BatchState` instead, so the user can decide?

src/jobflow_remote/cli/formatting.py
gpetretto: dosn't `show_all` make the difference between fetching from the auxiliary collection and from the batches one? So it seems that if show_all is False the batch state con only be `UNKNOWN`? 
An empty box will be better, but if the auxiliary is dropped this case will not happen anymore, right?

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -4731,9 +4749,45 @@ def get_batch_processes(
...
+
+        query: dict = {}
+        if worker:
+            query["worker"] = worker
+        if batch_state:
+            if not isinstance(batch_state, list):
+                query["batch_state"] = batch_state.value
+            else:
+                query["batch_state"] = {"$in": [bs.value for bs in batch_state]}
+        sort = sort or {"created_on": -1}
```
gpetretto: for the jobs and flows by default we sort on `updated_on`. Don't know if we want to keep the same convention?

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -4731,9 +4749,45 @@ def get_batch_processes(
...
+                {
+                    "$setWindowFields": {
+                        "partitionBy": "$worker",
+                        "sortBy": {**sort},
+                        "output": {"rank": {"$rank": {}}},
+                    }
+                },
+                {"$match": {"rank": {"$lte": max_batches_per_worker}}},
+                {"$unset": "rank"},
+            ]
```
gpetretto: I wonder if instead of making this complicated query, that I suppose may become more of a burden as the size of the collection grows, and is not really optimal for the user it wouldn't be easier to just have the options `--worker` and `--max-results`, so the user decides what to search for. Most likely one has only one batch worker, and if there are many it is not obvious that you want to see the last 10 jobs for every of these workers.
What you get would be clearer, more flexibility and the query much simpler, without the need to make an aggregation.

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: with this amount of information maybe it is worth to have a `BatchDoc` instead of working with dictionaries? Will such an object be used only for the CLI? Or will it give a benefit in other operations?

`start_time` and `end_time` should not be too difficult to get correctly, but can also be avoided as far as I am concerned.

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: ok, let's leave it like this for the moment.

src/jobflow_remote/jobs/runner.py
gpetretto: yes, I was thinking of keeping track of that in the runner. Of course it will try again if the runner is restarted, but at least only once.

src/jobflow_remote/jobs/runner.py
gpetretto: honestly this seems a challenging point. Maybe we can leave it in case this really starts to happen?

src/jobflow_remote/jobs/runner.py
gpetretto: I would say it is not worth (and likely involved). For the batch process it is enough to know that the job was terminated. If one want more info should go to the job details.

tests/integration/test_advanced_options.py
gpetretto: ok, let's keep it for another PR.

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -4731,9 +4749,45 @@ def get_batch_processes(
...
-    def add_batch_process(
-        self, process_id: str, process_uuid: str, worker: str
-    ) -> dict:
+    def get_all_batches(
+        self,
+        worker: str | None = None,
+        batch_state: BatchState | list[BatchState] | None = None,
+        max_batches_per_worker=None,
+        sort=None,
+    ):
```
gpetretto: return type is missing

src/jobflow_remote/jobs/runner.py
```diff
@@ -1118,221 +1121,370 @@ def update_batch_jobs(self) -> None:
...
+        while len(self._cached_batches[worker_name]) > 2 * worker.max_jobs + 5:
+            self._cached_batches[worker_name].popitem(last=False)
+
+    def batch_get_process_id(
+        self,
+        batch_processes: dict,
+        batch_uid: str,
+        worker_name: str,
+        worker: WorkerBase,
+    ):
```
gpetretto: return type missing

src/jobflow_remote/jobs/runner.py
gpetretto: raise is fine with me

gpetretto (commented):
Thanks for the updates. Here are some replies and additional small comments.
I suppose some changes can also be hold, depending on whether we drop the auxiliary document or not.

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: Ok, I'll add the BatchDoc. It can be useful anyway.
I've added start and end time

src/jobflow_remote/jobs/runner.py
davidwaroquiers: Ok, I'll put TERMINATED (will even make more "sense" when it's named RUN_FINISHED)

davidwaroquiers: Done

src/jobflow_remote/cli/batch.py
davidwaroquiers: Yes, that makes sense. Done

src/jobflow_remote/cli/formatting.py
davidwaroquiers: Ok I see. In the case of show_all = False, get_batch_processes_table is called with the default value of status (= to False). So I think it's already the case.
Indeed if we drop the auxiliary for batches it should not happen. I propose to resolve this conversation if what I'm saying is correct but maybe I misunderstood.
For dropping the auxiliary, let's see.

src/jobflow_remote/jobs/runner.py
davidwaroquiers: Done

src/jobflow_remote/jobs/runner.py
davidwaroquiers: ok I resolve

src/jobflow_remote/cli/batch.py
davidwaroquiers: Maybe we can do this then. We get max 10 jobs in total, but if there are more jobs, we inform how many more, e.g. if you were to have 15 batch processes running for one worker, you get 10 that are shown and a message saying something like "10/15 running batches shown"

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: Done

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: Yes, done

src/jobflow_remote/jobs/runner.py
davidwaroquiers: Done

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: Done

davidwaroquiers (commented):
More updates.
I think last thing is to decide to remove the dealing of the batches from the auxiliary collection (or do that later or not do that).
If you can go through the comments again and check if everything is solved and we can have a quick call to see what we do about the above.

src/jobflow_remote/cli/batch.py
davidwaroquiers: This is now done with the max_results. Resolving.

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: BatchDoc Done

src/jobflow_remote/cli/batch.py
```diff
@@ -39,29 +37,87 @@ def processes_list(
...
+            if verbosity > 0:
+                running_jobs = [
+                    [
+                        (str(jid), str(jidx))
+                        for jid, jidx, batch_uid in worker_running_jobs[batch["worker"]]
+                        if batch_uid == batch["batch_uid"]
+                    ]
+                    for batch in batch_processes
+                ]
+            table = get_batch_processes_table(
```
gpetretto, line 84: it seems that here the `max_results` is ignored. However it is unlikely that one will have more than 20 batch jobs running at the same time. And in addition the auxiliary collection could be removed. 
Is it fine to keep it like this?

src/jobflow_remote/cli/formatting.py
gpetretto: I see, but so the case where `UNKNOWN` is used should never happen, right?

src/jobflow_remote/jobs/batch.py
```diff
@@ -176,6 +178,15 @@ def cleanup(self) -> bool:
             return self.host.rmtree(self.files_dir, raise_on_error=False)
         return True
 
+    def get_batch_info(
+        self, batch_dir: Path | str, batch_info_file: str = BATCH_INFO_FILENAME
+    ):
```
gpetretto: missing return type

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: is it? It seems it does not show

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -4731,9 +4750,32 @@ def get_batch_processes(
...
+            else:
+                query["worker"] = {"$in": worker}
+        if batch_state:
+            if not isinstance(batch_state, list):
+                query["batch_state"] = batch_state.value
+            else:
+                query["batch_state"] = {"$in": [bs.value for bs in batch_state]}
+        sort = sort or {"updated_on": -1}
+
+        return list(self.batches.find(query, sort=sort).limit(max_results))
```
gpetretto: I was expecting that this would return a BatchDoc, otherwise I see little point in adding the model if we don't use it to pass the data around. It is at this point used only to insert the data, right?

src/jobflow_remote/jobs/run.py
```diff
@@ -119,81 +125,119 @@ def run_remote_job(run_dir: str | Path = ".") -> None:
...
+def ping(start_time, interval=60, filename=BATCH_INFO_FILENAME):
+    while True:
+        dumpfn(
+            {"start_time": start_time, "last_ping_time": datetime.datetime.utcnow()},
+            fn=filename,
+        )
+        time.sleep(interval)
+
+
+PING_TIME = 60
```
gpetretto: Pinging every minute seems an unnecessarily high frequency.

src/jobflow_remote/jobs/run.py
```diff
@@ -119,81 +125,119 @@ def run_remote_job(run_dir: str | Path = ".") -> None:
...
-    process_uuid: str,
-    max_time: int | None = None,
-    max_wait: int = 60,
+    batch_uid: str,
+    max_time: float | None = None,
+    max_wait: float = 60,
     max_jobs: int | None = None,
     parallel_jobs: int | None = None,
+    sleep_time: float = None,
+    batch_dir: str | Path = None,
```
gpetretto: I wonder if it makes sense that this is an option. If it is not in the CWD the runner will not even find it, since it assume it is there.

gpetretto (commented):
Thanks for the updates. I left a few more comments.

src/jobflow_remote/cli/batch.py
davidwaroquiers: I would indeed keep it like that and see if we switch to the no auxiliary collection for the batches

src/jobflow_remote/cli/formatting.py
davidwaroquiers: I think it should not indeed. You'd just get access to the key and it will then fail if the key does not exist (which again, should not happen) ?

src/jobflow_remote/jobs/batch.py
davidwaroquiers: Done

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: Ah probably did not go through. Now done :)

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: True, I thought that I had to do that then I forgot :) Will do!

src/jobflow_remote/jobs/run.py
davidwaroquiers: 600 ? 3600 ?

src/jobflow_remote/cli/formatting.py
davidwaroquiers: I changed to batch_data["batch_state"] as it should never happen (if it happens, it's a bug).

src/jobflow_remote/jobs/run.py
davidwaroquiers: I've put 600

src/jobflow_remote/remote/queue.py
davidwaroquiers: Removed for now

src/jobflow_remote/jobs/run.py
davidwaroquiers: Removed

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: Done

davidwaroquiers (commented):
I should have addressed all the comments.

gpetretto (approved):
LGTM. Thanks!

PR #363, opened 2025-10-01
Removed auxiliary based management of batch jobs

davidwaroquiers:


src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -4801,58 +4774,63 @@ def add_batch_process(self, process_id: str, batch_uid: str, worker: str) -> dic
...
-            )
-            self.batches.insert_one(batch_doc_dict)
-        return self.auxiliary.find_one_and_update(
-            {"batch_processes": {"$exists": True}},
-            {"$set": {f"batch_processes.{worker}.{process_id}": batch_uid}},
-            upsert=True,
+        batch_doc_dict = get_initial_batch_doc_dict(
+            batch_uid=batch_uid, process_id=process_id, worker=worker
         )
+        return self.batches.insert_one(batch_doc_dict)
```
gpetretto, line 4815: I wonder if it would be worth returning the BatchDoc. Considering where it is used it might be fine sticking with the dictionary, but I wanted to raise the question.

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -4861,21 +4839,15 @@ def remove_batch_process(
...
-            upsert=True,
+        return self.batches.update_one(
+            {"process_id": process_id},
+            {
+                "$set": {
+                    "batch_state": BatchState.FINISHED.value,
+                    "updated_on": datetime.now(),
+                    "end_time": end_time or datetime.now(),
+                }
+            },
```
gpetretto, line 4885: similarly to the other method, may be worth returning a BatchDoc? Or is it just a waste?

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -4801,58 +4774,63 @@ def add_batch_process(self, process_id: str, batch_uid: str, worker: str) -> dic
...
 
-    def remove_batch_process(
-        self, process_id: str, worker: str, end_time: datetime | None = None
+        Returns
+        -------
+        dict
+            The updated document.
+        """
+        return self.batches.update_one(
+            {"process_id": process_id, "start_time": None},
```
gpetretto: Since here the query changed, maybe it is worth changing the index that is created in `build_indexes` and make it just on `process_id`?
Also in `build_indexes` there is `if self.batches is not None:`, which can be removed?

src/jobflow_remote/jobs/runner.py
```diff
@@ -1183,8 +1183,10 @@ def batch_update_running_jobs(
             The list of job ids, job indexes and batch unique ids in the host
             running directory.
         """
-        batch_processes = self.job_controller.get_batch_processes(worker_name).get(
-            worker_name, {}
+        batch_processes = self.job_controller.get_all_batches(
+            worker=worker_name,
+            batch_state=[BatchState.SUBMITTED, BatchState.RUNNING],
```
gpetretto, line 1190: again on the indexes, maybe it is worth adding an index for this query (worker+state or at least worker and state separately)?

src/jobflow_remote/jobs/runner.py
```diff
@@ -1263,10 +1264,10 @@ def batch_get_process_id(
                 process_id=process_id,
             )
             return process_id
-        if self.job_controller.batches is not None:
-            doc = self.job_controller.batches.find_one(
-                {"batch_uid": batch_uid}, projection=["process_id"]
-            )
+        doc = self.job_controller.batches.find_one(
```
gpetretto: Sorry, did not realize it in the previous PR, but would it be possible to avoid explicit mongoDB operations outside the JobController? I know it does not make a big difference, as we implicitly use mongodb syntax in the `update_on_release` of the lock, but I tried to avoid this as much as possible.

Also, I have a doubt on the logic of this point: what is the case in which it gets here? Since in the original version there was not batches collection, where would it take the process_id if the previous logic failed to retrieve it?

src/jobflow_remote/jobs/runner.py
```diff
@@ -1325,8 +1330,12 @@ def batch_update_status(
             for pid in running_processes:
                 if pid in self._cached_running_batch_pids.get(worker_name, set()):
                     continue
+                batch_uid = next(
+                    (b.batch_uid for b in batch_processes_data if b.process_id == pid),
+                    None,
```
gpetretto, line 1347: Does this default make sense? if batch_id is `None`, the how it is used in `get_job_path` below? Shouldn't it give an error if there is no batch_uid matching the pid? Or at least give a warning and skip the next steps?

src/jobflow_remote/jobs/runner.py
```diff
@@ -1335,16 +1344,20 @@ def batch_update_status(
...
                 )
                 if start_time:
                     if worker_name not in self._cached_running_batch_pids:
                         self._cached_running_batch_pids[worker_name] = set()
                     self._cached_running_batch_pids[worker_name].add(pid)
 
             for pid in stopped_processes:
+                batch_uid = next(
+                    (b.batch_uid for b in batch_processes_data if b.process_id == pid),
+                    None,
```
gpetretto, line 1374: Same as above. Maybe there can be a helper function that does a standard check and gives a standard error in handling this?

gpetretto (commented):
Thanks. Here are my comments.

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: I've put an index on ("batch_uid", "worker") (used in update_job_in_batch) and on "process_id" (used in the other main queries)

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: I would probably leave it like that for now unless you think it's worth checking that the updated document follows the schema ? If at some point we need the BatchDoc itself, we can change.

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: Also, it is maybe not worth having it at this stage as it's just used internally. What do you think ?

src/jobflow_remote/jobs/runner.py
davidwaroquiers: Added the three: "worker", "batch_state" and ("worker", "batch_state"), also update_on. I've also modified set_running and set_finished_batch_process to use the batch_uid and not the process_id.

src/jobflow_remote/jobs/runner.py
davidwaroquiers: I've added a method in jobcontroller to get the process id (and worker) from the batch uid and use it here.

In terms of the logic, first it tries to take it from the cached batches (those that are kept in memory in the runner).

If it can't find it there (runner was restarted), then it takes it from the list of batch processes.

I am indeed not completely sure if it's possible that it can't find it there... at least the probability is low. I think it could happen (in the two batch_update_running_jobs and batch_updated_terminated_jobs methods where batch_get_process_id is used) if between the "get_all_batches" (at the beginning of the methods) that returns the list of batch processes  and the get_running/get_terminated just after, a new batch process started. Even then, I actually think it would fail because the info wouldn't be in the database yet.

What do you think ? I think at some point I also thought about having a fallback to get the process_id from the worker (using the RemoteBatchManager). Not sure if it's 1/ worth, 2/ would work, 3/ solves anything (i.e. can the above really happen)

src/jobflow_remote/jobs/runner.py
davidwaroquiers: Indeed. I added a warning and skip (continue the loop)

src/jobflow_remote/jobs/runner.py
davidwaroquiers: I did the same as above. What would you do with the helper function/standard error ?

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -3082,13 +3082,23 @@ def build_indexes(
...
+        )
+        self.batches.create_index(
+            [("batch_uid", 1), ("start_time", 1)], unique=True, background=background
+        )
+        self.batches.create_index(
+            [("worker", 1), ("batch_state", 1)], background=background
+        )
+        self.batches.create_index([("worker", 1)], background=background)
+        self.batches.create_index([("batch_state", 1)], background=background)
+        self.batches.create_index([("updated_on", -1)], background=background)
```
gpetretto, line 3097: Just to check, are all these indexes needed as they are used in the queries done by the runner? 
In general my understanding is that indexes are good for queries, but bad for insertions, so putting a lot of them if they are not useful may not be convenient overall. Maybe it is not a problem, given the likely small size of this collection, but then even the absence of the indexes will not be an issue either

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -4779,9 +4759,22 @@ def get_all_batches(
...
+
+        Parameters
+        ----------
+        batch_uid
+            The batch unique id.
+        """
+        doc = self.batches.find_one(
+            {"batch_uid": batch_uid}, projection=["process_id", "worker"]
+        )
+        return doc["process_id"], doc["worker"]
```
gpetretto, line 4788: If doc is None here it will raise an error that None cannot be indexed

gpetretto (commented):
I left a couple more comments.
Also, it seems that the tests failed in two different ways in the same run. 
I wonder if this can be related to the new changes, but it would seem weird.
For the failure in 3.11 maybe the reason is that it does not guarantee the order in which the jobs are finished and thus it checks the wrong process_id? Or does the test require a specific order of finishing of the batch jobs?

For the failure in 3.12 maybe wait longer before failing?

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: There is not much insertions but there may be a lot of queries so I think it makes sense.

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: Ok I've added a raise of MissingDocument if that happens. I'm dealing with the MissingDocument in runner.batch_get_process_id (used for updating job documents with the correct pid, if it's not found it will be set to None) using a try/except.

davidwaroquiers (commented):
Addressed more comments

davidwaroquiers:
> I left a couple more comments. Also, it seems that the tests failed in two different ways in the same run. I wonder if this can be related to the new changes, but it would seem weird. For the failure in 3.11 maybe the reason is that it does not guarantee the order in which the jobs are finished and thus it checks the wrong process_id? Or does the test require a specific order of finishing of the batch jobs?
> 
> For the failure in 3.12 maybe wait longer before failing?

For the failure in 3.11, it should be fixed. Indeed, the order is not guaranteed as the batches may have been updated at the exact same time. Took me a bit of time to understand why it happened so often (I also had it locally). One thing I did not know is that MongoDb "only" stores datetime up to milliseconds (not microseconds). I've added an additional sort on process_id to make sure the order is always the same (well it could still happen but the probability is way way lower).

For 3.12, I've put 30 seconds instead of 20 (let's see). Even though, 20 seconds already seemed quite a lot. Maybe at some point, we should look whether we can decrease that (maybe decrease all, or at least most of the delays to very small values, like 0.05 or something, but probably we have to do that with care).

src/jobflow_remote/jobs/runner.py
gpetretto: I was mostly referring to the logic of `next((b.batch_uid for b in batch_processes_data if b.process_id == pid), None,)`. A function that given a li of processes and the pid gives the batch_uid and raise (or warn) if no match found, in case it would be useful in other parts of the code. But if these two are the only places where this is done can also stay like this here.

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: fine with me for the moment

src/jobflow_remote/jobs/runner.py
```diff
@@ -1263,12 +1266,17 @@ def batch_get_process_id(
...
-                {"batch_uid": batch_uid}, projection=["process_id"]
+        try:
+            process_id, wk_name = self.job_controller.get_batch_process_id(batch_uid)
+            if wk_name != worker_name:
+                raise RuntimeError("Wrong worker")
+        except MissingDocumentError:
+            process_id = None
+            logger.warning(
+                f"error trying to get the process id and worker for batch with unique id: {batch_uid}",
+                exc_info=True,
```
gpetretto, line 1281: I am not 100% sure I understand the implication of this. What would happen if batch_get_process_id returns None? Wouldn't it set the process_id to None and then never be set correctly again? Or will be updated again later?

davidwaroquiers: No that's True, it will not be updated again in the JobDoc. Actually, it will be updated first by update_running_jobs, and be set to None that time, and maybe - but I wouldn't know in which case this could really happen - it will be updated "correctly", i.e. with a real pid value, by the update_terminated_jobs. Once updated_terminated_jobs has been called, the job is set to TERMINATED and it will not try again. Maybe we could have something to update batch job ids manually in case something like this happens, e.g. an admin command or else that looks for jobs where remote.process_id is None, and tries to figure out from which batch it came from ?

src/jobflow_remote/jobs/runner.py
davidwaroquiers: It's the only two places. Lets keep it like this for now then.

davidwaroquiers (commented):
One more step closer to merge :D

src/jobflow_remote/jobs/runner.py
gpetretto: Sorry, just one more comment here. If I remember correctly we agreed that unless there are more than one runner process dealing the the batch jobs this should never happen, right? So, since at the moment there should be no way of assigning more processes to that, this should basically mean that if it gets here there are either multiple runners running or there is a bug in the code.
If that is the case maybe it would be better to add a comment explaining this?

davidwaroquiers: Yes. Done

gpetretto (approved):
The code is fine with me. 
Should the documentation be updated in this PR?
Should we warn the interested users before we merge this breaking change?

davidwaroquiers:
> The code is fine with me. Should the documentation be updated in this PR? Should we warn the interested users before we merge this breaking change?

Good point. Just added some documentation about the batches. I don't think we need to inform now. We should put it in the release notes and do it when we do a release.

Can you review the last change (just the doc I think) ?

If it's fine for you, you can merge.

davidwaroquiers:
This pull request, together with #354 updates the way batch jobs are handled internally. A new collection has been added, which keeps track of past batch jobs while in the previous version, only current batch processes were available in the database. This is a backward incompatible change for batch processes. To prevent from having any problem while updating jobflow-remote using this new PR, you should wait until all your batch processes are finished. Note that, as batch processes are still an experimental feature, we haven't implemented an automatic upgrade mechanism that would have allowed the above (provided runner is shutdown). Pinging @blaked8619, @Andrew-S-Rosen, @JaGeo and @utf . If you have anyone else in mind that may be affected by this change, please forward.

Andrew-S-Rosen:
@davidwaroquiers: Thanks for the ping! 

Assuming I am starting from a fresh database and install, is there anything we have to do differently with regards to batch jobs given this PR? I see that we now need to have `batches_collection` defined in the YAML. Anything else?

We are running a 128-node parallel batch job soon...

davidwaroquiers:
> @davidwaroquiers: Thanks for the ping! 
> 
> Assuming I am starting from a fresh database and install, is there anything we have to do differently with regards to batch jobs given this PR? I see that we now need to have `batches_collection` defined in the YAML. Anything else?
> 
> We are running a 128-node parallel batch job soon...

In principle nothing to do if you start from a fresh database. Even the batches_collection is optional. It's only if you want to change the name of the collection used.

When we make the new release, maybe we will make an upgrade to make sure the collection is initialized if you were to use an existing project.

PR #364, opened 2025-10-02
Added check of environment variables in jf project check.

davidwaroquiers:
Will also suggest a variable if it finds one setting that is a close match.


doc/source/conf.py
```diff
@@ -216,3 +225,24 @@
...
+)
+
+settings_schema = JobflowRemoteSettings.model_json_schema()
+with open("_static/_tmp/jfremote_settings.json", "w") as f:
+    json.dump(settings_schema, f)
+generate_from_filename(
+    "_static/_tmp/jfremote_settings.json",
+    "_static/_tmp/jfremote_settings.html",
+    config=GenerationConfiguration(with_footer=False),
+)
```
gpetretto, line 248: should the json files be removed at the end?

src/jobflow_remote/cli/project.py
```diff
@@ -175,6 +175,46 @@ def check(
...
+    extra_vars = [
+        k
+        for k in os.environ
+        if k.lower().startswith(prefix)
+        and k[len(prefix) :].lower() not in SETTINGS.model_fields
+    ]
+    if extra_vars:
+        out_console.print(
+            "The following JFREMOTE_ prefixed environment variables were found:\n - "
+        )
```
gpetretto, line 195: It seems that this message does not specify explicitly that these are variable that are not present among the configuration options and thus might be wrong.

tests/db/cli/test_project.py
```diff
@@ -84,6 +84,28 @@ def test_check(job_controller, monkeypatch, tmp_dir, run_check_cli) -> None:
         run_check_cli(["project", "check"], required_out=[*output, duplicated_msg])
 
 
+def test_check_env_vars(run_check_cli):
+    import os
+
+    # explicit typo in variable (should be with an "S" on PROJECT)
+    os.environ["JFREMOTE_PROJECT_FOLDER"] = "my project folder"
+    # random variable (should not be suggested)
+    os.environ["JFREMOTE_ZZZZZZZZZZZZZZZ"] = "zzz"
```
gpetretto: Maybe use [MonkeyPatch.setenv](https://docs.pytest.org/en/stable/reference/reference.html#pytest.MonkeyPatch.setenv) instead, so that they are removed? Otherwise they may hinder the output of other tests with jf project check?

.gitignore
```diff
@@ -163,3 +163,5 @@ cython_debug/
 .DS_store
 
 .sesskey
+
+doc/source/_static/_tmp/
```
gpetretto, line 168: maybe add a comment on why it is here, in case we do not remember.

gpetretto (approved):
Thanks for adding this.
I approved, but left some minor comments

doc/source/conf.py
davidwaroquiers: I'd say we can keep them there. Anyway the files are ignored (.gitignore).

src/jobflow_remote/cli/project.py
davidwaroquiers: I've clarified the message. Thx

tests/db/cli/test_project.py
davidwaroquiers: good point. Done, thx

.gitignore
davidwaroquiers: Yes, done

davidwaroquiers (commented):
Thx for the review. I've addressed the comments.

PR #366, opened 2025-10-06
Initial implementation of metadata in flow info.

davidwaroquiers:


src/jobflow_remote/cli/flow.py
```diff
@@ -256,8 +258,44 @@ def delete(
...
+    verbosity: verbosity_opt = 0,
+    stored_data_keys: Annotated[
+        Optional[list[str]],
+        typer.Option(
+            "--stored-data-key",
+            "-sdk",
+            help="Key to be shown from the stored_data field.",
+        ),
+    ] = None,
+    cli_output_keys: Annotated[
```
gpetretto: Since these two seem equivalent to those for the jobs list, maybe can be made as common types in the `types` module?

Is it really 100% equivalent the field description? Or does it make something slightly different here for the flows?

src/jobflow_remote/cli/flow.py
```diff
@@ -256,8 +258,44 @@ def delete(
...
+    )
+    if not set(output_keys).issubset(header_name_data_getter_map):
+        exit_with_error_msg(
+            f"Header keys not supported: {set(output_keys).difference(header_name_data_getter_map)}"
+        )
+    if stored_data_keys and not set(output_keys).isdisjoint(stored_data_keys):
+        exit_with_error_msg(
+            "Specifying a stored data key which is a standard column is disallowed."
+        )
+
```
gpetretto, line 290: This also seems equivalent to the one in jobs list. Maybe would be worth having an helper function in `cli.utils`, so if something changes we do not have to worry changing in multiple places?

src/jobflow_remote/cli/formatting.py
```diff
@@ -120,6 +120,7 @@ def format_run_time(ji: JobInfo) -> str:
         if ji.lock_time
         else None,
     ),
+    "metadata": ("Metadata", lambda ji: render_scope(ji.metadata)),
```
gpetretto: This can be potentially be arbitrarily big. I have seen that `render_scope` has `max_length` and `max_string` arguments. The first seems to limit the number of elements that get printed, while the second the maximum length of every single string by truncating. 
I think it may be good to set them. From the few tests it seems that a small number like 1/2/3 could work for `max_length`. As for the maximum string length it can probably be bigger. 500? 1000?
In any case the full value should always be accessible from `jf job info`. 
In any case, the bigger the metadata are the less likely it is it may be useful in a huge table with many of them.

src/jobflow_remote/cli/formatting.py
```diff
@@ -211,6 +213,7 @@ def get_flow_info_table(flows_info: list[FlowInfo], verbosity: int) -> Table:
             row.append(", ".join(workers))
             job_states = "-".join(js.short_value for js in fi.job_states)
             row.append(job_states)
+            row.append(render_scope(fi.flow_metadata))
```
gpetretto: Same as for the job table, this can maybe be truncated?

src/jobflow_remote/cli/formatting.py
```diff
@@ -298,29 +301,19 @@ def format_job_info(
...
-            flow_info.job_names[i],
-            state,
-            f"{job_id}  ({flow_info.job_indexes[i]})",
-            flow_info.workers[i],
-        ]
-
-        table.add_row(*row)
-
+    if verbosity > 0:
+        title += f"\nMetadata: {flow_info.flow_metadata}"
```
gpetretto: Should this be a render_scope  as well (with no truncation)?

src/jobflow_remote/cli/formatting.py
```diff
@@ -298,29 +301,19 @@ def format_job_info(
...
+    table = get_job_info_table(
+        flow_info.jobs_info,
```
gpetretto: Should this check that `jobs_info` is not None? Or we assume that the input will be built correctly?

src/jobflow_remote/jobs/data.py
```diff
@@ -477,9 +480,12 @@ class FlowInfo(BaseModel):
     job_names: list[str]
     parents: list[list[str]]
     hosts: list[list[str]]
+    flow_metadata: dict
+    jobs_metadata: list[dict]
```
gpetretto: to be removed here and in the methods below

src/jobflow_remote/jobs/data.py
```diff
@@ -477,9 +480,12 @@ class FlowInfo(BaseModel):
...
+    jobs_info: Optional[list[JobInfo]] = None
 
     @classmethod
-    def from_query_dict(cls, d) -> "FlowInfo":
+    def from_query_dict(cls, d, with_jobs_info=False) -> "FlowInfo":
```
gpetretto: I wonder if `with_jobs_info` should be an argument or if the function should determine whether the required information is there or not.

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -2491,6 +2491,7 @@ def get_flows_info(
         limit: int = 0,
         skip: int = 0,
         full: bool = False,
+        with_jobs_info: bool = False,
```
gpetretto, line 2504: Isn't this redundant a duplicate of the `full` argument? At least it will not make sense to call this method with `with_jobs_info=True` and `full=False`, as the jobs data would not be there at all. And calling it with `full=True` and `with_jobs_info=False` would basically featch a lot of data for nothing.

gpetretto (commented):
Nice! 
Here are my initial comments.

src/jobflow_remote/jobs/data.py
davidwaroquiers: Done

src/jobflow_remote/cli/formatting.py
davidwaroquiers: Good idea, I'm just not sure what these number should be. It seems that max_length only applies on lists. Also any nested dict will be output. And for example, if a metadata is about the elements in a structure, I guess we want them all, let's imagine a quarternary or 5-ary system, we wouldn't want to not see the 4 or 5 elements). What do you think ?

src/jobflow_remote/cli/formatting.py
davidwaroquiers: Same comment as above, with what parameters (should this be an option in the cli ?). I've also seen rich.pretty.Pretty that does limit number of key-value pairs, can also limit depth, and do additional stuff like identation/justify/....

src/jobflow_remote/cli/formatting.py
davidwaroquiers: Yes, or a Pretty (see other comments). I'll not make any changes right now before discussing and deciding.

src/jobflow_remote/cli/formatting.py
davidwaroquiers: I think it should be built correctly right now but it's True that it would be better to handle that case if it happens. What would you do in that case ? The simplest thing could be to pass an empty list.

src/jobflow_remote/jobs/data.py
davidwaroquiers: Could be an option indeed. I'll see to add it afterwards.

src/jobflow_remote/cli/flow.py
davidwaroquiers: Tried to do something. What do you think ?

davidwaroquiers (commented):
Ok, I've made a few changes (still work in progress, up for discussion)

src/jobflow_remote/cli/formatting.py
gpetretto: maybe we can have `max_length` somewhere between 5 and 10? It could be an option, but it seems a bit of an overkill.
rich.pretty.Pretty does seem interesting as an alternative. `max_length` and `max_string` seem to have the same effect and it has the addition of `max_depth`. Maybe it can be used, but remains open the question of which value to put for that additional one.

In general, I think it would be better to have some limitation, even somewhat on the large size, rather than the possibility of having huge dictionaries. At least for jobs one can use the job info, if really the whole big metadata needs to be seen. A huge one will not be of much help in a table anyway.

src/jobflow_remote/cli/formatting.py
gpetretto: If the empty list works, maybe can just be `flow_info.jobs_info or []`, just to avoid potential hard failures?

gpetretto (commented):
It is fine with me. Just left to decide what to do about the shortening of the metadata.

Also maybe add a check on the metadata in the tests?

src/jobflow_remote/cli/formatting.py
davidwaroquiers: Sure, I've added as Pretty with max_length 10, max_string 1000 and max_depth 3

src/jobflow_remote/cli/formatting.py
davidwaroquiers: Added Pretty with the same max_* as above

src/jobflow_remote/cli/formatting.py
davidwaroquiers: Done

src/jobflow_remote/cli/formatting.py
davidwaroquiers: Yes, done

gpetretto (approved):

davidwaroquiers (commented):
Done

PR #374, opened 2025-10-12
Clearer error message for undefined active project

### LINKED ISSUE #286: Better error handling for incorrectly formatted project YAML
When the project YAML is incorrectly formatted, we get back

```
The active project could not be determined and it is required to execute this command
```

However, this is confusing for new users because it doesn't actually inform the user that their project YAML is misformatted, and it's not clear what went wrong. A clearer error message should be provided, perhaps even just specifying that the YAML is not formatted appropriately.

CC @blaked8619

## Reproduction

To reproduce the issue, make the following project YAML:

```yaml
name: cms
workers:
  example_worker:
    type: local
```

Then run any `jf` command, such as `jf project check --errors`.

Andrew-S-Rosen:
Closes https://github.com/Matgenix/jobflow-remote/issues/286 by providing a clearer error message to users when the YAML is likely incorrectly formatted. This isn't really a solution per se, but at the very least it prompts the user where to look to fix things.

### DIFF
--- a/src/jobflow_remote/cli/utils.py
+++ b/src/jobflow_remote/cli/utils.py
@@ -290,7 +290,8 @@ def cli_error_handler(func):
             raise  # Do not capture click or typer exceptions
         except ProjectUndefinedError:
             exit_with_error_msg(
-                "The active project could not be determined and it is required to execute this command"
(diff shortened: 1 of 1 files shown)


gpetretto (approved):
Thanks!

PR #383, opened 2025-10-21
Added the possibility to have a custom scheduler io

davidwaroquiers:
The scheduler_type option config can now take a str (which is then directly converted to an existing qtoolkit scheduler io) or a dict based representation of a class.
Added the tests.


gpetretto (approved):
Looks good to me, but I don't see the tests. Maybe they were not committed?
Also, it seems it contains the changes related to the flow info, so maybe better merge the other first?

davidwaroquiers:
> Looks good to me, but I don't see the tests. Maybe they were not committed? Also, it seems it contains the changes related to the flow info, so maybe better merge the other first?

Duh, I forgot to commit the test indeed. Now it's done. For the flow info I don't know why ... maybe I started the branch from the flow info (I thought I had done it from develop but probably not ...). Let's wait for the merge of flow info then I'll merge it back here and push.

gpetretto:
I see that the unit test is failing. Indeed I think that inside the model_validator it should raise a ValueError:
https://docs.pydantic.dev/latest/concepts/validators/#raising-validation-errors

PR #384, opened 2025-10-21
Changed TERMINATED state to RUN_FINISHED.

davidwaroquiers:
Also modified all the relevant methods to follow the same naming convention.
Also changed the batch directory for finished job runs to "run_finished" instead of "terminated".
Updated documentation.
Added upgrade method for this change.
Added tests.


davidwaroquiers:
@gpetretto This is getting ready I would say. Probably still a few things to check. I'd like to add another test also on the update method. But I guess it would be good to have a review at this stage if anything big needs to be modified I'd rather not finalize tests otherwise I'd need to also maybe change them. I also still need to add the move of the terminated dir to run_finished dir in the upgrade for the batch submission mode.

doc/source/user/advancedoptions.rst
```diff
@@ -208,7 +208,7 @@ You can customize the output using the following options:
 
 - ``--worker-name``: show only batches from a specific worker
 - ``--max-results``: limit the number of results shown (default: 20)
-- ``--batch-state``: filter by batch state (e.g. ``RUNNING``, ``TERMINATED``, ``FAILED``)
+- ``--batch-state``: filter by batch state (i.e. ``SUBMITTED``, ``RUNNING`` or ``FINISHED``)
```
gpetretto: In principle it is not really needed to specify the list of states, as the `-h` already provides all the options available.

src/jobflow_remote/cli/runner.py
```diff
@@ -446,3 +447,49 @@ def reset(
...
+        bool,
+        typer.Option(
+            "--connect-interactive",
+            "-i",
+            help="Activate the connection for interactive remote host",
+        ),
+    ] = False,
+) -> None:
+    """
+    Update the state of jobs, flows and batches.
```
gpetretto: This description is a bit misleading, since only jobs that are in the SUBMITTED or RUNNING states (or batch equivalent) will be modified here. 
Also, not sure there is any way the Flow state will change as a result of this action.

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -4764,6 +4759,16 @@ def get_all_batches(
                 query["batch_state"] = batch_state.value
             else:
                 query["batch_state"] = {"$in": [bs.value for bs in batch_state]}
+        return query
+
+    def get_all_batches(
```
gpetretto: I notice that this method is a bit weird as it is called get_all_batches but for example has a limit, so by default does not acutally return all the batches jobs. Also it can be filtered on states. 
Is there any downside in calling this `get_batches`? It would seem more accurate.

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -5306,14 +5326,19 @@ def backup_restore(
...
+                self.jobs_collection,
+                self.flows_collection,
+                self.auxiliary_collection,
+                self.batches_collection,
+            ],
+            [self.jobs, self.flows, self.auxiliary, self.batches],
         ):
             count = collection.count_documents({})
-            if count > 0:
+            if count > 0 and name != "jf_auxiliary":
```
gpetretto: It is not such a big deal, but why adding this check instead of not looping over jf_auxiliary as in the original implementation?

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -5337,9 +5367,19 @@ def backup_restore(
...
+                    # All collections except the "batches" one should have exactly one matching file
+                    raise RuntimeError(
+                        f"{len(files_paths)} files matching the name {file_name} were found in {dir_path}"
+                    )
+                if name == "batches" and len(files_paths) > 1:
+                    # Multiple files for the batches collection is not allowed (as for other collections)
+                    raise RuntimeError(
+                        f"{len(files_paths)} files matching the name {file_name} were found in {dir_path}"
+                    )
+                if name == "batches" and len(files_paths) == 0:
```
gpetretto: is this `if` needed? If it gets here can only be name=="batches" and len(file_paths)<1, given than len(file_paths)>=0 this seems redundant.

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -5337,9 +5367,19 @@ def backup_restore(
...
             if len(files_paths) != 1:
-                raise RuntimeError(
-                    f"{len(files_paths)} files matching the name {file_name} were found in {dir_path}"
-                )
+                if name != "batches":
+                    # All collections except the "batches" one should have exactly one matching file
+                    raise RuntimeError(
+                        f"{len(files_paths)} files matching the name {file_name} were found in {dir_path}"
+                    )
+                if name == "batches" and len(files_paths) > 1:
```
gpetretto: Also here it seems that can be reached only if name=="batches"

src/jobflow_remote/jobs/upgrade.py
```diff
@@ -267,3 +318,64 @@ def upgrade_to_0_1_5(
...
+            collection="auxiliary",
+            query={"batch_processes": {"$exists": True}},
+            one_doc_check=count_batch_processes_old,
+        ),
+        UpgradeCondition(
+            description="There should not be any SUBMITTED or RUNNING batch process in the batches collection",
+            collection="batches",
+            query={"batch_state": {"$in": ["SUBMITTED", "RUNNING"]}},
+        ),
+    ],
```
gpetretto: To be fair, I do not like much this approach of having a lot of arguments for the decorator. Makes it quite difficult to read and interpret. 
However, I am not sure I have a better solution. One option: the upgrade function itself performs the checks (that can also be easied by shared objects) and raises a specific exception that can be possibly be handled by the upgrader. But I don't like this much either, as you will not be able to know beforehand if the check passes without calling the registered upgrade function.
At least I wanted to check if we can do something easier.

src/jobflow_remote/jobs/upgrade.py
```diff
@@ -204,6 +246,15 @@ def upgrade(
...
 
+        if failed_conditions := self.check_upgrade_conditions(versions_needing_upgrade):
+            err = ["Some upgrade conditions were not satisfied:"]
+            for vv, failed_cond in failed_conditions:
+                err.append(
+                    f" - {failed_cond['condition'].description} (for version {vv}), found {failed_cond['count']}"
+                )
+            logger.error("\n".join(err))
+            return False
+
```
gpetretto, line 283: It is not obvious which should be the correct behaviour, but maybe this check should also be done in the `dry_run` execution? Not as a blocking error, but as an information, that the upgrade will fail due to the current status of the DB?

src/jobflow_remote/jobs/upgrade.py
```diff
@@ -37,19 +37,44 @@ class UpgradeAction:
...
+            return None
+        count = collection.count_documents(self.query)
+        if count == 0:
+            return None
+        if count == 1 and self.one_doc_check is not None:
+            doc = collection.find_one(self.query)
+            count = self.one_doc_check(doc)
+            if count == 0:
+                return None
+        return {"condition": self, "count": count}
```
gpetretto: I suppose the object can also change in the future, but it seems to handle a limited number of checks. I wonder if it would be easier that the `UpgradeCondition` only has a `description` and an abstract `check` method? Maybe one case make a subclass that implements potential standard checks like `NoDocuments` with also a `collection`+ `query` argument. The `one_doc_check` is not so different and it seems to make the logic less transparent.
If you think it is not worth, it can also be left for the future, altough it may be annoying to refactor this later on.

gpetretto (commented):
I think it is fine in general. I have left some comments, mostly on the upgrade part.
I have then two general points to raise:
1) If I am not mistaken, you have not added any safeguard in case the user has a `TERMINATED` Job and does not run the `jf admin upgrade`. I suppose in this case will just get a hard failure. I have pushed my original test implementation of the renaming of the states, where I had added a check so that in case it tries to create a JobState.TERMINATED` instance it raises a specific message mentioning that the upgrade should be executed. Not sure it is worth, in the sense that it means that this check will always be performed, but ifyou want to check it is here: https://github.com/Matgenix/jobflow-remote/commit/92815fb883dd6f999e89cd9e2d0faf6b1640dca4 (the parts related to the `DeprecatedStateError`
2) I wonder if we should really make this version as 1.0. Since there are quite some changes between this and the batches I am afraid of potential bugs there. But it is fine with me if you think it is better to a "stable" version.

doc/source/user/advancedoptions.rst
davidwaroquiers: Done, I removed

src/jobflow_remote/cli/runner.py
davidwaroquiers: I improved the doc.

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: Yes, done.

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: Yes, done.

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: Done

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: Yes, removed.

src/jobflow_remote/jobs/upgrade.py
davidwaroquiers: From our discussion, we did not find an easier solution. I still modified a bit the code here so that the decorator is on one line and is easier to read.

src/jobflow_remote/jobs/upgrade.py
davidwaroquiers: done

src/jobflow_remote/jobs/upgrade.py
davidwaroquiers: Done

davidwaroquiers (commented):
I think I have addressed all the comments

davidwaroquiers:
I've added the validation as suggested in 1.
I would go for a 1.0 once we have included the last important PRs. If bugs arise, we'd make new 1.0.x releases with bug fixes and so on.


> I think it is fine in general. I have left some comments, mostly on the upgrade part. I have then two general points to raise:
> 
> 1. If I am not mistaken, you have not added any safeguard in case the user has a `TERMINATED` Job and does not run the `jf admin upgrade`. I suppose in this case will just get a hard failure. I have pushed my original test implementation of the renaming of the states, where I had added a check so that in case it tries to create a JobState.TERMINATED`instance it raises a specific message mentioning that the upgrade should be executed. Not sure it is worth, in the sense that it means that this check will always be performed, but ifyou want to check it is here: https://github.com/Matgenix/jobflow-remote/commit/92815fb883dd6f999e89cd9e2d0faf6b1640dca4 (the parts related to the`DeprecatedStateError`
> 2. I wonder if we should really make this version as 1.0. Since there are quite some changes between this and the batches I am afraid of potential bugs there. But it is fine with me if you think it is better to a "stable" version.

src/jobflow_remote/cli/runner.py
```diff
@@ -446,3 +447,57 @@ def reset(
...
+            "-i",
+            help="Activate the connection for interactive remote host",
+        ),
+    ] = False,
+) -> None:
+    """
+    Update the "submitted" and "running" states of jobs, flows and batches.
+
+    This command updates the `"submitted"` and `"running"` states of jobs, flows, and batches.
+    It does **not** perform submissions, uploads, downloads, or any other side-effecting actions.
```
gpetretto: I have seen that you have put some words with the markdown notation `**not**` for bold, but it does not look like this is applied when running the command with the `-h` option. To me it just shows something like `**not**`. Is this the intended behaviour? Because a generic user might not know its meaning.

src/jobflow_remote/jobs/upgrade.py
```diff
@@ -267,3 +344,106 @@ def upgrade_to_0_1_5(
...
+    if len(batches_docs) > 1:
+        raise RuntimeError(
+            "More than one document with batch processes found in the auxiliary collection."
+        )
+    batch_doc = batches_docs[0]
+    if batch_doc["batch_processes"] is None:
+        return None
+    count = 0
+    for batch_processes_dict in batch_doc["batch_processes"].values():
+        count += len(batch_processes_dict)
```
gpetretto, line 366: isn't this the same code that is in `count_batch_processes_old`? Should that function be used? Or removed otherwise?

tests/db/jobs/test_upgrade.py
```diff
@@ -0,0 +1,93 @@
...
+    assert (
+        failed_condition3["condition"].description
+        == "There should be no document in the 'flows' collection"
+    )
+    assert failed_condition3["count"] == 2
+
+    with caplog.at_level("ERROR"):
+        assert db_upgrader.upgrade(from_version=None, target_version="99.0") is False
+    assert job_controller.get_current_db_version() == package_version
+    print(caplog.text)
```
gpetretto: Should this be removed?

gpetretto (approved):
All the comments have been addressed.
I left a few minor ones.

src/jobflow_remote/cli/runner.py
davidwaroquiers: Removed

src/jobflow_remote/jobs/upgrade.py
davidwaroquiers: True, I forgot to remove that function that is not used anywhere anymore.
Removed.

tests/db/jobs/test_upgrade.py
davidwaroquiers: Done

davidwaroquiers (commented):
Addressed the three comments.

PR #388, opened 2025-10-27
CLI commands to remove job folders

gpetretto:
Following the discussion in https://matsci.org/t/create-jobs-immune-to-stop-children/65765/1 I thought it would would be useful to have a command to just remove the execution folders, without the need to remove the content of the DB.
Two options implemented: `jf job files delete` and `jf flow clean`. I am not fully convinced about the namings, but changing will be easy.

In the same discussion it emerged that `jf job rerun` was setting to `READY` a `WAITING` Job. It is now not possible to rerun a `WAITING` Job. If needed it is still possible to change the state using `jf job set-state`.

### DIFF
--- a/src/jobflow_remote/cli/flow.py
+++ b/src/jobflow_remote/cli/flow.py
@@ -17,6 +17,7 @@ from jobflow_remote.cli.formatting import (
 from jobflow_remote.cli.jf import app
 from jobflow_remote.cli.jfr_typer import JFRTyper
 from jobflow_remote.cli.types import (
+    break_lock_opt,
     cli_output_keys_opt,
     count_opt,
     days_opt,
@@ -41,12 +42,15 @@ from jobflow_remote.cli.types import (
     start_date_opt,
     stored_data_keys_opt,
     verbosity_opt,
+    wait_lock_opt,
+    yes_opt,
 )
 from jobflow_remote.cli.utils import (
     ReportInterval,
     SortOption,
     check_incompatible_opt,
     check_output_stored_data_keys,
+    check_stopped_runner,
     exit_with_error_msg,
     exit_with_warning_msg,
     get_job_controller,
@@ -57,6 +61,7 @@ from jobflow_remote.cli.utils import (
 )
 from jobflow_remote.jobs.graph import get_graph, plot_dash
 from jobflow_remote.jobs.report import FlowsReport
+from jobflow_remote.jobs.state import JobState
 
 app_flow = JFRTyper(
     name="flow", help="Commands for managing the flows", no_args_is_help=True
@@ -181,6 +186,8 @@ def delete(
             "the worker associated with the deleted Flows",
         ),
     ] = False,
+    wait: wait_lock_opt = None,
+    break_lock: break_lock_opt = False,
 ) -> None:
     """Permanently delete Flows from the database"""
     check_incompatible_opt({"start_date": start_date, "days": days, "hours": hours})
@@ -244,17 +251,25 @@ def delete(
         if progress:
             progress.add_task(description="Deleting flows...", total=None)
 
-        jc.delete_flows(
+        deleted = jc.delete_flows(
             flow_ids=to_delete,
             delete_output=delete_output,
             delete_files=delete_files,
             cancel_processes=not keep_processes,
             max_limit=max_limit,
+            wait=wait,
+            break_lock=break_lock,
         )
 
-    out_console.print(
-        f"Deleted Flow(s) with id: {', '.join(str(i) for i in to_delete)}"
-    )
+    if not_deleted := set(to_delete) - set(deleted):
+        out_console.print(
+            f"Some of the selected Flows were not deleted: {', '.join(i for i in not_deleted)}"
+        )
+
+    if deleted:
+        out_console.print(
+            f"Deleted Flow(s) with id: {', '.join(str(i) for i in deleted)}"
+        )
 
 
 @app_flow.command(name="info")
@@ -476,13 +491,32 @@ def clean(
     metadata: metadata_opt = None,
     locked: locked_flow_opt = False,
     verbosity: verbosity_opt = 0,
-    force: force_opt = False,
+    force: Annotated[
+        bool,
+        typer.Option(
+            "--force",
+            "-f",
+            help="Do not check if runner is active",
+        ),
+    ] = False,
+    yes: yes_opt = False,
+    all_states: Annotated[
+        bool,
+        typer.Option(
+            "--all-states",
+            "-as",
+            help="Delete files for Jobs in any state",
+        ),
+    ] = False,
 ):
     """
     Remove the files of the executed Jobs.
     """
     check_incompatible_opt({"start_date": start_date, "days": days, "hours": hours})
     check_incompatible_opt({"end_date": end_date, "days": days, "hours": hours})
+
+    if all_states and not force:
+        check_stopped_runner(error=True)
 
     jc = get_job_controller()
 
@@ -506,7 +540,7 @@ def clean(
     if not flows_info:
         exit_with_warning_msg("No flows matching criteria")
 
-    if not force:
+    if not yes:
         if verbosity:
             preamble = Text.from_markup(
                 f"[red]This operation will [bold]delete the files of the following {len(flows_info)} Flow(s)[/bold][/red]"
@@ -531,6 +565,13 @@ def clean(
         out_console.print("Deleting files...")
     else:
         spinner_cm = loading_spinner(processing=False)
+
+    skipped_jobs = []
+    cleanable_states = (
+        JobState.COMPLETED,
+        JobState.FAILED,
+        JobState.REMOTE_ERROR,
+    )
     with spinner_cm as progress:
         if progress:
             progress.add_task(description="Deleting files...", total=None)
@@ -538,7 +579,11 @@ def clean(
         for fi in flows_info:
             for ji in fi.jobs_info:
                 if ji.run_dir:
-                    jobs_info[ji.db_id] = ji
+                    if not all_states and ji.state not in cleanable_states:
+                        skipped_jobs.append(ji)
+                    else:
+                        jobs_info[ji.db_id] = ji
+
         deleted = jc.safe_delete_files(list(jobs_info.values()))
 
     deleted_dict = {ji.db_id: ji for ji in deleted}
@@ -547,6 +592,25 @@ def clean(
         out_console.print("Folder was not deleted for the following jobs:")
         for db_id in not_deleted:
             out_console.print(f" - {db_id}: {jobs_info[db_id].run_dir}")
+    if skipped_jobs:
+        out_console.print(
+            f"{len(skipped_jobs)} were not cleaned-up due to their state:"
+        )
+        if len(skipped_jobs) < 10:
+            for ji in skipped_jobs:
+                out_console.print(f" - {ji.db_id} - {ji.state}")
+        else:
+            confirmed = False
+            if not yes:
+                text = (
+                    "The number of skipped jobs is too large to be printed, the list can be "
+                    "dumped to the `skipped_cleanup.dat` file. Do you want to create the file?"
+                )
+                confirmed = Confirm.ask(text, default=False)
+            if yes or confirmed:
+                with open("skipped_cleanup.dat", "w") as f:
+                    for ji in skipped_jobs:
+                        f.writelines(f" - {ji.db_id} - {ji.state}")
 
     out_console.print(f"Deleted execution folders of {len(deleted)} Jobs")
 
--- a/src/jobflow_remote/cli/flow.py
+++ b/src/jobflow_remote/cli/flow.py
@@ -462,6 +462,92 @@ def resume(
     out_console.print(f"{n_jobs} Job(s) resumed")
 
 
+@app_flow.command()
+def clean(
+    job_id: job_ids_opt = None,
+    db_id: db_ids_opt = None,
+    flow_id: flow_ids_opt = None,
+    state: flow_state_opt = None,
+    start_date: start_date_opt = None,
+    end_date: end_date_opt = None,
+    name: name_opt = None,
+    days: days_opt = None,
+    hours: hours_opt = None,
+    metadata: metadata_opt = None,
+    locked: locked_flow_opt = False,
+    verbosity: verbosity_opt = 0,
+    force: force_opt = False,
+):
+    check_incompatible_opt({"start_date": start_date, "days": days, "hours": hours})
+    check_incompatible_opt({"end_date": end_date, "days": days, "hours": hours})
+
+    jc = get_job_controller()
+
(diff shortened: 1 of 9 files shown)


src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -2562,7 +2568,12 @@ def get_flows_info(
         # Only use the full aggregation if more job details are needed.
         # The single flow document is enough for basic information
         if with_jobs_info:
```
davidwaroquiers: A weird usage of this would be to pass an empty list, which would then not follow what's mentioned in the docstring. Do you think it's worth changing the condition to something like:


```suggestion
        if with_jobs_info is not False:
```

src/jobflow_remote/cli/flow.py
```diff
@@ -462,6 +462,95 @@ def resume(
...
+        spinner_cm = loading_spinner(processing=False)
+    with spinner_cm as progress:
+        if progress:
+            progress.add_task(description="Deleting files...", total=None)
+        jobs_info = {}
+        for fi in flows_info:
+            for ji in fi.jobs_info:
+                if ji.run_dir:
+                    jobs_info[ji.db_id] = ji
+        deleted = jc.safe_delete_files(list(jobs_info.values()))
```
davidwaroquiers, line 587: This can potentially delete files for jobs that are currently RUNNING no (actually even UPLOADED and SUBMITTED ?) ? Or maybe I missed something somewhere ? If it's the case, shoulnd't there be a confirmation in that case ?

tests/db/jobs/test_jobcontroller.py
```diff
@@ -1347,3 +1347,28 @@ def test_delete_flow(job_controller, runner, caplog):
 
     # Test deleting non-existent flow
     assert not job_controller.delete_flow("non-existent-uuid")
+
```
davidwaroquiers, line 1383: Maybe add a test of a rerun of a WAITING job, checking that it does not make it as READY ?

tests/db/cli/test_flow.py
```diff
@@ -285,3 +285,70 @@ def test_set_store(job_controller, runner, one_job, run_check_cli):
...
+    ]
+    run_check_cli(
+        ["flow", "clean", "-v", "-fid", two_flows_four_jobs[0].uuid],
+        required_out=required_out_1,
+        cli_input="y",
+    )
+    assert job_controller.count_flows() == 2
+
+    # check that the directory was deleted
+    assert not os.path.isdir(job_1_doc.run_dir)
```
davidwaroquiers: Should it check that it deleted both directories ? (not sure if it's so important ... well it is, but not sure if it is important to add the check in the test I mean)

tests/db/cli/test_flow.py
```diff
@@ -285,3 +285,70 @@ def test_set_store(job_controller, runner, one_job, run_check_cli):
...
+    run_check_cli(
+        ["flow", "delete", "-fid", "wrong_uuid"],
+        required_out="No flows matching criteria",
+        cli_input="y",
+    )
+
+    # run one of the jobs to check that the output is not deleted
+    runner = Runner()
+    runner.run_all_jobs(max_seconds=30)
+    job_1_uuid = two_flows_four_jobs[0].jobs[0].uuid
```
davidwaroquiers: Why the variable for job_1_uuid and not for the other 2 and 3 ? (not really important, but I was just wondering if there was a reason)
Also, there are four jobs in total, maybe calling them job_1_1 (or similar, could be flow_1_job_1 or something) for the first job of the first flow can help see exactly what is what directly.
Anyway, not super important so I let you decide :)

tests/db/cli/test_flow.py
```diff
@@ -285,3 +285,70 @@ def test_set_store(job_controller, runner, one_job, run_check_cli):
...
+        cli_input="n",
+    )
+    assert os.path.isdir(job_2_doc.run_dir)
+    assert os.path.isdir(job_3_doc.run_dir)
+
+    os.unlink(os.path.join(job_3_doc.run_dir, "jfremote_in.json"))
+
+    required_out_3 = [
+        "Deleted execution folders of 1 Jobs",
+        "Folder was not deleted for the following jobs:",
```
davidwaroquiers, line 376: Just wondering if it would be useful to have an additional message as to why it was not deleted (basically the safe delete part that checks there is the jfremote_in.json, is there any reason it could happen ?) and mention that there is no way to force the deletion for safety reasons and that it should be done manually ? (not to be put here, but I thought about it here ...). Up to you, maybe it's a bit too much.

src/jobflow_remote/cli/job.py
```diff
@@ -1337,3 +1337,41 @@ def files_get(
...
+        exit_with_error_msg("No data matching the request")
+
+    remote_dir = job_info.run_dir
+
+    if not remote_dir:
+        exit_with_warning_msg("The remote folder has not been created yet")
+
+    with loading_spinner(processing=False) as progress:
+        progress.add_task(description="Deleting files...", total=None)
+        deleted = jc.safe_delete_files([job_info])
```
davidwaroquiers, line 1402: This can potentially delete files for jobs that are currently RUNNING no (actually even UPLOADED and SUBMITTED ?) ? Or maybe I missed something somewhere ? If it's the case, shoulnd't there be a confirmation in that case ? (same as for the clean in the flows)

davidwaroquiers (commented):
Great addition! I left a few comments, my main one being on the possible deletion of "ongoing" jobs (i.e. jobs in an UPLOADED, SUBMITTED, or RUNNING state) but maybe I missed something that prevents that.

src/jobflow_remote/cli/flow.py
gpetretto: This is a good point, but I am unsure how to best deal with this case. 
Should it try to stop the jobs being executed? I think this might be impractical
Is a confirmation just fine? But also, since the jobs are first retrieved and then files deleted, if the runner is active there might be inconsistencies between when the list is retrieved and when the files are actually removed. I would not go to the extent of locking a job though.
Maybe I am overthiking this and a confirmation is fine...

tests/db/cli/test_flow.py
gpetretto: The problem here is that the function just returns a list of deleted folders. No errors related messages. In principle there should be a warning for the specific case of missing jfremote_in.json. I can add a check for that.

src/jobflow_remote/cli/flow.py
davidwaroquiers: I think I would do one the following two options:
1. not allow to delete files for jobs that are UPLOADED, SUBMITTED, or RUNNING
2. allow it but with a confirmation (or some kind of force option ?). That makes me think that maybe we can change the "current" force option above with --yes ? It is just that we directly confirm to delete instead, but we do not force anything right ? This kind of argument (--yes or -y) is also used for other non-jfr commands like apt update or others, which would make it more "easy" to understand. What do you think ?

I'd be in favor of doing 1., it's easier and I think it makes sense to prevent the user from deleting files of a running job.

gpetretto: What do you mean exactly for not allow? Should it raise an error, print a warning or just silently skip them? Also there is still the point that unless the job is locked, there is no way of guaranteeing that the job will not switch state between the retrieval of the jobs and the actual deletion of the folders. 

For the `--yes` option, I see your point and it may be good, but the same is used in a bunch of other functions and it would maybe be more confusing to have a different one just for this command alone. If we switch I would say we do for all the commands where the meaning of `-- force` is the same as `--yes`. Indeed there are cases where the meaning is slightly different (e.g. to rerun some kind of jobs). And in that case maybe we should first deprecate the `--force`option by allowing both for a period but warn that the option is deprecated and will be switched to `--yes`

davidwaroquiers: I would print a warning mentioning for which jobs the files were not cleaned (or just mentioning that some files were not cleaned ?) and why, e.g. "Cleanup incomplete: files for certain jobs could not be removed because those jobs are in a state that prevents cleaning. Job IDs: [list]."

Regarding the guarantee of switching states between retrieval, I think it's ok, as the "cleanable" states should be downstream states. So basically, the "only" thing that can happen is to "not clean" some files that could have been. (while the opposite, cleaning some files that shouldn't have been cleaned would be more problematic)

For the --yes, let's keep that for another PR then and go through the different cli commands, properly deprecate and do this switch.

src/jobflow_remote/cli/job.py
davidwaroquiers: So here, same idea as for the cleaning of flows

tests/db/cli/test_flow.py
davidwaroquiers: Ok for me

davidwaroquiers (approved):
Great!
Ok for me, pending a few small changes.

src/jobflow_remote/cli/flow.py
gpetretto: > Regarding the guarantee of switching states between retrieval, I think it's ok, as the "cleanable" states should be downstream states. So basically, the "only" thing that can happen is to "not clean" some files that could have been. (while the opposite, cleaning some files that shouldn't have been cleaned would be more problematic)

Actually this would be possible. It is not impossible that while the list of jobs is being retrieved the job is being uploaded and so between the retrieval of the list and the start of the cleanup the job switches to the `UPLOADED` state, but for this code the state is still `CHECKED_OUT`, so here it is not even a matter of putting a warning, it is entirely unknown to the code.
To solve such a problem either the runner should be stopped or we need to put a lock on each single job in the flow. I don't see any other way.

Note that the same could actually happen when running `jf flow delete --files`. 

None of these will break the workflow execution, but can potentially leave leftover files. The only point is that it is probably highly unlikely (a user has to decide to delete files or an entire flow while it is still running and at the exact time the runner is working on it), but if we want to take a strict approach the logic should probably be revised to be more rigorous. Considering that deleting Jobs/Flows while they are locked will likely also be somewhat of a pain.

davidwaroquiers: Ok from our discussion:

- Only allow deletion of jobs in a COMPLETED/FAILED/REMOTE_ERROR state
- Have a --all-states option to delete files for jobs in any state
- If --all-states is used, only allow when the runner is stopped, unless you use --force
- "Switch" the confirmation from --force to --yes here (other cli commands in another PR)
- Provide the list of jobs for which the files were not deleted (less than X=10?) or ask for confirmation about dumping to a given file with the list of files that were not deleted (case in which we have e.g. 10000 jobs that could not be deleted)

Similar checks could probably be implemented for jf flow delete (another PR, if really needed).

A "separate" thing but somewhat related on the jf flow delete:
- Lock the flow when deleting

gpetretto: I should have done all these points.

tests/db/jobs/test_jobcontroller.py
gpetretto: done

tests/db/cli/test_flow.py
gpetretto: done

tests/db/cli/test_flow.py
gpetretto: done

tests/db/cli/test_flow.py
gpetretto: I added the check in the required that the warning is printed.

src/jobflow_remote/cli/job.py
gpetretto: It should be done here as well.

src/jobflow_remote/cli/types.py
```diff
@@ -266,6 +266,15 @@
     ),
 ]
 
+yes_opt = Annotated[
+    bool,
+    typer.Option(
+        "--yes",
+        "-y",
+        help="Sets any confirmation values to 'yes' automatically",
```
davidwaroquiers, line 274: Maybe this ?
```suggestion
        help="Automatic yes to prompts; assume 'yes' as answer to all prompts and run non-interactively.",
```
Just a very tiny detail. (this is exactly what is in the doc of apt-get)

src/jobflow_remote/cli/job.py
```diff
@@ -1367,6 +1375,17 @@ def files_delete(
     if not remote_dir:
         exit_with_warning_msg("The remote folder has not been created yet")
 
+    cleanable_states = (
+        JobState.COMPLETED,
+        JobState.FAILED,
+        JobState.REMOTE_ERROR,
+    )
+    if not all_states and job_info.state not in cleanable_states:
```
davidwaroquiers, line 1394: I don't remember if it's needed to do the same check that the runner is stopped (as in the flows) when all-states is True and force (currently there is no force option in this method) is not True.

tests/db/cli/test_flow.py
```diff
@@ -331,24 +363,88 @@ def test_clean(job_controller, two_flows_four_jobs, run_check_cli) -> None:
...
+
+    # add enough jobs to trigger
+    running_fids = []
+    for _ in range(3):
+        add_jobs = [add(1, 2) for _ in range(4)]
+        flow = Flow(add_jobs)
+        running_fids.append(flow.uuid)
+        submit_flow(flow, worker="test_local_worker")
+        for j in add_jobs:
+            # job_controller.set_job_state(state=JobState.RUNNING, job_id=j.uuid)
```
davidwaroquiers: Remove this commented line ?

tests/db/cli/test_job.py
```diff
@@ -521,6 +522,29 @@ def test_files_delete(job_controller, two_flows_four_jobs, run_check_cli) -> Non
         required_out=["The remote folder has not been created yet"],
     )
 
+    # assert job_controller.set_job_doc_properties(
+    #     {
+    #         "state": JobState.RUNNING.value,
+    #         "run_dir": "/SOME/not/EXISTING/fake/PaTh",
+    #     },
+    #     db_id="3",
+    # )
```
davidwaroquiers: Delete these commented lines ?

davidwaroquiers (approved):
Thanks!
All good to me! Just a few commented lines to remove, one suggestion for the cli doc (up to you, it's also fine as it is now), and one question on the job cleaning option (if it's needed to check that the runner is stopped as is done for the flow cleaning).

src/jobflow_remote/cli/types.py
gpetretto: I would keep as it is for the moment (for the record, copied by the help of the conda command...), as it is not entirely accurate. In case the worker has a 2FA the password will be asked, so it will not be "non-interactive". I propose to revise this description in the PR where we switch to the `--yes` option for all the commands and we have a better idea of when it gets applied.

davidwaroquiers: sure ok for me!

src/jobflow_remote/cli/job.py
gpetretto: I added the check of the runner

PR #390, opened 2025-10-31
Add CLI option to edit project files

gpetretto:
Small addition to the CLI. Command to replace some content directly in the project yaml files. 
The command `jf project edit replace xxx yyy` will replace all the occurrences of the string `xxx` with `yyy`. 
By default acts only on the current active project, but can also act on all the correctly parsed project files. It also shows a diff of the changes applied before proceeding.
Convenient for example in case of upgrade of conda environment and needing to switch many projects consistently.

I wonder if it would be worth to add a `--regex` option where the first string can also be a regex.

### DIFF
--- a/src/jobflow_remote/cli/project.py
+++ b/src/jobflow_remote/cli/project.py
@@ -20,6 +20,7 @@ from jobflow_remote.cli.types import (
 from jobflow_remote.cli.utils import (
     SerializeFileFormat,
     check_incompatible_opt,
+    check_stopped_runner,
     exit_with_error_msg,
     exit_with_warning_msg,
     get_config_manager,
@@ -28,7 +29,7 @@ from jobflow_remote.cli.utils import (
     out_console,
     print_success_msg,
 )
-from jobflow_remote.config import ConfigError, ConfigManager
+from jobflow_remote.config import ConfigError, ConfigManager, Project
 from jobflow_remote.config.helper import (
     check_jobstore,
     check_queue_store,
@@ -419,13 +420,27 @@ def replace(
         ),
     ] = False,
     force: force_opt = False,
+    no_backup: Annotated[
+        bool,
+        typer.Option(
+            "--no-backup",
+            "-nb",
+            help="Avoid creating a backup copy of the project file",
+        ),
+    ] = False,
 ) -> None:
     """
     Replace a string in one or more project files.
 
     This command performs a text replacement in the YAML project files,
     replacing all instances of old_string with new_string.
+    By default, creates a backup of the original file.
     """
+    import json
+
+    import tomlkit
+    from ruamel.yaml import YAML
+
     cm = get_config_manager()
 
     # Determine which projects to process
@@ -439,6 +454,8 @@ def replace(
     else:
         project_data = cm.get_project_data()
         projects_to_process = [project_data.project.name]
+
+        check_stopped_runner(error=True)
 
     modified_count = 0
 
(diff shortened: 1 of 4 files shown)


src/jobflow_remote/cli/project.py
```diff
@@ -387,3 +391,114 @@ def list_worker(
...
+    new_string: Annotated[str, typer.Argument(help="String to replace with")],
+    all_projects: Annotated[
+        bool,
+        typer.Option(
+            "--all",
+            "-a",
+            help="Apply replacement to all project files",
+        ),
+    ] = False,
+    force: force_opt = False,
```
davidwaroquiers, line 422: Probably to be changed to always_confirm / --yes / -y when we do that PR (or would you start this option directly with --yes ?)

davidwaroquiers: I'm wondering if it would be better to always ask for confirmation here. Or have a backup of the files. Because if someone does this replace cli by "error" using --force then he may end up with a modified config that might be difficult to revert back to the right one.

src/jobflow_remote/cli/project.py
```diff
@@ -387,3 +391,114 @@ def list_worker(
...
+) -> None:
+    """
+    Replace a string in one or more project files.
+
+    This command performs a text replacement in the YAML project files,
+    replacing all instances of old_string with new_string.
+    """
+    cm = get_config_manager()
```
davidwaroquiers, line 444: Shouldn't this be done only when the runner is stopped maybe ? That would "force" the user to do it somehow. It's of course not "strictly" necessary, but at least it is sure that the changes will be picked up. Maybe a bit tricky for multiple projects though ?

src/jobflow_remote/cli/project.py
```diff
@@ -387,3 +391,114 @@ def list_worker(
...
+                            title="[bold]Changes Preview[/bold]",
+                            border_style="blue",
+                        )
+                    )
+
+                    if not Confirm.ask(f"Apply these changes to {project_name}?"):
+                        continue
+
+                # Write back the modified content
+                filepath.write_text(modified_content)
```
davidwaroquiers, line 529: Before writing back, shouldn't we check that the project yaml file is valid ? The old_string could be one of the keys and be replaced by something that makes the entire project invalid.

davidwaroquiers (commented):
Thanks for that addition! I added a few comments.

I think it could be nice to have something to create/edit projects configurations "automatically". I've seen this typer-pydantic-config project but it's still early stages I think... Anyway, I wouldn't go beyond much more of what you did right now.

src/jobflow_remote/cli/project.py
gpetretto: Now it creates a backup of the file by default. Can be disabled with an option.

src/jobflow_remote/cli/project.py
gpetretto: good point, but a bit tricky to handle. I have made the check for the stopped runner in the case the modified project is the current one, but not if all projects are selected. I also modified the final message to mention that the runner should be restarted for the changes to be registered.

src/jobflow_remote/cli/project.py
gpetretto: right. I added this check.
It is not done if `--force`, as I have put it together with the ask of the confirmation. Do you think the check should be done anyway? There is also the backup now in case of big issues with the modified file.
Also one might want to modify a project in the projects folder that is for a different version of jobflow-remote and thus will be recognized as invalid, but it may still be for the other version. So I think it should still be possible to make the change even if the project is invalid. Assuming the users know what they are doing.

src/jobflow_remote/cli/project.py
davidwaroquiers: great

src/jobflow_remote/cli/project.py
davidwaroquiers: great!

src/jobflow_remote/cli/project.py
davidwaroquiers: great!

davidwaroquiers (approved):
Great, to me this can be merged!

PR #392, opened 2025-11-01
Fix for batch submission

### LINKED ISSUE #391: More than the `max_jobs` number of parallel batch jobs are submitted, after some time
We (@Naisargi-Goyal and I) are running a parallel batch job with Jobflow-Remote where we request a maximum of one 128-node Slurm allocation and run 128 jobs concurrently on the allocation. This has been going very smoothly! However, one issue we have continued to run into is that at some point during the run --- for reasons we have not yet been able to pinpoint --- a second Slurm job will be submitted to the queue even though the `max_jobs` is set to 1. It does not happen right away, and we do not do anything specific to kickstart it, but it happens. We have made sure to only run Jobflow-Remote on the same login node, just as a precaution, so it is not that. Running `ps -u <Name>` shows the correct number of `jf` processes. We did a fully clean install of everything before this; we deleted `~/.jfremote`, cleared our MongoDB, and removed all pre-existing folders, so it was a clean slate. Still, a second job sometimes gets launched for reasons unknown to us.

I know this is not a lot of information to go off of, but I'm reporting it in case others see it and in case we get more information. Our YAML is below. If there is information in logs that might be helpful here, let us know.

```yaml
name: cms
workers:
  basic_vasp:
    type: local
    scheduler_type: slurm
    work_dir: /global/cfs/cdirs/m5034/ngoyal23/jfr
    pre_run: |
      source ~/.bashrc
      module load conda
      conda activate cms
      module load vasp/6.5.1_gpu
      export OMP_NUM_THREADS=8
      export OMP_PLACES=threads
      export OMP_PROC_BIND=spread
    timeout_execute: 60
    max_jobs: 1
    resources:
      nodes: 128
      gres: gpu:4
      time: 24:00:00
      account: m5039_g
      qos: regular
      constraint: gpu
      qverbatim: "#SBATCH --reservation=<REDACTED>"
    batch:
      jobs_handle_dir: /global/cfs/cdirs/m5034/ngoyal23/jfr_handles
      work_dir: /global/cfs/cdirs/m5034/ngoyal23/jfr
      parallel_jobs: 128
queue:
  store:
    <REDACTED>
exec_config: {}
```

This is with commit 99d4bd804344fc35cb08415f28ef7027392493fc.

davidwaroquiers:
Added test for batch submission.
Fixed batch submission.

Should fix #391


gpetretto:
This should fix the issue, but I have two general comments:
1) I wonder if this is the best solution, because if the command fails, it seems that basically everything that happens afterwards in `batch_update_status` and `submit_batch_processes` is pointless, isn't it? It is not such a big burden, but maybe it can be that instead if the slurm command fails `batch_update_status` returns `None` and `submit_batch_processes` is also skipped?
Or is there any action that is still worth performing?
2) Probably for another PR, but maybe we could somehow make these errors more accessible to the user. Imagine the runner never manages to get the list of slurm jobs, the only way for the user to know is to go the runner logs. The batch collections could contain an `error` field with some information. However I don't think it will be so trivial, because there are some potential errors, so it is not clear if these should all put a message there and when these messages should be cleaned up as the correct behaviour is restored.

davidwaroquiers:
> This should fix the issue, but I have two general comments:
> 
> 1. I wonder if this is the best solution, because if the command fails, it seems that basically everything that happens afterwards in `batch_update_status` and `submit_batch_processes` is pointless, isn't it? It is not such a big burden, but maybe it can be that instead if the slurm command fails `batch_update_status` returns `None` and `submit_batch_processes` is also skipped?
>    Or is there any action that is still worth performing?

Good point, I changed it. We might have something worth doing if we add the "last_ping_time" from running jobs. Let's keep it this way for now.

> 2. Probably for another PR, but maybe we could somehow make these errors more accessible to the user. Imagine the runner never manages to get the list of slurm jobs, the only way for the user to know is to go the runner logs. The batch collections could contain an `error` field with some information. However I don't think it will be so trivial, because there are some potential errors, so it is not clear if these should all put a message there and when these messages should be cleaned up as the correct behaviour is restored.

Good idea, for another PR indeed.

gpetretto (approved):
looks good to me

PR #396, opened 2025-11-07
Update `set_run_config` to allow multiple changes

gpetretto:
Currently, if multiple calls of `set_run_config` apply to the same Job the `JobConfig` is overwritten every time and thus only the last one is taken into account. 
Switch to use dict_mod updates so that each all updates will be considered. 

Requires https://github.com/materialsproject/jobflow/pull/817 to be merged (and a new release of jobflow).

### DIFF
--- a/tests/db/config/test_jobconfig.py
+++ b/tests/db/config/test_jobconfig.py
@@ -5,46 +5,112 @@ def test_set_run_config(job_controller):
     from jobflow_remote.config.jobconfig import set_run_config
     from jobflow_remote.testing import add
 
-    j1 = add(1, 2)
-    flow1 = Flow([j1])
-    submit_flow(flow1, worker="test_local_worker")
+    for overwrite in (False, True):
+        j1 = add(1, 2)
+        flow1 = Flow([j1])
+        submit_flow(flow1, worker="test_local_worker")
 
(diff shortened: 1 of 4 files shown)


tests/db/config/test_jobconfig.py
```diff
@@ -5,46 +5,112 @@ def test_set_run_config(job_controller):
...
-    assert doc1.worker == "test_local_worker_2"
-    assert doc1.priority == 5
-    assert doc1.exec_config == "test"
-    assert doc1.resources == {"test_res": 1}
-
-    assert doc2.worker == "test_local_worker"
-    assert doc2.priority == 0
-    assert doc2.exec_config is None
-    assert doc2.resources is None
+    for overwrite in (False, True):
```
davidwaroquiers: Would it make sense to do it as a pytest.mark.parametrize option ? I guess if it is done with the parametrize option, we directly know from which overwrite option(s) the test fails (if it fails), otherwise we have to change the test/put some print/...

davidwaroquiers (approved):
All good to me!
Maybe just overwrite as a pytest.mark.parametrize option ?
I let you decide anyway

tests/db/config/test_jobconfig.py
gpetretto: done

PR #399, opened 2025-11-11
Report option for jf flow info

gpetretto:
For big Flows `jf flow info` is difficult to use because too many jobs are present. Here a `--report` option is introduced that just gives a summary of the jobs states and execution progress. I initially thought of adding this as an option for `jf flow report` by allowing to pass the flow id, but in the end I would have needed to have all the same options as in `jf flow info` and made the thing much more complicated both at the API and code level.

Here is an example of how it looks:
Screenshot 2025-11-11 alle 11 03 46

### DIFF
--- a/src/jobflow_remote/cli/formatting.py
+++ b/src/jobflow_remote/cli/formatting.py
@@ -19,7 +19,7 @@ from rich.text import Text
 from jobflow_remote.cli.utils import ReprStr, fmt_datetime, render_scope_jfr
 from jobflow_remote.jobs.state import FlowState, JobState
 from jobflow_remote.remote.data import get_job_path
-from jobflow_remote.utils.data import convert_utc_time
+from jobflow_remote.utils.data import convert_utc_time, get_h_m_s
 
 if TYPE_CHECKING:
     from packaging.version import Version
@@ -88,6 +88,7 @@ def format_run_time(ji: JobInfo) -> str:
         return ""
     m, s = divmod(run_time, 60)
     h, m = divmod(m, 60)
+    h, m, s = get_h_m_s(run_time)
     return prefix + f"{h:g}:{m:02g}"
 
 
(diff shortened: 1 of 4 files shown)


davidwaroquiers (approved):
Thanks a lot for this! Will be very useful for large flows!
Only question I would have is: should this be the default maybe for large flows ? Or maybe if the flow has more than X jobs (which should already be known in the flows_info) - but then we'd need an option to actually get the list if we really want to.
What do you think ?
If you think it's not worth let's just keep it as it is now. It's already a great addition!

src/jobflow_remote/cli/formatting.py
```diff
@@ -579,6 +579,169 @@ def get_flow_report_components(report: FlowsReport) -> list[RenderableType]:
...
+
+    # If flow is completed, calculate total time
+    if flow_info.state == FlowState.COMPLETED:
+        end_time = flow_info.updated_on
+        header_table.add_row(
+            "Complete Time",
+            convert_utc_time(end_time).strftime(fmt_datetime) + f" [{time.tzname[0]}]",
+        )
+        total_time = (end_time - start_time).total_seconds()
+        hours, remainder = divmod(total_time, 3600)
```
davidwaroquiers: Maybe make a utility function ?

davidwaroquiers (approved):
Great

src/jobflow_remote/cli/formatting.py
gpetretto: you mean here, or a generic one in the `util` module?

davidwaroquiers: I don't know, is there somewhere else where such a thing is also done ? If so I would say maybe in util ? If not, maybe here ? Anyway, it's not super important I would say. If you think it can stay like this, is fine with me too.

gpetretto: I added a function in util and used it in a few other places

PR #400, opened 2025-11-11
Replace --force with --yes

gpetretto:
As discussed previously, we replace the `--force` option with `--yes` to skip confirmations. The `--force` option remains for actions that have a chance of disrupting the system.

The deprecated option will need to be removed in the future.

I did not modify the description of the `--yes` option for the moment.


davidwaroquiers (approved):
All good to me! I did not check if there were any --force left (corresponding to a "confirm all") but I guess you did.
Can be merged.
Thanks!

PR #401, opened 2025-11-12
fix upgrade to 1.0

gpetretto:
The `previous_state` attribute was not considered in the upgrade procedure, but can also be set as `TERMINATED`.

Fix one issue with the print of conditions for failure. Also, if not present the procedure is interrupted.


davidwaroquiers (approved):
Thanks for spotting and fixing these issues!
Looks good to me!

PR #402, opened 2025-11-12
Setting up uv to do lowest-direct and highest deps resolutions tests.

davidwaroquiers:


gpetretto (approved):
Fine with me. 
I would say that we monitor how long it takes to run the tests with all these python versions and consider reducing at least some of them (e.g. similarly to what is done in qtk) in the future if it is too slow.

PR #404, opened 2025-11-12
fix upgrade to 1.0

gpetretto:
Sorry, I am not sure what happened, but I probably messed up something with git, the branch was reset and the previous PR closed automatically.

The previous_state attribute was not considered in the upgrade procedure, but can also be set as TERMINATED.

Fix one issue with the print of conditions for failure. Also, if not present the procedure is interrupted.

In addition to the previous PR, allow a condition on the batches to be skipped with a `--force` option and a bit more detailed information about the condition if batch documents are present.


davidwaroquiers (approved):
LGTM!
Thanks!

PR #407, opened 2025-11-19
Add warn option

FabiPi3:
While creating our config files, quite often we introduce unwanted issues. Given that jobflow-remote is not able to parse those files, there is only a generic error message:

```
An Error occurred during the command execution: ConfigError The selected project my_project does not exist or could not be parsed correctly
```

We found it would be quite useful to see the full error message. I propose a change to enable the printing of the warning message.

### DIFF
--- a/src/jobflow_remote/cli/project.py
+++ b/src/jobflow_remote/cli/project.py
@@ -81,9 +81,7 @@ def list_projects(
     for pn in sorted(full_project_list):
         out_console.print(f" - {pn}", style="green" if pn == project_name else None)
 
-    not_parsed_projects = set(full_project_list).difference(cm.projects_data) | set(
-        erroneous_files
-    )
+    not_parsed_projects = set(full_project_list).difference(cm.projects_data)
     if not_parsed_projects:
         out_console.print(
             "The following project names exist in files in the project folder, "
@@ -91,9 +89,17 @@ def list_projects(
             f"{', '.join(not_parsed_projects)}.",
             style="yellow",
         )
+    if erroneous_files:
+        out_console.print(
+            "The following files exist in the project folder, "
+            "but could not properly parsed as projects: "
+            f"{', '.join(erroneous_files)}.",
+            style="yellow",
+        )
+    if (not_parsed_projects or erroneous_files) and not warn:
         from jobflow_remote import SETTINGS
 
-        if SETTINGS.cli_suggestions and not warn:
+        if SETTINGS.cli_suggestions:
             out_console.print(
                 "Run the command with -w option to see the parsing errors",
                 style="yellow",
--- a/src/jobflow_remote/cli/project.py
+++ b/src/jobflow_remote/cli/project.py
@@ -70,7 +70,9 @@ def list_projects(
     except ConfigError:
         pass
 
-    full_project_list = cm.project_names_from_files()
+    full_project_list, erroneous_files = cm.project_names_from_files(
+        suppress_warnings=True
+    )
 
     if not full_project_list:
         exit_with_warning_msg(f"No project available in {cm.projects_folder}")
@@ -79,7 +81,9 @@ def list_projects(
     for pn in sorted(full_project_list):
         out_console.print(f" - {pn}", style="green" if pn == project_name else None)
 
-    not_parsed_projects = set(full_project_list).difference(cm.projects_data)
+    not_parsed_projects = set(full_project_list).difference(cm.projects_data) | set(
+        erroneous_files
+    )
     if not_parsed_projects:
         out_console.print(
             "The following project names exist in files in the project folder, "
@@ -89,7 +93,7 @@ def list_projects(
         )
         from jobflow_remote import SETTINGS
 
-        if SETTINGS.cli_suggestions:
+        if SETTINGS.cli_suggestions and not warn:
             out_console.print(
                 "Run the command with -w option to see the parsing errors",
                 style="yellow",
--- a/src/jobflow_remote/cli/utils.py
+++ b/src/jobflow_remote/cli/utils.py
@@ -22,7 +22,7 @@ from rich.tree import Tree
 from typer.core import TyperCommand, TyperGroup
 
 from jobflow_remote import ConfigManager, JobController
-from jobflow_remote.config.base import ProjectUndefinedError
+from jobflow_remote.config.base import ProjectParsingError, ProjectUndefinedError
 from jobflow_remote.jobs.daemon import DaemonError, DaemonManager, DaemonStatus
 
 if TYPE_CHECKING:
@@ -314,6 +314,11 @@ def cli_error_handler(func):
             exit_with_error_msg(
                 "The active project could not be determined and it is required to execute this command. Please "
                 "check the formatting of your YAML."
+            )
+        except ProjectParsingError:
+            exit_with_error_msg(
+                "The active projects config file could not be parsed. Please check the formatting of your YAML. "
+                "You can use the command 'jf project list -w' to get the parsing errors."
             )
         except Exception as e:
             from jobflow_remote import SETTINGS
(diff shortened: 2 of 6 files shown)


gpetretto:
Hi @FabiPi3,
thanks for this PR. Overall I am fine with the change. I have just one request and a couple of comments
* Can you rename this to something like `--warn-project`? `--warn` is quite generic and I would like to keep it in case we want to have a global flag to handle all the warnings
* Consider that this will print the warning for all the projects in your `~/.jfremote` folder that fail validation. I suppose that this does not happen often, but I just wanted to point it out. Changing this will require updating the configuration manager and is probably not so straightforward.
* Just mentioning that anyway this will be more or less equivalent to run `jf project list -w`.

FabiPi3:
Hi @gpetretto,

thanks for mentioning the `jf project list -w` command. I wasn't aware of it. I think this would fit our use case as well. Maybe we could add a hint in the error message to this command? What do you prefer?

While testing it, I found another subtle issue. Given that the `jf project list` command tries to find the default project first, and you are actually modifying the default projects file, you will see the warning message twice which might be confusing to the user.

gpetretto:
> thanks for mentioning the `jf project list -w` command. I wasn't aware of it. I think this would fit our use case as well. Maybe we could add a hint in the error message to this command? What do you prefer?

If you think that the additional option for `jf` will be helpful it is still fine to add. If you instead think that the current `jf project list -w` is enough for your use case it can also stay like this. I totally agree that in both cases the command should be suggested in the output when an error like this happens.
 
> While testing it, I found another subtle issue. Given that the `jf project list` command tries to find the default project first, and you are actually modifying the default projects file, you will see the warning message twice which might be confusing to the user.

I tried to reproduce this, but with a malformed yaml file in my `~/.jfremote`, even if it that project is selected by setting the environment variable `jfremote_project`, I still see only one warning related to it when I run `jf project list -w`. Can you give more details about the conditions under which you get the double warning?

FabiPi3:
> If you think that the additional option for `jf` will be helpful it is still fine to add. If you instead think that the current `jf project list -w` is enough for your use case it can also stay like this. I totally agree that in both cases the command should be suggested in the output when an error like this happens.

I renamed the option. Concerning the suggestion in the output, I am not sure where to put it. I assume it should be in the cli script and not in the config/manager.py file. I could wrap the call `cm.get_project()` in a try/except block, but I would need to do this very often as this method is called quite often in the CLI scripts. Do you have an idea?

> I tried to reproduce this, but with a malformed yaml file in my `~/.jfremote`, even if it that project is selected by setting the environment variable `jfremote_project`, I still see only one warning related to it when I run `jf project list -w`. Can you give more details about the conditions under which you get the double warning?

Bildschirmfoto 2025-11-26 um 22 31 07

Here is the picture of the output. Basically I added a tab in the yaml file which is not allowed. I have a `~/.jfremote.yaml` file which contains:
```yaml
project: exwoflow
```
Else I thought I have very standard conditions. I am using this branch here. What else do you need to know?

gpetretto:
> I renamed the option. Concerning the suggestion in the output, I am not sure where to put it. I assume it should be in the cli script and not in the config/manager.py file. I could wrap the call `cm.get_project()` in a try/except block, but I would need to do this very often as this method is called quite often in the CLI scripts. Do you have an idea?

Thanks for thinking this through. You are right that this would better sit in the CLI rahter than in manager.py. 
A solution here could be the following:
* create a subclass of `ConfigError` (e.g. ParsingProjectError)
* raise it here instead of ConfigError: https://github.com/Matgenix/jobflow-remote/blob/a784351417e06fdb9ed2d3a7b3f8620b0ab1897e/src/jobflow_remote/config/manager.py#L168 
* handle it here, similarly to `ProjectUndefinedError`, taking the error message and appending something like "use `jf project list -w` to get the parsing errors" : https://github.com/Matgenix/jobflow-remote/blob/a784351417e06fdb9ed2d3a7b3f8620b0ab1897e/src/jobflow_remote/cli/utils.py#L313

I think this should cover all the use cases, but it has to be tested.
Anyway, if this does not work or proves to be too involved, I am also fine just adding directly to the error message in manager.py something like: "to get the parsing errors run `jf project list -w` or instantiate the ConfigManager with the `warn` option activated"


> Here is the picture of the output. Basically I added a tab in the yaml file which is not allowed. I have a `~/.jfremote.yaml` file which contains:
> 
> ```yaml
> project: exwoflow
> ```
> 
> Else I thought I have very standard conditions. I am using this branch here. What else do you need to know?

I see, I only tested a case where the Project validation failed (e.g. adding a keyword that is not acceptable for the `Project` configuration), instead here the problem is if the yaml is malformed. So the warning is printed from two different points:
https://github.com/Matgenix/jobflow-remote/blob/a784351417e06fdb9ed2d3a7b3f8620b0ab1897e/src/jobflow_remote/config/manager.py#L112
https://github.com/Matgenix/jobflow-remote/blob/a784351417e06fdb9ed2d3a7b3f8620b0ab1897e/src/jobflow_remote/config/manager.py#L302
A quick fix that I can think about would be to add an argument like `suppress_warnings` to `project_names_from_files` `False` by default and set it to `True` when calling it from `jf project list`. But since this seems a bit hackish and `project_names_from_files` is only used there, if you have any better solution it is fine with me.

FabiPi3:
Hi @gpetretto,

I tried to incorporate your ideas and did a few other fixes. Locally everything seems to be fine for me. Let me know what you think.

src/jobflow_remote/cli/project.py
```diff
@@ -70,7 +70,9 @@ def list_projects(
     except ConfigError:
         pass
 
-    full_project_list = cm.project_names_from_files()
+    full_project_list, erroneous_files = cm.project_names_from_files(
+        suppress_warnings=True
+    )
```
gpetretto, line 75: This is an interesting suggestion. However, I would propose that erroneous_files contains the actual file names and if erroneous_files is not empty a separate message will be printed stating that these files could not be parsed. What do you think?

src/jobflow_remote/cli/utils.py
```diff
@@ -315,6 +315,11 @@ def wrapper(*args, **kwargs):
                 "The active project could not be determined and it is required to execute this command. Please "
                 "check the formatting of your YAML."
             )
+        except ProjectParsingError:
+            exit_with_error_msg(
+                "The active projects config file could not be parsed. Please check the formatting of your YAML. "
```
gpetretto: If I am not wrong the code will arrive here even if the project name is misspelled in the `jfremote_project` variable (or with the `-p` option in the CLI). I think the error message could reflect a broader range of cases. Maybe it can use directly the error message coming from the error:
```python
except ProjectParsingError as exc:
    msg = f"{getattr(exc, 'message', str(exc))} You can use the command 'jf project list -w' to get the parsing errors."
```
It is fine to improve the message raised in ProjectParsingError, if you think it is not clear enough. Also note that yaml is not the only format allowed for the configuration file. Although I suppose is the most diffused, I would not specify the format in the message.

gpetretto (commented):
Sorry for the delay and thanks for the updates and improvements. I have left a few minor comments

src/jobflow_remote/cli/project.py
FabiPi3: Ok, I added a separate error message.

src/jobflow_remote/cli/utils.py
FabiPi3: I agree. But note that I just copied from the case above (`ProjectUndefinedError`) where it also says `YAML`. Would the same apply there as well?

src/jobflow_remote/cli/project.py
gpetretto: thanks

src/jobflow_remote/cli/utils.py
gpetretto: good point. Sorry. I will probably fix this in another PR.

gpetretto (approved):

PR #412, opened 2025-11-24
Fix job names in tests to prevent some random test failures

davidwaroquiers:
Renamed add1, add2, ... to add_job1 etc... as well as f1 to flow_1 to prevent random clashes in the tests (e.g. add1 could be found in the uuid while it is supposed to be excluded).


gpetretto (approved):
Thanks! This was indeed quite annoying.

PR #413, opened 2025-11-26
Allow jf job set-state for many Jobs at once

gpetretto:
Following [this discussion](https://matsci.org/t/create-jobs-immune-to-stop-children/65765/10), increase the options for `jf job set-state` to modify multiple Jobs with a single command.

Slight update of `set_job_doc_properties` and cleanup of some other job related methods in JobController.

### DIFF
--- a/src/jobflow_remote/cli/job.py
+++ b/src/jobflow_remote/cli/job.py
@@ -68,7 +68,6 @@ from jobflow_remote.cli.utils import (
     hide_progress,
     loading_spinner,
     out_console,
-    print_success_msg,
     str_to_dict,
 )
 from jobflow_remote.jobs.report import JobsReport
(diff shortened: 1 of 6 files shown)


davidwaroquiers (approved):
Looks good to me!
These mypy checks are indeed a bit stupid ...
I did not test but if in MongoLock the self.locked_document is typed to dict | None in the __init__, would that change anything ?

gpetretto:
> I did not test but if in MongoLock the self.locked_document is typed to dict | None in the **init**, would that change anything ?

Good call. I assumed it was already like that and that mypy was just not able to follow the logic through the context manager. Setting it seems to have done the trick. However this triggered a few other typing errors (that were not even directly related to `locked_document`), so I had to make a few more updates. Can you make a quick check that everything is reasonable for you with the last changes?

davidwaroquiers (approved):
Sounds good to me!
Great

PR #416, opened 2025-11-27
Runner updates

gpetretto:
In light of improving the management of the Runner from the user:
* rewire `jf runner stop` to `jf runner shutdown`. It is still possible to "stop" the runner either with the hidden command `jf runner stop-processes` or with the python API. Internally there is thus still a difference between shutdown and stop. Only the CLI will see them as equivalent. This may cause confusion due to the clash of terms, but should avoid having to deal with both states.
* Improve (hopefully) the error message related to the runner being already active from the DB. I believe several errors may be related to the user not reading or not understanding properly the error message, which may be clear for us, but not for a first time user. The idea here is that the massage focuses only on specifying the relevant information and points to documentation resources about the meaning of the problem and how to deal with it. If this is fine I will add another section to the Runner page that targets explicitly this issue in a concise way (I assume most people will not survive reading all that stuff)

### DIFF
--- a/doc/source/user/runner.rst
+++ b/doc/source/user/runner.rst
@@ -92,8 +92,8 @@ or the command::
     jf runner shutdown
 
 relies on Supervisor to send a ``SIGTERM`` signal (a termination signal that allows
-the process to exit cleanly) to all the ``Runner`` processes. After this also the
-Supervisor process will be stopped. The two commands are equivalent and will result in
+the process to exit cleanly) to all the ``Runner`` processes. After this the
+Supervisor process will also be stopped. The two commands are equivalent and will result in
 a Runner in the ``shut_down`` state.
 
 Unless the ``--wait`` option is specified, the completion of the command will not imply
@@ -106,7 +106,7 @@ that all the ``Runner`` processes have been terminated.
 .. note::
     In older versions of jobflow-remote the ``jf runner stop`` command used to stop
     only the ``Runner`` processes, but leaving the Supervisor process active. Since version
-    1.0 this has been removed from the CLI to simply the handling of the runner.
+    1.0 this has been removed from the CLI to simplify the handling of the runner.
     If needed, the option is still available through the python API.
 
 Kill
--- a/doc/source/user/runner.rst
+++ b/doc/source/user/runner.rst
@@ -78,38 +78,36 @@ the steps 3 and 4.
     the project configuration is changed the ``Runner`` needed the runner needs
     to be restarted.
 
-.. _runner stop:
+.. _runner stop_shutdown:
 
-Stop
-----
+Stop - Shutdown
+---------------
 
 Executing the stop command::
 
     jf runner stop
 
+or the command::
+
+    jf runner shutdown
+
 relies on Supervisor to send a ``SIGTERM`` signal (a termination signal that allows
-the process to exit cleanly) to all the ``Runner`` processes.
-In this case the supervisor process will remain active. Unless the ``--wait`` option
-is specified, the completion of the command will not imply that all the ``Runner``
-processes have been terminated.
+the process to exit cleanly) to all the ``Runner`` processes. After this also the
+Supervisor process will be stopped. The two commands are equivalent and will result in
+a Runner in the ``shut_down`` state.
+
+Unless the ``--wait`` option is specified, the completion of the command will not imply
+that all the ``Runner`` processes have been terminated.
 
 .. warning::
     The ``Runner`` is designed to recognize the signal and **wait for the completion of
     the action being performed**, before actually exiting.
 
 .. note::
-    Since the supervisor process remains active, when starting the runner again after
-    a stop it is not possible to switch from a single process to a split configuration
-    or the other way round. It is necessary to shut down the whole daemon in that case.
-
-Shutdown
---------
-Shutting down the runner with the command::
-
-    jf runner shutdown
-
-is equivalent to the :ref:`runner stop`, except that also the Supervisor process will
-be stopped.
+    In older versions of jobflow-remote the ``jf runner stop`` command used to stop
+    only the ``Runner`` processes, but leaving the Supervisor process active. Since version
+    1.0 this has been removed from the CLI to simply the handling of the runner.
+    If needed, the option is still available through the python API.
 
 Kill
 ----
--- a/src/jobflow_remote/jobs/daemon.py
+++ b/src/jobflow_remote/jobs/daemon.py
@@ -133,6 +133,7 @@ class DaemonStatus(Enum):
     """
 
     SHUT_DOWN = "SHUT_DOWN"
+    SHUTTING_DOWN = "SHUTTING_DOWN"
     STOPPED = "STOPPED"
     STOPPING = "STOPPING"
     PARTIALLY_RUNNING = "PARTIALLY_RUNNING"
@@ -343,7 +344,7 @@ class DaemonManager:
 
         return running
 
-    def check_status(self) -> DaemonStatus:
+    def check_status(self, raise_on_shutdown: bool = True) -> DaemonStatus:
         """
         Get the current status of the daemon based on the state of the supervisord
         process and the running processes.
@@ -384,9 +385,11 @@ class DaemonManager:
                 return DaemonStatus.SHUT_DOWN
 
             if exc.faultString == "SHUTDOWN_STATE":
-                raise DaemonError(
-                    "The daemon is likely shutting down and the actual state cannot be determined"
-                ) from exc
+                if raise_on_shutdown:
+                    raise DaemonError(
+                        "The daemon is likely shutting down and the actual state cannot be determined"
+                    ) from exc
+                return DaemonStatus.SHUTTING_DOWN
             raise
         if not proc_info:
             raise DaemonError(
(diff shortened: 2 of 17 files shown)


doc/source/user/runner.rst
```diff
@@ -78,38 +78,36 @@ the steps 3 and 4.
...
---------
-Shutting down the runner with the command::
-
-    jf runner shutdown
-
-is equivalent to the :ref:`runner stop`, except that also the Supervisor process will
-be stopped.
+    In older versions of jobflow-remote the ``jf runner stop`` command used to stop
+    only the ``Runner`` processes, but leaving the Supervisor process active. Since version
+    1.0 this has been removed from the CLI to simply the handling of the runner.
```
davidwaroquiers:
```suggestion
    1.0 this has been removed from the CLI to simplify the handling of the runner.
```

src/jobflow_remote/jobs/daemon.py
```diff
@@ -861,6 +865,19 @@ def shut_down(self, raise_on_error: bool = False) -> bool:
...
+                    try:
+                        current_status = self.check_status(raise_on_shutdown=False)
+                    except Exception as exc:
+                        if raise_on_error:
+                            raise DaemonError(
+                                "Error while waiting for the daemon to shut down"
+                            ) from exc
+                        return False
+                    if current_status == DaemonStatus.SHUT_DOWN:
+                        break
```
davidwaroquiers, line 879: Should there be (optionally) some kind of timeout? Is that useful anywhere ?

src/jobflow_remote/jobs/daemon.py
```diff
@@ -1134,25 +1151,31 @@ def _check_running_runner(
...
-        error = (
-            "A daemon runner process may be running on a different machine.\n"
-            "Here is the information retrieved from the database:\n"
-            f"- hostname: {db_data['hostname']}\n"
-            f"- project_name: {db_data['project_name']}\n"
-            f"- start_time: {db_data['start_time']}\n"
-            f"- last_pinged: {db_data['last_pinged']}\n"
-            f"- daemon_dir: {db_data['daemon_dir']}\n"
-            f"- user: {db_data['user']}"
+        error = "A daemon runner process associated to this database may be already running.\n"
```
davidwaroquiers, line 1158: Maybe to this project ?
```suggestion
        error = "A daemon runner process associated to this project may be already running.\n"
```

doc/source/user/runner.rst
```diff
@@ -78,38 +78,36 @@ the steps 3 and 4.
...
+or the command::
+
+    jf runner shutdown
+
 relies on Supervisor to send a ``SIGTERM`` signal (a termination signal that allows
-the process to exit cleanly) to all the ``Runner`` processes.
-In this case the supervisor process will remain active. Unless the ``--wait`` option
-is specified, the completion of the command will not imply that all the ``Runner``
-processes have been terminated.
+the process to exit cleanly) to all the ``Runner`` processes. After this also the
```
davidwaroquiers:
```suggestion
the process to exit cleanly) to all the ``Runner`` processes. After this the Supervisor process will also be stopped
```

davidwaroquiers (approved):
Looks good to me!
Thanks!

src/jobflow_remote/jobs/daemon.py
gpetretto: Good question, but how long should this timeout be? In principle there is no upper limit. And what should happen if the timeout is reached? Should it consider the runner shut down or not?
Just to clarify, I mainly added the `--wait` option since it existed `jf runner stop` and I did not want to break the API. If we decide that having it is not a good idea we can also deprecate it and remove the functionality (at least from the CLI)

src/jobflow_remote/jobs/daemon.py
gpetretto: Here I picked the word specifically, since if you created two projects with the same DB (as a user likely did a few days ago), it will be wrong stating that the runner is associated with the same "project". Still I agree that the "database" word may be confusing. Do you have other suggestions?

davidwaroquiers: No you are right let's keep it this way. Or maybe refer one more time to the documentation where we insist that projects should not share the same collection/database.

src/jobflow_remote/jobs/daemon.py
davidwaroquiers: I would keep the --wait option. Maybe I would not put a timeout in the end. If the user wants to get out of that he just has to do ctrl-c right ? Maybe one useful thing (not in this PR and not sure how easy that would be) could be to show dynamically what's been shutdown and what's left to shutdown after e.g. 3-5 seconds. And if there is something that takes time like a download or upload we could have a percentage (again, no idea how this would be easy or even possible to do). So for now I would just do nothing here.

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -4432,22 +4433,65 @@ def clean_running_runner(self, break_lock: bool = False) -> None:
...
+        ----------
+        data
+            A dictionary of data to be added to the ping information
 
         Returns
         -------
-        bool
-            True if the ping was successful.
+        dict
+            The content of the running_runner document, if present.
```
davidwaroquiers: Returns a tuple with the bool as previously and the dict

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -4432,22 +4433,65 @@ def clean_running_runner(self, break_lock: bool = False) -> None:
...
+            {"$set": {"running_runner.last_pinged": ping_time}},
             upsert=False,
         )
 
-        return ping_result is not None
+        data = dict(data or {})
+        data["time"] = ping_time
+        runner_pings_doc = self.auxiliary.find_one_and_update(
+            {"runner_pings": {"$exists": True}},
+            {"$push": {"runner_pings": {"$each": [data], "$slice": -50}}},
```
davidwaroquiers: Maybe I'd put 50 as a class variable. I would not go to be able to tune that in the settings and all but at least it can potentially be checked that it works, without having to ping the db 51 times. (see comment in the tests)

tests/db/jobs/test_jobcontroller.py
```diff
@@ -1112,6 +1112,15 @@ def test_running_runner(job_controller, daemon_manager, wait_daemon_started):
     assert job_controller.get_running_runner() == "NO_DOCUMENT"
 
 
+def test_runner_pings(job_controller):
+    data = {"test": "XX"}
+    job_controller.ping_running_runner(data=data)
+    job_controller.ping_running_runner(data=data)
+    assert len(job_controller.get_runner_pings()) == 2
+    job_controller.clean_runner_pings()
+    assert len(job_controller.get_runner_pings()) == 0
```
davidwaroquiers: I'd make a check that it does not go beyond the max number of pings (and also that when it reaches the max number, it is the oldest one that gets kicked). That means 50 pings, but I also suggested to put the 50 as a class variable that can then be used to dynamically change it here and make the test.

davidwaroquiers (approved):
Thanks for the updates!
Very good to me. Just a few suggestions. I let you decide in any case.

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: Thanks, I changed the output quite a few times...

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: done, but as module constant, rather than class attribute.

tests/db/jobs/test_jobcontroller.py
gpetretto: done

tests/db/jobs/test_jobcontroller.py
```diff
@@ -1112,6 +1112,22 @@ def test_running_runner(job_controller, daemon_manager, wait_daemon_started):
...
+    with monkeypatch.context() as m:
+        m.setattr(jobflow_remote.jobs.jobcontroller, "_MAX_ARCHIVED_RUNNER_PINGS", 3)
+        job_controller.ping_running_runner(data=data)
+        job_controller.ping_running_runner(data=data)
+        assert len(job_controller.get_runner_pings()) == 2
+        for _ in range(3):
+            job_controller.ping_running_runner(data=data)
+        assert len(job_controller.get_runner_pings()) == 3
+        job_controller.clean_runner_pings()
+        assert len(job_controller.get_runner_pings()) == 0
```
davidwaroquiers: Check that it is that it is the last three that are there ?

```suggestion
    import jobflow_remote.jobs.jobcontroller

    with monkeypatch.context() as m:
        m.setattr(jobflow_remote.jobs.jobcontroller, "_MAX_ARCHIVED_RUNNER_PINGS", 3)
        job_controller.ping_running_runner(data={"test": "1"})
        job_controller.ping_running_runner(data={"test": "2"})
        assert len(job_controller.get_runner_pings()) == 2
        for ii in range(3):
            job_controller.ping_running_runner(data={"test": str(ii+3)})
        pings = job_controller.get_runner_pings()
        assert len(pings) == 3
        test_numbers = [ping["test"] for ping in pings]
        assert set(test_numbers) == {"3", "4", "5"}
        job_controller.clean_runner_pings()
        assert len(job_controller.get_runner_pings()) == 0
```

davidwaroquiers (approved):
Final suggestion ?

tests/db/jobs/test_jobcontroller.py
gpetretto: done, but I also checked the order

davidwaroquiers:
Great! Thanks, this can be merged

PR #422, opened 2025-12-04
Fix dev docs

gpetretto:
The `dev/index.rst` was excluded as it was in `.gitignore`. This should fix the missing development section.

Also removed the banner stating that jobflow-remote is in beta version, as next release will be 1.0.


davidwaroquiers (approved):
All good to me, you can merge
Thanks!

PR #424, opened 2025-12-09
Do not support python 3.14.1, as it breaks networkx

gpetretto:
Attempt to fix the errors in the CI due to python 3.14.1 breaking networkx

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -71,9 +71,9 @@ jobs:
           - { os: ubuntu-latest, python: "3.11", resolution: highest }
           - { os: ubuntu-latest, python: "3.12", resolution: highest }
           - { os: ubuntu-latest, python: "3.13", resolution: highest }
-          - { os: ubuntu-latest, python: "3.14", resolution: lowest-direct }
-          - { os: macos-latest, python: "3.14", resolution: highest }
-          - { os: ubuntu-latest, python: "3.14", resolution: highest }
+          - { os: ubuntu-latest, python: "3.14.x", resolution: lowest-direct }
+          - { os: macos-latest, python: "3.14.x", resolution: highest }
+          - { os: ubuntu-latest, python: "3.14.x", resolution: highest }
 
     services:
       local_mongodb:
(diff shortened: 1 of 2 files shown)


davidwaroquiers:
Sounds good to me. The only thing maybe could be to have the name of the tests for 3.14.x to be like the others, e.g. "test (ubuntu-latest, 3.14, lowest-direct)". Not sure if it is super important anyway. Quick llm lookup suggested to add this in the job:

name: >-
    test (${{ matrix.config.os }}, 
          ${{ replace(matrix.config.python, '.x', '') }}, 
          ${{ matrix.config.resolution }})

Did not test so not sure it works of course ... and again, not sure if it is super important anyway. As you wish!

gpetretto:
I used another approach but it seems ok.
Overall, the other changes seems necessary to trigger the usage of 3.14.2, which was available anyway. However, according to the [action documentation](https://github.com/actions/setup-python/blob/83679a892e2d95755f2dac6acb0bfd1e9ac5d548/docs/advanced-usage.md#check-latest-version):
> Setting check-latest to true impacts performance as downloading Python or PyPy versions is slower than using cached versions.

I would propose to revert this change as seen soon as the default version goes to 3.14.2. I would open a draft PR with these changes removed and periodically check if that is the case. Would that be fine?

davidwaroquiers:
> I used another approach but it seems ok. Overall, the other changes seems necessary to trigger the usage of 3.14.2, which was available anyway. However, according to the [action documentation](https://github.com/actions/setup-python/blob/83679a892e2d95755f2dac6acb0bfd1e9ac5d548/docs/advanced-usage.md#check-latest-version):
> 
> > Setting check-latest to true impacts performance as downloading Python or PyPy versions is slower than using cached versions.
> 
> I would propose to revert this change as seen soon as the default version goes to 3.14.2. I would open a draft PR with these changes removed and periodically check if that is the case. Would that be fine?

Perfect!

davidwaroquiers (approved):
Perfect, thanks!
Nice solution for the names of the tests!

PR #425, opened 2025-12-10
Updates for batch jobs management

gpetretto:
Some updates and fixes for the management of batch processes:
* fix the dates to UTC in all the updates
* allow to delete batch job entries from DB
* a `jf batch info` command for details of a single batch job
* renamed `max_results` to `limit` for consistency with other Jobcontroller methods. Up for discussion if it should be the other way around

Unrelated to batch jobs: changed the error message for `ProjectUndefinedError`. Hope it is clearer.


src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -4951,30 +4951,93 @@ def _cancel_queue_process(self, job_doc: dict) -> None:
 
     @staticmethod
     def _build_query_batch(
+        process_id: str | list[str] | None = None,
+        batch_uid: str | list[str] | None = None,
         worker: str | list[str] | None = None,
         batch_state: BatchState | list[BatchState] | None = None,
```
davidwaroquiers, line 4972: Maybe interesting (here but then also to be added in the cli of course) to also have the possibility to get the list (should be only 1 in principle) that contains a given job_id ? and/or job_id+index ? and/or db_id ? (for the latter case, I think it would not be possible right now without an aggregation, except if we add the db_id to the info field of the BatchDoc or somewhere else in the BatchDoc). What do you think ?

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -4985,17 +5048,81 @@ def get_batches(
...
 
+    def delete_batches(
+        self,
+        process_id: str | list[str] | None = None,
+        batch_uid: str | list[str] | None = None,
+        worker: str | list[str] | None = None,
+        batch_state: BatchState | list[BatchState] | None = None,
+    ) -> int:
+        """
+        Delete batch processes entried from the database.
```
davidwaroquiers:
```suggestion
        Delete batch processes from the database.
```

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -4985,17 +5048,81 @@ def get_batches(
...
+        No attempt to cancel the running processes on the worker if they
+        are still running.
+
+        Parameters
+        ----------
+        process_id
+            One or more process ids of the batch processes to retrieve.
```
davidwaroquiers:
```suggestion
            One or more process ids of the batch processes to delete.
```

src/jobflow_remote/jobs/jobcontroller.py
```diff
@@ -4985,17 +5048,81 @@ def get_batches(
...
+        batch_uid
+            One or more batch uids of the batch processes to retrieve.
```
davidwaroquiers:
```suggestion
            One or more batch uids of the batch processes to delete.
```

src/jobflow_remote/cli/batch.py
```diff
@@ -67,3 +78,121 @@ def processes_list(
...
+
+    cm = get_config_manager()
+    project = cm.get_project()
+    workers = project.workers
+
+    with loading_spinner():
+        batch_processes = jc.get_batches(
+            process_id=process_id,
+            batch_uid=batch_uid,
+            limit=1,
```
davidwaroquiers: Maybe limit to 2 and check that it is not 2 ? Should not happen for batch_uid (at least very very unlikely) but with the process_id (even though very small too), it could if we have two workers and by bad luck they took the same id at some point. Not sure if it's worth anyway. I let you decide.

src/jobflow_remote/cli/batch.py
```diff
@@ -67,3 +78,121 @@ def processes_list(
...
+            "--process-id",
+            "-pid",
+            help="One or more process ids",
+        ),
+    ] = None,
+    batch_uid: Annotated[
+        str | None,
+        typer.Option(
+            "--batch-uid",
+            "-uid",
```
davidwaroquiers: Maybe "-bid" ? For jobs and flows it is "-jid" and "-fid", so it might make sense to use -bid here to follow the same convention ? Or we might think of changing -jid and -fid ? But there I would maybe expect to have some clash in some cases where you can query from a job id or a flow id (I think there are a few places where this is done right ?)

src/jobflow_remote/cli/batch.py
```diff
@@ -67,3 +78,121 @@ def processes_list(
...
+            help="One or more process ids",
```
davidwaroquiers:
```suggestion
            help="One or more batch unique ids",
```

src/jobflow_remote/cli/batch.py
```diff
@@ -67,3 +78,121 @@ def processes_list(
...
+        ),
+    ] = None,
+    worker_name: worker_name_opt = None,
+    yes_all: yes_opt = False,
+):
+    """Remove one or more batch processes from the database. No effect on the processes running on the worker."""
+
+    print(batch_state)
```
davidwaroquiers:
```suggestion
```

src/jobflow_remote/cli/batch.py
```diff
@@ -67,3 +78,121 @@ def processes_list(
...
+    if len(set(batch_state).difference([BatchState.FINISHED])) > 0 and not yes_all:
+        text = Text.from_markup(
+            "[red]This operation may remove batch processes in states other than 'FINISHED'. "
+            "This could lead to [bold]inconsistencies or data loss[/bold]. Proceed anyway?[/red]"
+        )
+
+        confirmed = Confirm.ask(text, default=False)
+        if not confirmed:
+            raise typer.Exit(0)
```
davidwaroquiers: If we have all batches that are finished, we will always get this message even though it is finished. Would it be better to put this check on the actual batch processes retrieved for deletion ?

src/jobflow_remote/cli/batch.py
```diff
@@ -67,3 +78,121 @@ def processes_list(
...
+
+    with loading_spinner():
+        n_deleted = jc.delete_batches(
+            process_id=process_id,
+            batch_uid=batch_uid,
+            worker=worker_name,
+            batch_state=batch_state,
+        )
+
+    print_success_msg(f"Operation completed. {n_deleted} batch processes deleted")
```
davidwaroquiers, line 239: Just raising the comment here but I know it is also done like this for jobs and flows but I'm wondering if we could have the n_batch_processes above be equal to n_deleted here. One way could be to get the batch_processes above (line 171), probably with a projection on the batch_uid, then delete based on the list of batch_uids obtained. Just a comment that I thought about here and that we may think of doing for jobs and flows as well. What do you think ?

davidwaroquiers (approved):
Very good as usual.
A few typos and docstring corrections + a few comments/suggestions/to discuss (maybe outside of this).
Feel free to go forward with this without addressing all the comments if you think it is not very useful.

src/jobflow_remote/cli/batch.py
gpetretto: I thought about this possibility, but if there are two than what do you do? Raise an error, so that you can't see either of them? print the info for both? add an option to decide which to show?

src/jobflow_remote/cli/batch.py
gpetretto: True, but this would require an explicit query and a check. It is probably not a big overhead. I did not do that since I thought that since one had to explicitly select the other states I suppose one already knows that there are some. However, I think I can add it.

src/jobflow_remote/cli/batch.py
gpetretto: Actually this is how it is done for the flows. For the jobs I think it does not asks for confirmation with the number.
I will do it here for the batch as well.

src/jobflow_remote/jobs/jobcontroller.py
gpetretto: it may be fine, but querying for job_id+index is usually annoying. should we add the db_id in the batch document? Where? 
https://github.com/Matgenix/jobflow-remote/blob/da70730a441bc94035be86238eef9d143fe68c94/src/jobflow_remote/jobs/jobcontroller.py#L5049
In the info (so added by the runner, but it is not guaranteed to be there)? As an additional mandatory argument and than it is added to the info? Also, why is the info of type `Any`? Shouldn't it be a dict?

src/jobflow_remote/cli/batch.py
davidwaroquiers: Probably just tell the user there is more than one and he should use something else to discriminate, e.g. use the uuid. Probably suggesting jf batch list to get the list of 2 (or possibly more) batches so that he can take the uiid from there

src/jobflow_remote/jobs/jobcontroller.py
davidwaroquiers: For querying on jobid+index, isn't that just something like "info.SOMEID.SOMEINDEX": exists: True ? (I'm on my phone sorry for the lack of correctness 😃)

For the db id, probably good to add indeed. Maybe in the info dict? (But the query is not so easy then) or just as a list at root level ? (But then the correspondance with jobid+index is not known from the inside of the BatchDox).

Info should indeed be typed as a dict.

gpetretto: > For querying on jobid+index, isn't that just something like "info.SOMEID.SOMEINDEX": exists: True ? (I'm on my phone sorry for the lack of correctness 😃)

Here I was rather referring to it being annoying for the user, but yes, it is relatively easy to code.

> For the db id, probably good to add indeed. Maybe in the info dict? (But the query is not so easy then) or just as a list at root level ? (But then the correspondance with jobid+index is not known from the inside of the BatchDox).

Now I don't know what to do for the db_id. Even adding db_id in the info, such a nested query with multiple arbitrary values as keys, even the aggregation query turns out to be quite a mess. I am not convinced that this is the best way forward. Also does this data structure for the `jobs` serve any purpose at all? Is there any use for the `state` being in there? Because for example, when you get the content of the document you get the list of Jobs and once the job is finished the state remains `RUN_FINISHED` forever, which eventually will not be the real state of the Job. Is it used for something? Can't we just get rid of this complicated structure and just call `update_job_in_batch` once when it is first picked up, consider `jobs` a list and add `[uuid, index, db_id]` to it?

davidwaroquiers: I agree that we can try to have something simpler, though I think having the state of the job within the batch is somewhat useful for debugging purposes at least. But maybe we can anyway get rid of it, by default, not getting the state of the jobs and if we do jf batch info -v for example, it would do an aggregation with the jobs collection getting the state from there ?

davidwaroquiers: Ok let's go to jobs being a list of [uuid, index, db_id] as in the FlowDoc ("ids" key). And not add the jf batch info -v command for now to get the states of the jobs (or potentially any other information from these jobs).

src/jobflow_remote/cli/batch.py
```diff
@@ -31,11 +36,43 @@
...
+        val_errors = exc.errors()
+        if (
+            exc.error_count() == 1
+            and val_errors[0]["loc"] == ("jobs",)
+            and isinstance(val_errors[0]["input"], dict)
+        ):
+            exit_with_error_msg(
+                "It seems that you have used a development version of jobflow-remote. The internal format"
+                "of the batch jobs have changed before the release. Please run 'jf batch fix-batch-doc-jobs-dict'"
+                " to upgrade the database content."
```
davidwaroquiers: Maybe add something like this ?

" to upgrade the database content (note that this command is hidden from the help and can only be discovered through this message)."

src/jobflow_remote/cli/batch.py
```diff
@@ -196,3 +238,12 @@ def delete(
         )
```
davidwaroquiers, line 237: I think the actual jc.delete_batches should take the ids from the to_delete variable here so that what's deleted is exactly what's written before no ?

src/jobflow_remote/cli/batch.py
```diff
@@ -145,43 +185,45 @@ def delete(
...
+        ):
+            text = Text.from_markup(
+                "[red]This operation may remove batch processes in states other than 'FINISHED'. "
+                "This could lead to [bold]inconsistencies or data loss[/bold]. "
+                "Overall it will [bold]delete {len(to_delete)} batch processes[/bold]. Proceed anyway?[/red]"
+            )
+        else:
+            text = Text.from_markup(
+                f"[red]This operation will [bold]delete {len(to_delete)} batch processes[/bold]. Proceed anyway?[/red]"
+            )
```
davidwaroquiers, line 228: Just thinking here that we could have three options:
- Don't delete
- Delete all
- Delete only those that are finished
Probably relatively easy to do as we already have the info in the to_delete of course.

tests/db/cli/test_batch.py
```diff
@@ -91,6 +94,32 @@ def test_batch_worker(
         ],
     )
 
+    job_info = job_controller.get_job_info(job_id=job_ids_db_ids[0][0])
+    missing_proc_ids = [
+        ob.process_id
+        for ob in ordered_batches
+        if ob.process_id != job_info.remote.process_id
+    ]
```
davidwaroquiers: Just to be sure I understood correctly. This variable contains the list of batches that do NOT contain the job provided right ? Could it be that one (or more) of these process ids sometimes appear in the table and make the test fail ? (somewhat similar to the names of the jobs that were add3 and could be within a uuid if we were unlucky). If that is the case, maybe we could replace the check using the batch_uuid instead ? (note that I think this requires changing the RemoteInfo to add the batch_uid there, set to None when it's not a batch process).
What do you think ?
I think it's not adding any relevant feature (the fact of adding batch_uid within RemoteInfo) so I guess we can (should ?) do that only if this random failure actually happens sometimes..

tests/db/cli/test_batch.py
```diff
@@ -143,4 +183,69 @@ def test_batch_worker(
...
+    job_controller.add_batch_process(
+        process_id="1234", batch_uid=uuid1, worker="test_local_batch_worker"
+    )
+    job_controller.add_batch_process(
+        process_id="5678", batch_uid=uuid2, worker="test_local_batch_worker"
+    )
+    job_controller.update_job_in_batch(
+        batch_uid=uuid1, job_id="jobuuid1", job_index=1, db_id="1"
+    )
+    job_controller.batches.find_one_and_update(
```
davidwaroquiers, line 219: Maybe add a comment here to explicitly mention that a dict is added here on purpose.

davidwaroquiers (approved):
Thanks a lot for the updates!
Just made a few minor comments. I let you decide what to do, just maybe the one on the deletion needs to be addressed as I think we said to have it the same way as it is in the flows, or maybe something changed ?

src/jobflow_remote/cli/batch.py
davidwaroquiers: Could be:
- "y" => delete all finished batch processes
- "n" => stop and don't delete anything
- "a" => delete all batch processes (proceed with care)

src/jobflow_remote/cli/batch.py
gpetretto: right, I forgot to do that.

src/jobflow_remote/cli/batch.py
gpetretto: I am not fully convinced that it is needed. Since the default is that states="FINISHED", the only use case where this would happen is if a user explcitly pass more than one state and one of these is `FINISHED`: `jf batch delete -s FINISHED -s SUBMITTED`. So I think that if someone goes to the extent of adding two states in the same command it is fairly reasonable that this is what they want. I this it is good to have the warning in any case, as in most of the cases potentially problematic.

It could make sense having this multiple option if removing the default value for `--state`, but I think that 99% of the cases one should not have the option to delete `SUBMITTED` and `RUNNING` jobs by default.

tests/db/cli/test_batch.py
gpetretto: good point. Indeed this might be possible. However, it is not needed to change RemoteInfo. I modified this part to collect the `batch_uid` but still matching on `process_id`.

src/jobflow_remote/cli/batch.py
gpetretto: Done. I used a slightly different sentence.

src/jobflow_remote/cli/batch.py
davidwaroquiers: Ah true, I forgot about the default value of batch_state. Thx! Let's keep it this way which is much more logical indeed

davidwaroquiers (approved):
All good to me!
Can be merged

PR #440, opened 2026-01-14
Update release action

gpetretto:
Small updates to the release action and configuration:
* license in pyproject.toml should comply to new format: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license
* in the `CasperWA/push-protected@v2` action the `sleep` keyword is deprecated in favor of `pre_sleep`: https://github.com/CasperWA/push-protected?tab=readme-ov-file#deprecated-inputs
* adding some checks to try to debug the release failure (to be removed afterwards)


davidwaroquiers (approved):
All good to me

PR #456, opened 2026-02-22
New SeparatedTransferHost that accommodates SFTP-disabled submission nodes

mcgalcode:
Hi all! I have been using jobflow-remote lately and it is such a joy. Thank you for the work you've put into it, and for the thoughtful designs.

The purpose of this PR is to accommodate the situation where the node that accommodates SSH access and queue access cannot also accommodate file transfer protocols. At our supercomputing cluster, we have separate data-transfer nodes and login nodes (a common arrangement I think). The data-transfer nodes do not allow job submission, and the login nodes do not allow SFTP connections. As a result, the jobflow-remote copy-files-then-submit process cannot occur on either one of these node types. To get around this, I introduced a new type of host which routes command execution and file movement to different nodes (these nodes would canonically be the SFTP-enabled data transfer node and the SLURM/PBS-enabled login node).

This implementation has been working for me, so I thought I would offer it to upstream. I am looking forward to hearing your thoughts on it, and also to potentially be corrected if there is a way to do this already with jobflow-remote configuration that I missed.

Thanks again!
Max

### DIFF
--- a/src/jobflow_remote/remote/host/separated.py
+++ b/src/jobflow_remote/remote/host/separated.py
@@ -165,7 +165,7 @@ class SeparatedTransferHost(BaseHost):
         return self.command_host.rmtree(path, raise_on_error)
 
     # -------------------------------------------------------------------------
-    # File operations via SSH commands - delegated to command_host
+    # File operations via SFTP commands - delegated to transfer_host
     # -------------------------------------------------------------------------
 
     def write_text_file(self, filepath: str | Path, content: str) -> None:
--- a/src/jobflow_remote/remote/host/separated.py
+++ b/src/jobflow_remote/remote/host/separated.py
@@ -162,7 +162,7 @@ class SeparatedTransferHost(BaseHost):
         bool
             True if the directory tree was successfully removed.
         """
-        return self.command_host.rmtree(path, raise_on_error)
+        return self.command_host.rmtree(path, raise_on_error=raise_on_error)
 
     # -------------------------------------------------------------------------
     # File operations via SFTP commands - delegated to transfer_host
--- a/src/jobflow_remote/remote/host/separated.py
+++ b/src/jobflow_remote/remote/host/separated.py
@@ -101,14 +101,10 @@ class SeparatedTransferHost(BaseHost):
         """
         return self.command_host.shell(pre_cmd, shell)
 
-    # -------------------------------------------------------------------------
-    # File operations - delegated to transfer_host
-    # -------------------------------------------------------------------------
-
     def mkdir(
         self, directory: str | Path, recursive: bool = True, exist_ok: bool = True
     ) -> bool:
-        """Create directory on the transfer host.
+        """Create directory on the command host.
 
         Parameters
         ----------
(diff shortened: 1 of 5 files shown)


gpetretto:
Hi @mcgalcode, thanks a lot for the feedback and for sharing your implementation. I think it will be useful for other users as well.

We had some internal discussion because we considered if this could be asbtracted to be the base for a more general mix of command execution and tools for transfer. However, since there is no request in that direction at the moment, it would be difficult to make a meaningful abstraction at this time. Thus we think that it would be good to go on with merging your implementation almost as it is and see if there is a need to make a more general split in the future.

Related to the above point, if at some point we have to change the structure of the Host objects one potential problem would be the structure of the configuration file, which is validated by pydantic. Any backward incompatible change may result in an invalid configuration file for the users. For this reason I would ask you if, instead of making the additional `transfer` an attribute of `RemoteWorker`, it could be possible to make an additional type of worker (e.g. `SeparetedTransferRemoteWorker`, or if you can think of a more suitable name) which is a subclass of `RemoteWorker`, contains the additional `transfer` option and overrides the `type` attribute to something like `separated_transfer_remote`. I hope this will make it easier for us if in the future to handle other potential use cases, if we will ever have a need for them. What do you think?

I will also add a few more comments in the code.

src/jobflow_remote/remote/host/separated.py
```diff
@@ -0,0 +1,305 @@
...
+            If True, create parent directories as needed.
+        exist_ok : bool
+            If True, do not raise an error if directory exists.
+
+        Returns
+        -------
+        bool
+            True if the directory was created successfully.
+        """
+        return self.transfer_host.mkdir(directory, recursive, exist_ok)
```
gpetretto: If you check the original implementation of `mkdir` you see that it does not create the directory using the sfpt server, but rather executing `mkdir -p` in the shell. This should be to handle the case of creation of directories recursively. I think that in your case it would be more reasonable to address this with the `command_host`. Do you agree?

src/jobflow_remote/remote/host/separated.py
```diff
@@ -0,0 +1,305 @@
...
+        """Copy a file on the transfer host.
+
+        Parameters
+        ----------
+        src : str or Path
+            Source path on remote host.
+        dst : str or Path
+            Destination path on remote host.
+        """
+        return self.transfer_host.copy(src, dst)
```
gpetretto: same as `mkdir`

src/jobflow_remote/remote/host/separated.py
```diff
@@ -0,0 +1,305 @@
...
+        """Move a file on the transfer host.
+
+        Parameters
+        ----------
+        src : str or Path
+            Source path on remote host.
+        dst : str or Path
+            Destination path on remote host.
+        """
+        return self.transfer_host.move(src, dst)
```
gpetretto: same as `mkdir`

src/jobflow_remote/remote/host/separated.py
```diff
@@ -0,0 +1,305 @@
...
+        raise_on_error : bool
+            If False (default), errors will be ignored. Otherwise, errors
+            will raise an exception.
+
+        Returns
+        -------
+        bool
+            True if the directory tree was successfully removed.
+        """
+        return self.transfer_host.rmtree(path, raise_on_error)
```
gpetretto: same as `mkdir`

gpetretto (commented):
Thanks a lot for the updates!

I have one last point: I noticed that some of the methods should maybe be switched to the `command_host`. In fact, due to some limitations in the fabric interface or the sftp commands that can execute, even if they act on the files they are done by calling the `execute` method rather than the sftp interface. I suppose you still have a full ssh access to the other node, so it probably works fine, but if that was only an sftp server I imagine that those commands may have failed. 
What do you think?

src/jobflow_remote/remote/host/separated.py
mcgalcode: Yes I completely agree!

src/jobflow_remote/remote/host/separated.py
mcgalcode: Yes, that makes sense.

mcgalcode:
Hi @gpetretto thanks for your review! And sorry to push code with out an accompanying message, I was still thinking things over a bit here.

All your comments make sense to me and I'll make the changes and test them with my set up. I agree that having those bash commands running on the command host makes more sense. I remember thinking about this earlier but I can't recall why I did that in the first place.

gpetretto:
Thanks! 
I have seen that the change that I suggested abouth the `SeparatedTransferWorker` subclass is breaking the typing. 
I think that the correct approach here would be to have a `BaseRemoteWorker` that inherits from `BaseWorker` and contains all the attributes that are now in `RemoteWorker` and `type` is left as inherited from `BaseWorker`. Then both `RemoteWorker` and `SeparatedTransferWorker` inherit from  `BaseRemoteWorker`, define their `type` and implement their version of `get_host`.
Anyway, I am open to other suggestions if you think that there would be a better way of fixing this.

mcgalcode:
@gpetretto Ahh yes I should have run the pre-commit hooks. You're right. I tried to think of a better way than what you described here, but nothing came up. I think it's pretty clean still, and makes sense for the class hierarchy.

src/jobflow_remote/remote/host/separated.py
```diff
@@ -0,0 +1,305 @@
...
+
+        Returns
+        -------
+        bool
+            True if the directory tree was successfully removed.
+        """
+        return self.command_host.rmtree(path, raise_on_error)
+
+    # -------------------------------------------------------------------------
+    # File operations via SSH commands - delegated to command_host
```
davidwaroquiers:
```suggestion
    # File operations via SSH commands - delegated to transfer_host
```

mcgalcode: Ah good catch!

gpetretto:
Everything looks good to me. Thanks! Can you maybe just fix the two linting errors? They seem to be quite pointless in the tests, so I think you can even mark them to be ignored if it is easier than changing the code.

After that I would merge, unless there is anything else that you plan to add.

mcgalcode:
Nothing more planned here for now on my end! Linting errors are fixed.

gpetretto (approved):

mcgalcode:
@gpetretto Thank you for fixing that last issue! Sorry about that, not sure how I missed it!

PR #464, opened 2026-03-23
Use projection in MongoLock

gpetretto:
I realized that the `projection` field was never really used in the `MongoLock`. Everything always worked since all the document was pulled, so it was a superset of what was really needed. I don't think that this had a big impact on performances, since it is used in some cases and the amount of data left out is probably not too big.
I am fixing it so that it is actually taken into account.

### DIFF
--- a/src/jobflow_remote/utils/db.py
+++ b/src/jobflow_remote/utils/db.py
@@ -261,7 +261,7 @@ class MongoLock:
         # if projecting always get the lock as well
         if projection:
             projection = list(projection)
-            projection.extend([self.LOCK_KEY, self.lock_id])
+            projection.extend([self.LOCK_KEY, self.lock_id, self.LOCK_TIME_KEY])
 
         # Modify the filter if the document should not be fetched if
         # the lock cannot be acquired. Otherwise, keep the original filter.
@@ -305,6 +305,7 @@ class MongoLock:
                 update,
                 upsert=False,
                 return_document=ReturnDocument.AFTER,
+                projection=projection,
                 **self.kwargs,
             )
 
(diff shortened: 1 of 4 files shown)


src/jobflow_remote/utils/db.py
```diff
@@ -261,7 +261,7 @@ def acquire(self) -> None:
         # if projecting always get the lock as well
         if projection:
             projection = list(projection)
-            projection.extend([self.LOCK_KEY, self.lock_id])
+            projection.extend([self.LOCK_KEY, self.lock_id, self.LOCK_TIME_KEY])
```
davidwaroquiers: I don't remember what is the goal of the lock_time. It is just for information purposes right ? It might be good to add a comment in the docstring of MongoLock above as to what is used internally somehow. Maybe a "Notes" section or something for how it is implemented and what is used to make the matching. What do you think ?

davidwaroquiers: Reading again this line, is it actually correct ? Shouldn't it be:

```suggestion
            projection.extend([self.LOCK_KEY, self.LOCK_TIME_KEY])
```

Because basically the lock_id is what is in the LOCK_KEY of the MongoLock dict document no ?

davidwaroquiers (approved):
All good to me, thanks!
Just two comments (on the same line)
- One on docs for the lock time (and other related internal stuff used in MongoLock). Maybe it could be useful to have a Notes section for how it is implemented, what is used exactly
- One on the line itself, which I think is not correct (even though I think it's not detrimental anyway, but I may be wrong - about it being not correct or about it being not detrimental)

src/jobflow_remote/utils/db.py
gpetretto: I think you are right. Probably when I wrote it the `lock_id` should have been `LOCK_TIME_KEY` and it was never noticed because of the bug.

I removed it and improved the docstring.

davidwaroquiers (approved):
Great, thx!

Issue #483, opened 2026-05-13
wk_name and worker_name different


jsyony37:
Hello,

I have two workers defined in the project.yaml file, one called "cascade" and another called "rome". But when I submit a workflow specifically to one of the workers (e.g. "rome") it sometimes (but apparently not always) fails with an error that says "Remote error: wk_name: rome, worker_name: cascade." See below. I can't figure out why it confuses with the two workers. Any idea why this is happening and what may rectify this? 

<img width="1457" height="273" alt="Image" src="https://github.com/user-attachments/assets/e5b23fae-6992-4785-9989-39d7da2805e7" />

The line that submits is
`submit_flow(flow,worker="rome")`

Thanks,
Junsoo

gpetretto:
Hi @jsyony37, 
thanks for reporting this.
A few questions:
* which version are you using?
* Since I don't find that error message anywhere, could it be that this line was modified? https://github.com/Matgenix/jobflow-remote/blob/57f1633940d1b7e9b775267c454df587a2a55f9b/src/jobflow_remote/jobs/runner.py#L1295
* Can you check if the two workers have the same `jobs_handle_dir` in the configuration?

If the `jobs_handle_dir` is shared, I think this would explain the random behaviour, since the two workers may be "stealing" jobs from each other. In that case, could you try setting two different folders and verify if the problem comes up again? If this is the issue I will clarify this in the documentation (and maybe with a validation of the configuration?), otherwise I will try to find another possible sorce of the issue.

jsyony37:
Yes, it was indeed the problem of handle_dir being shared between the two nodes. Upon separation, both workers work without an issue. Thanks!

gpetretto:
Great! Thanks for the update. I will try to improve the code to prevent this occurrence.

PR #484, opened 2026-05-21
Improve project validation and add project cross checks

gpetretto:
Following #483, this PR adds a check to prevent having the same `jobs_handle_dir` for two batch workers on the same host.
As a general improvement, also added a `jf project check-conflicts` that will verify this and a few other incompatibilities among different projects. This is not executed when the projects are read, so it needs to be called explicitly. I don't think it would be good to enfornce it and having the need to perform all the checks every time.

Also, basically unrelated, I have realized that the case `OnMissing.PASS` from jobflow was not handled at all in jobflow-remote. This should fix it.

### DIFF
--- a/src/jobflow_remote/cli/project.py
+++ b/src/jobflow_remote/cli/project.py
@@ -197,13 +197,14 @@ def check(
     import os
 
     from jobflow_remote import SETTINGS
+    from jobflow_remote.config.settings import JobflowRemoteSettings
 
     prefix = SETTINGS.model_config["env_prefix"]
     extra_vars = [
         k
         for k in os.environ
         if k.lower().startswith(prefix)
-        and k[len(prefix) :].lower() not in SETTINGS.model_fields
+        and k[len(prefix) :].lower() not in JobflowRemoteSettings.model_fields
     ]
     if extra_vars:
         out_console.print(
@@ -321,6 +322,12 @@ def check(
         for e in errors:
             out_console.print(e[0], style="bold")
             out_console.print(e[1])
+    if not errors and SETTINGS.cli_suggestions:
+        out_console.print(
+            "The project configuration is correct. You might want to check for "
+            "conflicts with other project by running `jf project check-conflicts`",
+            style="green",
+        )
 
 
 @app_project.command(name="check-conflicts")
@@ -349,10 +356,11 @@ def check_conflicts() -> None:
         out_console.print(msg, style="yellow")
 
     if len(cm.projects) < 2:
-        exit_with_warning_msg(
-            f"Only {len(cm.projects)} project(s) parsed in {cm.projects_folder}; "
-            "nothing to cross-check."
-        )
+        if len(cm.projects) == 1:
+            msg = f"Only 1 project parsed in {cm.projects_folder}; "
+        else:
+            msg = f"No projects parsed in {cm.projects_folder}; "
+        exit_with_warning_msg(msg + "nothing to cross-check.")
 
     with loading_spinner(processing=False) as progress:
         progress.add_task("Checking projects for conflicts")
--- a/src/jobflow_remote/cli/project.py
+++ b/src/jobflow_remote/cli/project.py
@@ -33,6 +33,7 @@ from jobflow_remote.cli.utils import (
 from jobflow_remote.config import ConfigError, ConfigManager, Project
 from jobflow_remote.config.helper import (
     check_jobstore,
+    check_projects_conflicts,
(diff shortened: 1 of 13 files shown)


src/jobflow_remote/config/base.py
```diff
@@ -835,6 +836,34 @@ def check_jobstore(cls, jobstore: dict) -> dict:
...
+            if key in seen:
+                raise ValueError(
+                    f"Workers {seen[key]!r} and {worker_name!r} share the "
+                    f"same `jobs_handle_dir` ({key[1]}) on the same host. It "
+                    "must be unique across batch workers that share a host, "
+                    "to avoid unpredictable behaviour at run time."
+                )
+            seen[key] = worker_name
+        return self
+
```
davidwaroquiers, line 866: So this is the validation of a given project (done everytime) while the cross-validation is done in the check-conflicts cli. Right ?

src/jobflow_remote/cli/project.py
```diff
@@ -322,6 +323,59 @@ def check(
...
+        )
+        if not_parsed:
+            msg += f"\n - projects: {', '.join(not_parsed)}"
+        if erroneous_files:
+            msg += f"\n - files: {', '.join(erroneous_files)}"
+        out_console.print(msg, style="yellow")
+
+    if len(cm.projects) < 2:
+        exit_with_warning_msg(
+            f"Only {len(cm.projects)} project(s) parsed in {cm.projects_folder}; "
```
davidwaroquiers: This would either be "Only 1 project parsed" or "No project parsed". Not super important but maybe change that ?

davidwaroquiers (approved):
All good to me, you can merge as is or change the small thing about 1 project / No project.
Do you think the check-conflicts should be added to the documentation maybe somewhere ? (one question also, maybe when doing jf project check, it should check conflicts with other projects no ?)

src/jobflow_remote/config/base.py
gpetretto: yes

src/jobflow_remote/cli/project.py
gpetretto: good point. Will do.

gpetretto:
> All good to me, you can merge as is or change the small thing about 1 project / No project. Do you think the check-conflicts should be added to the documentation maybe somewhere ? (one question also, maybe when doing jf project check, it should check conflicts with other projects no ?)

I see the point here. I will add it to the docs. I am not fully convinced of adding this to add it to the jf project check. Maybe a warning that one might consider running jf project check-conflicts to check potential conflicts with other projects?

davidwaroquiers:
> > All good to me, you can merge as is or change the small thing about 1 project / No project. Do you think the check-conflicts should be added to the documentation maybe somewhere ? (one question also, maybe when doing jf project check, it should check conflicts with other projects no ?)
> 
> I see the point here. I will add it to the docs. I am not fully convinced of adding this to add it to the jf project check. Maybe a warning that one might consider running jf project check-conflicts to check potential conflicts with other projects?

For the addition to the project check, I don't know either, it's just that it probably does not take much time to do. Maybe it's done when you do a full project check (using jf project check --full) ? Why are you not convinced ? (I can't say exactly myself why I'm not :D)

gpetretto:
mainly because it is not obvious that one wants that check to be done and may be somehow confusing. The description says 
> Check that the connection to the different elements of the projects are working.

not that it will make a validation. That is in principle more for `jf project list`. But I am not really convinced of having this additional message there either. Mainly bacause if you have many projects the output can be already a bit messy on its own. Maybe changing this entirely and have it as an option for `jf project list` and remove `jf project check-conflicts`?

davidwaroquiers:
> mainly because it is not obvious that one wants that check to be done and may be somehow confusing. The description says
> 
> > Check that the connection to the different elements of the projects are working.
> 
> not that it will make a validation. That is in principle more for `jf project list`. But I am not really convinced of having this additional message there either. Mainly bacause if you have many projects the output can be already a bit messy on its own. Maybe changing this entirely and have it as an option for `jf project list` and remove `jf project check-conflicts`?

Yes true, that might be misleading or we should change the description at least. I let you decide what you think is best. Maybe I would change the description to something like : "Check that the connection to the different elements of the project are working and checks potential conflicts with other projects". But if you prefer to keep it just as such it's fine too. It's just that in that case we are not adding an additional cli just for that (but again I let you decide)

gpetretto:
decided to merge as it is

PR #489, opened 2026-06-01
compatibility for typer dropping click depenency

gpetretto:
Typer dropped click as a backend from 0.26.0 and this breaks an import from click.
Here the attempt is to cover all the possible cases and cover different typer versions.

### DIFF
--- a/src/jobflow_remote/cli/job.py
+++ b/src/jobflow_remote/cli/job.py
@@ -19,7 +19,6 @@ from jobflow_remote.cli.formatting import (
 from jobflow_remote.cli.jf import app
 from jobflow_remote.cli.jfr_typer import JFRTyper
 from jobflow_remote.cli.types import (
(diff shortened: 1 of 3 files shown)


davidwaroquiers (approved):
LGTM!
Thanks

gpetretto:
Since the `_click` submodule in typer is likely going to be modified, I dropped the old approach to the custom parser, thus removing the click dependence in that point. 
The try/except for importing the exception should not be a problem, even if the `_click` module is removed. It will just be useless, but should still guarantee backward compatibility.

Along with this, I also removed a couple of custom types that should be a leftover of when the minimum version of typer was 0.9. I did some local tests, but most importantly since the tests pass with all python versions+lowest/highest dependencies it seems fine to remove them.

davidwaroquiers (approved):
All good to me!

Issue #490, opened 2026-06-02
Graph Flow Visualization fails in GUI


Luguza:
In the web GUI (jf gui), clicking the Graph Flow tab in a flow's dialog does nothing visible — the content panel never switches to the graph; it stays on the Details tab. 
The endpoint get_graph_job_flow calls JobController.get_flows_info(limit=1, full=True), but in 1.0.0 that parameter was renamed full → with_jobs_info. The GUI was never updated to match. The CLI equivalent (jf flow graph ... --mermaid) was updated and works, so this is GUI-only.

<img width="228" height="156" alt="Image" src="https://github.com/user-attachments/assets/cb31c52a-eba7-4873-8268-7a716409e967" />

```
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/lustre/work/ws/ws1/ka_qa7606-dft_pipeline/DFT_Pipeline2/.venv/lib/python3.12/site-packages/uvicorn/protocols/http/httptools_impl.py", line 421, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lustre/work/ws/ws1/ka_qa7606-dft_pipeline/DFT_Pipeline2/.venv/lib/python3.12/site-packages/uvicorn/middleware/proxy_headers.py", line 63, in __call__
    return await self.app(scope, receive, send)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lustre/work/ws/ws1/ka_qa7606-dft_pipeline/DFT_Pipeline2/.venv/lib/python3.12/site-packages/starlette/applications.py", line 113, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/lustre/work/ws/ws1/ka_qa7606-dft_pipeline/DFT_Pipeline2/.venv/lib/python3.12/site-packages/starlette/middleware/errors.py", line 187, in __call__
    raise exc
  File "/lustre/work/ws/ws1/ka_qa7606-dft_pipeline/DFT_Pipeline2/.venv/lib/python3.12/site-packages/starlette/middleware/errors.py", line 165, in __call__
    await self.app(scope, receive, _send)
[...]

PR #491, opened 2026-06-02
Fix get_flows_info call in gui

gpetretto:
Fixing #490


davidwaroquiers (approved):
All good to me. Just maybe add why this is needed as discussed.

davidwaroquiers (approved):
All good thx!

PR #494, opened 2026-06-15
Several small bugfix

gpetretto:
Fix for several bugs identified by AI. Mostly harmless (text format or in functions not used anymore) or with impact only in limited corner cases.


src/jobflow_remote/utils/remote.py
```diff
@@ -49,11 +54,24 @@ def __init__(self, project: Project | None = None):
...
+            cls._project = project
+        elif project is not None and project.name != cls._project.name:
+            # A context is already open and bound to another project. Sharing
+            # host connections across projects is not supported, so refuse
+            # instead of silently using the wrong project's workers.
+            raise ValueError(
+                f"SharedHosts is already active for project '{cls._project.name}' "
+                f"and cannot be used for a different project ('{project.name}') "
+                "at the same time."
+            )
```
davidwaroquiers, line 74: Do we need to have a test for this maybe ?

src/jobflow_remote/cli/utils.py
```diff
@@ -205,7 +205,7 @@ def check_at_least_one_opt(d: dict) -> None:
         if v:
             not_none.append(k)
 
-    if len(not_none) > 1:
+    if len(not_none) <= 1:
```
davidwaroquiers: Here it should be < 1 no ? Also, I don't remember if an empty string can be a meaningful value for  a given option, in which case the "if v:" should rather be "if v is not None:" (but I have the feeling that an empty string is not a meaningful value/possible thing in the cli ?)

tests/db/jobs/test_jobcontroller.py
```diff
@@ -795,6 +796,26 @@ def test_delete_job(job_controller, two_flows_four_jobs, runner):
...
+            job_controller.delete_jobs(
+                job_ids=[(add2.uuid, 1)], delete_output=True, delete_files=True
+            )
+        )
+        == 1
+    )
+    with pytest.raises(ValueError, match=".*has no outputs.*"):
+        job_controller.jobstore.get_output(add2.uuid)
+
+    assert not os.path.isdir(add2_info.run_dir)
```
davidwaroquiers, line 819: Does it need to check that get_output has something before the deletion and the run_dir exists before the deletion ?

davidwaroquiers (approved):
Almost all good to me! You can go ahead after checking if what I reviewed is reasonable :)

src/jobflow_remote/utils/remote.py
gpetretto: done

src/jobflow_remote/cli/utils.py
gpetretto: Indeed it should be < 1

tests/db/jobs/test_jobcontroller.py
gpetretto: even if the point was actually to run delete_jobs, as it was bugged, it is good to also check that they existed before. I added the check

PR #498, opened 2026-06-25
Add --latest-flow / -lf option to jf job list

Luguza:
### Summary
Adds a new --latest-flow (-lf) option to the jf job list command that restricts the output to Jobs belonging to the most recently created Flows. This makes it easy to inspect the jobs from your latest submission(s) without having to look up flow IDs manually.

### Behavior
jf job list -lf 3 — show jobs from the 3 most recently created Flows (by creation date).
jf job list -lf 1 — show jobs from the single latest Flow.
The filter is applied before all other filters.
Incompatible with --flow-id; passing both raises a BadParameter error.
Must be a positive integer; values < 1 are rejected.
Errors out cleanly if there are no Flows in the database.

### Notes
The option takes its value space-separated (-lf 3) or with = (-lf=3). It is a plain integer option — the bare flag form is intentionally not supported, since Click 8.2 removed options with an optional value; use -lf 1 for the single latest Flow.

### DIFF
--- a/src/jobflow_remote/cli/job.py
+++ b/src/jobflow_remote/cli/job.py
@@ -99,7 +99,7 @@ def jobs_list(
     locked: locked_opt = False,
     custom_query: query_opt = None,
     latest_flow: Annotated[
-        int,
+        int | None,
         typer.Option(
             "--latest-flow",
             "-lf",
@@ -108,7 +108,7 @@ def jobs_list(
             "consider, e.g. '--latest-flow 3' or '-lf 1' for the single latest Flow. "
             "Incompatible with --flow-id; applied before all other filters.",
         ),
-    ] = 0,
+    ] = None,
     error: Annotated[
         bool,
         typer.Option(
@@ -153,7 +153,7 @@ def jobs_list(
 
     jc = get_job_controller()
 
-    if latest_flow:
+    if latest_flow is not None:
         if flow_id:
             raise typer.BadParameter("--latest-flow cannot be combined with --flow-id")
         if latest_flow < 1:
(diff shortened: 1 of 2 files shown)


src/jobflow_remote/cli/job.py
```diff
@@ -142,6 +153,21 @@ def jobs_list(
...
+                "--latest-flow cannot be combined with --flow-id"
+            )
+        if latest_flow < 1:
+            raise typer.BadParameter("--latest-flow must be a positive integer")
+        flows_info = jc.get_flows_info(
+            sort=[("created_on", -1)],
+            limit=latest_flow,
+        )
+        if not flows_info:
+            exit_with_error_msg("No Flows in the database")
```
gpetretto: I don't think this error is needed. If you do not specify `-lf` an empty DB would just give an empty list of jobs. I think it would be fine in this case as well.

gpetretto (commented):
Hi @Luguza, sorry for the delay and thanks for the contribution! I think this is indeed a valid use case.
I just left a minor comment. 
Can you also add tests for the new functionality?

gpetretto:
Sorry, I have just seen that the linting is also failing. For that you would need to install the [pre-commit hook](https://pre-commit.com/) that will run all the linting tools with the appropriate rules when you `git commit` your modifications.

For the file already modified you can run `pre-commit run --files src/jobflow_remote/cli/job.py`.

Luguza:
Sorry for the unnecessary CI triggers earlier. I forgot to run the pre-commit hook on the tests.

gpetretto:
> Sorry for the unnecessary CI triggers earlier. I forgot to run the pre-commit hook on the tests.

No Problem. Thanks for fixing them and for the new feature!

gpetretto (approved):

PR #509, opened 2026-07-21
Fix failing tests

gpetretto:
The latest version of monty changes something in `loadfn` (or `MontyDecoder`) so that deserializes some thing that it did not use to. Parsing json/yaml project files with the library. 
Also fixing timings for a test usually failing in CI


davidwaroquiers (approved):
LGTM. Just make the small change of moving _yaml consistently outside or inside the extensions loop

